PR #57, opened 2019-04-01
Added gitattributes file to preserve LF format on Windows

mfigurski80:
Tested and functional


mfigurski80:
Updated. 
Not sure if windows edits would actually deformat, since it seems like github does convert filetypes, but everything runs smoothly on LF so it doesn't make a difference.

akotlar:
Great job!

akotlar:
> Updated.
> Not sure if windows edits would actually deformat, since it seems like github does convert filetypes, but everything runs smoothly on LF so it doesn't make a difference.

I think by default github uses native line endings. The LOC would be huge for small edit size.

PR #62, opened 2019-06-28
Create Elasticsearch.md

akotlar:
How a new node is created. This should generally not need to be done more than once every few months (I expect i3.2xlarge spot nodes not to die much more often than 3x a year), but should be automated as a startup script (most basic would be user data), or in the case we move to Kubernetes, as a deployment.yml  + Docker image


akotlar:
added note about config's elasticsearch hosts

PR #76, opened 2020-01-26
[infrastructure] update dependencies, install script

akotlar:


install-apt.sh
```diff
@@ -16,7 +16,7 @@ fi
 . ~/.profile;
 
 # Perlbrew simplifies version management
-. ./install/install-perlbrew-linux.sh $INSTALL_DIR perl-5.28.0 ~/.profile;
+. ./install/install-perlbrew-linux.sh $INSTALL_DIR perl-5.30.1 ~/.profile;
```
akotlar: 5.28.0 doesn't exist in perlbrew any longer, 5.28.2 won't build on the linux 2 ami

akotlar: Will need to do the same thing for the master branch

akotlar:
bump, could we review this on Monday?

PR #86, opened 2021-07-20
revert hg19.clean.yml, add expanded hg19_2020.clean.yml; add sort::xs to install

akotlar:
hg19.clean.yml shold remain the simpler example config. The expanded config for gnomad 2.1 is moved to hg19_2020.clean.yml.

This also adds Sort::XS earlier into the install process, as in a recent install it failed to install


akotlar:
out of date

PR #92, opened 2023-03-19
Feature/cythonize

akotlar:
* Adds python version of indexer, with cython file reader as proof of concept
* Restructures all search services as Maturin project, with support for Rust and Cython libraries

Supersedes https://github.com/akotlar/bystro/pull/91
We will move over to Cython + Rust where Cython requires too much C/C++ to be worthwhile, and forego Go and Perl where possible. 

Cython is an incredibly easy way to extract significant performance gains from python code. Code largely is unchanged from regular python, just with addition of types. 

In a followup PR I will implement that last remaining pieces of the save function (streaming filters) and remove the rest of the Perl code related to SaveFromQuery


akotlar:
cc @wingolab

akotlar:
In a followup PR I will implement that last remaining pieces of the save function (streaming filters) and remove the rest of the Perl code related to SaveFromQuery

dlin30 (approved):

PR #98, opened 2023-05-05
Add contrib and ML package directory for analysis packages in testing phase

cristinaetrv:


poneill:
Hey Cristina, thanks for this.  It looks like this PR is dragging in a lot of other commits going back to August.  If it just adds the contrib folder, could we do this off a fresh branch from akotlar:master?

(I'm not sure off the top of my head whether this repo's policy is to squash commits by default.  My vote is that it should, but if we're not currently squashing then we should definitely rebranch off of master.)

cristinaetrv:
Hi Patrick, I can go ahead and squash these commits before merging if that sounds like the best option. It's all related the same analysis tool.

poneill:
sgtm, thanks!

poneill (approved):
lgtm

PR #100, opened 2023-05-06
Create pylint.yml

akotlar:
Basic pylint config

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,23 @@
+name: Pylint
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.11"]
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pylint.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
```
poneill, line 20: I think we want to specify a version here, and add a .pylintrc so developers can know what they're on the hook for

poneill: also, is this going to block all further development until the codebase is up to spec?  We also have the option of linting only those files in the changeset, which would allow us to lint incrementally.  Happy to play it either way, lmk what you think.

poneill (approved):
LGTM, We should probably specify this further but happy to get the ball rolling for now :)

.github/workflows/pylint.yml
akotlar: Specified the version.

Let's agree to a standard to set, and follow that. For now, I see it as an assistant during review, not a requirement.

PR #101, opened 2023-05-08
Add global ancestry analysis tool pre-alpha stage in contrib/MLpackages

cristinaetrv:


Contrib/MLpackages/GlobalAncestry.py
```diff
@@ -0,0 +1,147 @@
...
+headervcf=pd.read_csv(genofile,delimiter='\t',skiprows=25)
+IDswAncst = pd.read_csv(AncstFile,delimiter = ',')
+
+
+
+#Turn genotypes into dosage format for PCA
+dosagevcf = headervcf.replace('0/0', 0)
+dosagevcf = dosagevcf.replace('0/1', 1)
+dosagevcf = dosagevcf.replace('1/0', 1)
+dosagevcf = dosagevcf.replace('1/1', 2)
```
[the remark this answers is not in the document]
poneill: Also, we're applying these `replace` transformations over the whole file, but we only intend for them to have effect in certain columns.  It might be clearer in that case to specify the cols where we expect this replace to apply, which would help with legibility and reduce the risk of a solution that works by accident.  Off the top of my head I couldn't be sure that no other string in the df contains a substring of the form '{0,1}/{0,1}'.

Contrib/MLpackages/GlobalAncestry.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+#Separate the genotype dosages
+IDs = headervcf.iloc[1]
```
poneill, line 38: slightly surprising that we're referring to both headervcf and dosagevcf here, since one was derived from the other and I wouldn't expect the transformation we applied to affect the IDs [?]

Contrib/MLpackages/GlobalAncestry.py
```diff
@@ -0,0 +1,147 @@
...
+ChromPos = dosagevcf.iloc[:, 0:9]
+Genos = dosagevcf.iloc[:, 9:]
+
+
+#Separate IDs out as a column by just pulling out one row
```
poneill, line 43: what's going on in these three lines?  I know you're documenting line by line here but I think I'm still losing you :)

Contrib/MLpackages/GlobalAncestry.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+#Convert ints to floats and remove excess whitespace for converting to tensor
+Genos = Genos.astype(float)
+Genos.columns = Genos.columns.str.strip()
+
+
+
+#Find the rows with missing data
+missing_mask = np.isnan(Genos).any(axis=1)
```
poneill, line 59: how much data are we missing here?   IIRC 1KGP is pretty clean [?] We should definitely at least be logging the variants we're discarding due to missingness.

Contrib/MLpackages/GlobalAncestry.py
```diff
@@ -0,0 +1,147 @@
...
+log.write('Matrix size:'+str(mat2.size())+'\n')
+
+
+#Convert matrix with PCs back to pandas and add back IDs
+px = pd.DataFrame(mat2.numpy())
+IDsWpcs = pd.concat([IDsascol, px], axis = 1)
+IDsWpcs = IDsWpcs.rename(columns={'index':'Sample name'})
+
+
+IDsAncPCA=pd.merge(IDswAncst,IDsWpcs, how='right', on='Sample name')
```
poneill, line 90: What's the idea for the right merge?  Do we expect any rows in IDsWpcs to be unmatched?

Contrib/MLpackages/GlobalAncestry.py
```diff
@@ -0,0 +1,147 @@
...
+ohe = OneHotEncoder(sparse = False)
+y = ohe.fit_transform(IDsAncPCA[['Superpopulation code']])[:3202]
+
+
+#Keep track of categories to add back later
+SuperpopCat = ohe.categories_
+
+
+#Run the linear regression
+reg = LinearRegression().fit(X_train.values, y)
```
poneill, line 118: Blocking comment: why are we using linear regression for classification here?

Contrib/MLpackages/GlobalAncestry.py
```diff
@@ -0,0 +1,147 @@
...
+#Define X as the PCs w ancestry
+X_train = IDsAncPCA.loc[:, 0:9]
+
+#Define X_test as the sample IDs wo ancestry
+X_test = IDsWoAnc.loc[:, 0:9]
+
+
+#Define Y as Populations after converting to numerical with ohe
+ohe = OneHotEncoder(sparse = False)
+y = ohe.fit_transform(IDsAncPCA[['Superpopulation code']])[:3202]
```
poneill, line 110: where does 3202 come from?

Contrib/MLpackages/GlobalAncestry.py
```diff
@@ -0,0 +1,147 @@
...
+#Log population sizes
+log.write('IDs per population group:'+'\n'+str(IDsAncPCA.value_counts('Superpopulation code'))+'\n')
+
+#IDsAncPCA
+
+#Define X as the PCs w ancestry
+X_train = IDsAncPCA.loc[:, 0:9]
+
+#Define X_test as the sample IDs wo ancestry
+X_test = IDsWoAnc.loc[:, 0:9]
```
poneill, line 105: How do we use this as a test set if we don't have ancestry info for these samples?

poneill (changes requested):
My one blocking change is about the choice of linear regression as a classifier.  Happy to discuss further over a higher bandwidth medium.

poneill:
> After addressing Pat's comments:
> 
> Let's wait to merge this until #105 is in. Then, move this to contrib/ml/experiment/global_ancestry/linear_model.py. I'm basing this on sklearn's project structure, see https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/experimental/enable_hist_gradient_boosting.py.
> 
> We can choose another location, cc @poneill is you have thoughts.

`contrib/ml/experimental` works for me

akotlar:
@cristinaetrv is this PR still relevant?

cristinaetrv:
> @cristinaetrv is this PR still relevant?


This code is continued in #135 , so we can close this one.

PR #104, opened 2023-05-09
[search2/save] Cleanup search2, enable by default

akotlar:
* Python search.save and search.index servers are now enabled by default
* Improves code quality of search module (pylint score >8)

Saving has two outstanding tickets before it is feature complete (https://github.com/akotlar/bystro/issues/102, https://github.com/akotlar/bystro/issues/103), but we should begin relying on it now as the missing outputs do not block the annotation->search->save->search loop and by relying on it now we will identify more issues than if we wait.

### DIFF
--- a/search2/python/search/index/handler.py
+++ b/search2/python/search/index/handler.py
@@ -1,3 +1,6 @@
+# TODO: allow reading directly from annotation_path or get rid of that possibility in annotator
+# TODO: read delimiters from annotation_conf
+
 import argparse
 import asyncio
 from math import ceil
@@ -7,26 +10,33 @@ import time
 
 from opensearchpy._async.client import AsyncOpenSearch
 from opensearchpy._async import helpers as async_helpers
-from orjson import dumps
-from pystalk import BeanstalkClient
+from orjson import dumps  # pylint: disable=no-name-in-module
+from pystalk import BeanstalkClient  # type: ignore
 from ruamel.yaml import YAML
 
-from search.index.bystro_file import read_annotation_tarball
+from search.index.bystro_file import (
+    read_annotation_tarball,
+)  # pylint: disable=no-name-in-module,import-error
+from search.utils.beanstalkd import Publisher
 
 import ray
 
-
-ray.init(ignore_reinit_error='true', address='auto')
-
-# TODO: allow reading directly from annotation_path or get rid of that possibility in annotator
-# TODO: read delimiters from annotation_conf
+ray.init(ignore_reinit_error="true", address="auto")
 
 n_threads = multiprocessing.cpu_count()
 
 
 @ray.remote
 class Indexer:
-    def __init__(self, search_client_args, progress_tracker, chunk_size=1_000, reporter_batch=15_000):
+    """TODO: Add description here"""
+
+    def __init__(
+        self,
+        search_client_args,
+        progress_tracker,
+        chunk_size=1_000,
+        reporter_batch=15_000,
+    ):
         self.search_client_args = search_client_args
         self.client = AsyncOpenSearch(**self.search_client_args)
         self.progress_tracker = progress_tracker
@@ -35,11 +45,15 @@ class Indexer:
         self.counter = 0
 
     async def run(self, data):
-        resp = await async_helpers.async_bulk(self.client, iter(data), chunk_size=self.chunk_size)
+        resp = await async_helpers.async_bulk(
+            self.client, iter(data), chunk_size=self.chunk_size
+        )
         self.counter += resp[0]
 
         if self.counter >= self.reporter_batch:
-            await asyncio.to_thread(self.progress_tracker.increment.remote, self.counter)
+            await asyncio.to_thread(
+                self.progress_tracker.increment.remote, self.counter
+            )
             self.counter = 0
 
         return resp
@@ -50,25 +64,28 @@ class Indexer:
 
 @ray.remote(num_cpus=0)
 class ProgressReporter:
-    def __init__(self, publisher: dict):
+    """A class to report progress to a beanstalk queue"""
+
+    def __init__(self, publisher: Publisher):
         self.value = 0
         self.publisher = publisher
-        self.client = BeanstalkClient(
-            publisher['host'], publisher['port'], socket_timeout=10)
+        self.message = publisher.message._asdict()
+
+        self.message["data"] = {"progress": 0, "skipped": 0}
+
+        self.client = BeanstalkClient(publisher.host, publisher.port, socket_timeout=10)
 
     def increment(self, count: int):
+        """Increment the counter by processed variant count and report to the beanstalk queue"""
         self.value += count
-        self.publisher['messageBase']['data'] = {
-            "progress": self.value,
-            "skipped": 0
-        }
+        self.message["data"]["progress"] = self.value
 
-        self.client.put_job_into(
-            self.publisher['queue'], dumps(self.publisher['messageBase']))
+        self.client.put_job_into(self.publisher.queue, dumps(self.message))
 
         return self.value
 
     def get_counter(self):
+        """Get the current value of the counter"""
         return self.value
 
 
@@ -86,15 +103,14 @@ class ProgressReporterStub:
 
 
 async def go(
-        index_name: str,
-        tar_path: str,
-        annotation_conf: dict,
-        mapping_conf: dict,
-        search_conf: dict,
-        chunk_size=500,
-        paralleleism_chunk_size=5_000,
-        publisher: dict = None,
-        annotation_path: str = None
+    index_name: str,
+    tar_path: str,
+    mapping_conf: dict,
+    search_conf: dict,
+    chunk_size=500,
+    paralleleism_chunk_size=5_000,
+    publisher: Publisher = None,
+    annotation_path: str = None,
 ):
     assert not (tar_path is not None and annotation_path is not None)
     assert tar_path is not None
@@ -112,45 +128,54 @@ async def go(
     )
 
     if publisher:
-        reporter = ProgressReporter.remote(publisher)
+        reporter = ProgressReporter.remote(publisher)  # pylint: disable=no-member
     else:
-        reporter = ProgressReporterStub.remote()
+        reporter = ProgressReporterStub.remote()  # pylint: disable=no-member
 
     client = AsyncOpenSearch(**search_client_args)
 
-    post_index_settings = mapping_conf['post_index_settings']
-    boolean_map = {x: True for x in mapping_conf['booleanFields']}
+    post_index_settings = mapping_conf["post_index_settings"]
+    boolean_map = {x: True for x in mapping_conf["booleanFields"]}
 
     index_body = {
-        'settings': mapping_conf['index_settings'],
-        'mappings': mapping_conf['mappings'],
+        "settings": mapping_conf["index_settings"],
+        "mappings": mapping_conf["mappings"],
     }
 
     delimiters = {
-        'field': "\t",
+        "field": "\t",
         "allele": ",",
         "position": "|",
         "value": ";",
-        "empty_field": "!"
+        "empty_field": "!",
     }
 
-    if not index_body['settings'].get('number_of_shards'):
+    if not index_body["settings"].get("number_of_shards"):
         file_size = os.path.getsize(tar_path)
-        index_body['settings']['number_of_shards'] = ceil(
-            float(file_size) / float(1e10))
+        index_body["settings"]["number_of_shards"] = ceil(
+            float(file_size) / float(1e10)
+        )
 
     try:
         await client.indices.create(index_name, body=index_body)
     except Exception as e:
         print(e)
 
-    data = read_annotation_tarball(index_name=index_name, tar_path=tar_path,
-                                   boolean_map=boolean_map, delimiters=delimiters,
-                                   chunk_size=paralleleism_chunk_size)
+    data = read_annotation_tarball(
+        index_name=index_name,
+        tar_path=tar_path,
+        boolean_map=boolean_map,
+        delimiters=delimiters,
+        chunk_size=paralleleism_chunk_size,
+    )
 
     start = time.time()
-    actors = [Indexer.remote(search_client_args, progress_tracker=reporter, chunk_size=chunk_size)
-              for x in range(n_threads)]
+    actors = [
+        Indexer.remote( # pylint: disable=no-member
+            search_client_args, progress_tracker=reporter, chunk_size=chunk_size
+        )
+        for x in range(n_threads)
+    ]
     actor_idx = 0
     results = []
     for x in data:
@@ -165,13 +190,13 @@ async def go(
 
     errors = []
     total = 0
-    for x in res:
+    for x in res: # pylint: disable=invalid-name
         total += x[0]
         if x[1]:
             errors.append(",".join(x[1]))
 
     if errors:
-        raise Exception("\n".join(errors))
+        raise ValueError("\n".join(errors))
 
     print(f"Processed {total} records")
 
@@ -179,7 +204,8 @@ async def go(
         await actor.close.remote()
 
     result = await client.indices.put_settings(
-        index=index_name, body=post_index_settings)
+        index=index_name, body=post_index_settings
+    )
 
     print(result)
 
@@ -187,13 +213,11 @@ async def go(
 
     return data.get_header_fields()
 
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Process some config files.")
     parser.add_argument(
         "--tar", type=str, help="Path to the tarball containing the annotation"
-    )
-    parser.add_argument(
-        "--annotation_conf", type=str, help="Path to the genome/assembly config, e.g. hg19.yml"
     )
 
     parser.add_argument(
@@ -216,17 +240,17 @@ if __name__ == "__main__":
 
     args = parser.parse_args()
 
-    with open(args.annotation_conf, "r", encoding="utf-8") as f:
-        annotation_conf = YAML(typ="safe").load(f)
-
     with open(args.search_conf, "r", encoding="utf-8") as f:
         search_conf = YAML(typ="safe").load(f)
 
     with open(args.mapping_conf, "r", encoding="utf-8") as f:
         mapping_conf = YAML(typ="safe").load(f)
 
-    asyncio.run(go(index_name=args.index_name,
-                   tar_path=args.tar,
-                   annotation_conf=annotation_conf,
-                   mapping_conf=mapping_conf,
-                   search_conf=search_conf))
+    asyncio.run(
+        go(
+            index_name=args.index_name,
+            tar_path=args.tar,
+            mapping_conf=mapping_conf,
+            search_conf=search_conf,
+        )
+    )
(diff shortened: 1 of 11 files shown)


search2/requirements.txt
```diff
@@ -1,8 +1,9 @@
-maturin==0.14.13
-ray==2.3.0
--f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-jax[cuda]==0.4.4
-pyarrow==11.0.0
+maturin==0.15.1
+ray==2.4.0
```
akotlar, line 2: jax, numpyro will likely be used for performing inference during search and indexing (e.g making vector embeddings, filtering rows based on the results of a statistical test, as seen in SeqFromQuery.pm. By having these in now, we can build a Docker image that can be used for CI/CD. We can also hold off until the modules that need these are built, your call.

search2/python/search/index/listener.py
```diff
@@ -190,15 +26,53 @@ def main():
...
+                                           job_details['inputFileNames']['annotation'])
+
+        return go(index_name=job_details["indexName"],
+                  tar_path=tar_path,
+                  annotation_path=annotation_path,
+                  mapping_conf=mapping_conf,
+                  search_conf=search_conf,
+                  publisher=publisher)
+
+    def submit_msg_fn(base_msg: dict, job_details: dict): # pylint: disable=unused-argument
```
poneill, line 60: why don't we use job_details here?  In any case you can suppress the warning by renaming as `_job_details`, which also makes it clearer  to the reader that you're ignoring the arg.

search2/python/search/save/handler.py
```diff
@@ -1,43 +1,26 @@
...
-        return self.value
-
-    def get_counter(self):
-        return self.value
+from search.utils.beanstalkd import Publisher
+from search.index.handler import ProgressReporter
 
+ray.init(ignore_reinit_error='true', address='auto')
 
 default_delimiters = {
```
poneill, line 25: nit: module-level constants should be UPCASED, and if a dict should be immutable you can lock it down with `frozendict`

search2/python/search/save/handler.py
```diff
@@ -164,15 +148,15 @@ def _make_output_string(rows, delims):
 
 
 @ray.remote
-def _process_query_chunk(query_args: dict, search_client_args: dict, field_names: list, chunk_output_name: str, reporter):
+def _process_query_chunk(query_args: dict, search_client_args: dict, field_names: list,
```
poneill, line 151: what's the type of reporter?

search2/python/search/save/handler.py
```diff
@@ -195,9 +179,9 @@ def _process_query_chunk(query_args: dict, search_client_args: dict, field_names
             row[y] = _populate_data(
                 child_fields[y], doc["_source"][parent_fields[y]])
 
-        if row[discordant_idx][0][0] == False:
+        if row[discordant_idx][0][0] is False:
```
poneill, line 182: could this be condensed to: `row[discordant_idx][0][0] = int(row[discordant_idx][0][0])`?  Is there a third possibility we have to account for?

search2/python/search/save/handler.py
```diff
@@ -212,21 +196,21 @@ def _process_query_chunk(query_args: dict, search_client_args: dict, field_names
...
     input_body: dict,
     search_conf: dict,
+    publisher: Publisher,
     max_query_size: int = 10_000,
     max_slices=1024,
     keep_alive="1d",
 ):
+    """ Main function for running the query and writing the output """
     if not input_body['compress']:
         print("\nWarning: still compressing outputs\n")
```
poneill, line 209: should we be using `logging` here instead?

search2/python/search/save/handler.py
```diff
@@ -212,21 +196,21 @@ def _process_query_chunk(query_args: dict, search_client_args: dict, field_names
...
 
     search_client_args = dict(
         hosts=list(search_conf["connection"]["nodes"]),
-        http_compress=True,  # enables gzip compression for request bodies
```
poneill, line 229: Is the comment not actually true?  we could preserve the info in this comment by renaming http_compress -> use_gzip_on_request_body.

Here, as elsewhere, we should probably be using a less flexible datatype than a dict for this config issue.

search2/python/search/save/handler.py
```diff
@@ -212,21 +196,21 @@ def _process_query_chunk(query_args: dict, search_client_args: dict, field_names
 
     return resp["hits"]["total"]["value"]
 
-
-# TODO: get max_slices from opensearch index settings
-def go(
+async def go( # pylint:disable=invalid-name
```
poneill, line 199: This method is also over 100 lines long, and it would be good to break up into smaller meaningful chunks.  Part of what makes it challenging is that we're operating at several different levels of abstraction at once, ranging from manipulating queries to building file paths to configuring file compression parameters.  Keeping this method at the highest common level of abstraction and suppressing the incidental details, farming out to helper methods as necessary, would go a long way towards readability imho.

search2/python/search/save/listener.py
```diff
@@ -1,194 +1,40 @@
...
-    event_queue,
-    config_path_base_dir,
+    job_details: dict,
 ):
     job_specific_args = {}
 
     for key in required_keys:
         if key not in job_details:
-            raise Exception(f"Missing required key: {key} in job message")
+            raise ValueError(f"Missing required key: {key} in job message")
```
poneill, line 21: This is great, and we could go even further in the spirit of "parse, don't validate" by creating a config class that can only be constructed by passing this validation check-- once it's constructed, we know forevermore that it's correct.

search2/python/search/save/listener.py
```diff
@@ -211,8 +57,29 @@ def main():
...
-    listen(queue_conf, search_conf, config_path_base_dir)
 
+    def handler_fn(publisher: Publisher, job_details: dict):
+        input_body = _coerce_inputs(job_details)
+        return go(input_body=input_body, search_conf=search_conf, publisher=publisher)
+
+    def submit_msg_fn(base_msg: dict, job_details: dict):
+        config_path = get_config_file_path(config_path_base_dir, job_details['assembly'])
+
+        with open(config_path, 'r', encoding='utf-8') as f: # pylint: disable=invalid-name
```
poneill, line 68: another instance of this pattern :)  

again, what's with invalid-name here?

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,155 @@
...
+import time
+import traceback
+from typing import Any, NamedTuple, List, Callable, Optional
+
+import asyncio
+from orjson import loads, dumps  # pylint: disable=no-name-in-module
+from pystalk import BeanstalkClient, BeanstalkError  # type: ignore
+from opensearchpy.exceptions import NotFoundError
+
+class Message(NamedTuple):
```
poneill, line 13: These NamedTuple classes look great, I just wouldn't defer the docstrings

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,155 @@
...
+    """TODO: Add description here"""
+    event: str
+    queueID: str
+    submissionID: str
+    data: Optional[Any]
```
poneill, line 18: nit: this is just Any

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,155 @@
...
+
+    host: str
+    port: str
+    queue: str
+    message: Message
+
+class QueueConf(NamedTuple):
+    """TODO: Add description here"""
+
+    address: List[str]
```
poneill, line 31: nit: address -> addresses

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,155 @@
...
+    clients = tuple(
+        (h, ports[i], BeanstalkClient(h, ports[i], socket_timeout=10))
+        for i, h in enumerate(hosts)
+    )
+
+    i = 0
+    while True:
+        i += 1
+        offset = i % len(hosts)
+        host = clients[offset][0]
```
poneill, line 97: nit: `host, port, client = clients[offset]`.  Also, if you can unpack an element of `clients` this way it's a sign that `clients` could be more accurately named [?]

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,155 @@
...
+):
+    """TODO: Listen on beanstalkd here directly"""
+    hosts, ports = queue_conf.split_host_port()
+
+    for event in ("progress", "failed", "started", "completed"):
+        assert event in queue_conf.events
+
+    tube_conf = queue_conf.tubes[tube]
+    events_conf = queue_conf.events
+    clients = tuple(
```
poneill, line 88: why is this a tuple?

poneill (changes requested):
Hey, this looks like a ton of great work.  I left a few comments, but not all of them are blockers: I think the biggest priorities from my POV are:

- Reasonably not-wrong docstrings for any code belonging to a public API (and if it's difficult to describe them that's probably a sign there's something bugging you about the design)
- Not-wrong typehints for new code.  (For example, very few of these methods should work on arbitrary `dict` arguments, we can usually say a lot more about what kind of dict it should be and let the type system work for us.  But for now I think we should at least have something down for each arg and return value, so `dict` is better than nothing)
- Breaking up long methods by refactoring them into consistent levels of abstraction.   Yurtsev's Rule states: "if a method is bigger than my head, it cannot fit inside my head" :)

(lmk if easier to discuss further offline)

poneill (changes requested):
Hey, this looks like a ton of great work.  I left a few comments, but not all of them are blockers: I think the biggest priorities from my POV are:

- Reasonably not-wrong docstrings for any code belonging to a public API (and if it's difficult to describe them that's probably a sign there's something bugging you about the design)
- Not-wrong typehints for new code.  (For example, very few of these methods should work on arbitrary `dict` arguments, we can usually say a lot more about what kind of dict it should be and let the type system work for us.  But for now I think we should at least have something down for each arg and return value, so `dict` is better than nothing)
- Breaking up long methods by refactoring them into consistent levels of abstraction.   Yurtsev's Rule states: "if a method is bigger than my head, it cannot fit inside my head" :)

(lmk if easier to discuss further offline)

search2/requirements.txt
poneill: I'm probably missing something here-- can we not build the CI/CD image without them if they're not currently required?  My priorities would be:

1. Having CI up and running
2. Having the requirements.txt faithfully reflect the current requirements.

search2/python/search/index/handler.py
```diff
@@ -35,11 +46,16 @@ def __init__(self, search_client_args, progress_tracker, chunk_size=1_000, repor
         self.counter = 0
 
     async def run(self, data):
-        resp = await async_helpers.async_bulk(self.client, iter(data), chunk_size=self.chunk_size)
+        """TODO: Add description here"""
+        resp = await async_helpers.async_bulk(
+            self.client, iter(data), chunk_size=self.chunk_size
```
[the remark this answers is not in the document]
akotlar: converts an iterable to an iterator; async_bulk requires iterators

search2/python/search/index/handler.py
```diff
@@ -50,30 +66,34 @@ async def close(self):
 
 @ray.remote(num_cpus=0)
 class ProgressReporter:
-    def __init__(self, publisher: dict):
+    """A class to report progress to a beanstalk queue"""
+
+    def __init__(self, publisher: Publisher):
         self.value = 0
```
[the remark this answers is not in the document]
akotlar: There is no way to access that value directly, afaik, it is defacto private

search2/python/search/index/handler.py
```diff
@@ -86,17 +106,17 @@ def get_counter(self):
...
-        tar_path: str,
-        annotation_conf: dict,
-        mapping_conf: dict,
-        search_conf: dict,
-        chunk_size=500,
-        paralleleism_chunk_size=5_000,
-        publisher: dict = None,
-        annotation_path: str = None
+    index_name: str,
+    mapping_conf: dict,
```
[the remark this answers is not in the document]
akotlar: It cannot be an arbitrary dict, but it is a dict. At this stage expanding this to TypedDict is outside the scope of this PR

search2/python/search/save/handler.py
akotlar: the reporter reports values to the frontend via beanstalkd

search2/python/search/save/handler.py
akotlar: That can be a future enhancement

search2/python/search/save/handler.py
akotlar: See https://opensearch.org/docs/2.3/clients/python-low-level/

search2/python/search/save/handler.py
akotlar: We've discussed this,

search2/python/search/save/listener.py
akotlar: pylint doesn't like single letter names. single letter names are preferable where they are locally obvious

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,155 @@
...
+    handler_fn: Callable,
+    submit_msg_fn: Callable,
+    completed_msg_fn: Callable,
+    queue_conf: QueueConf,
+    tube: str,
+):
+    """TODO: Listen on beanstalkd here directly"""
+    hosts, ports = queue_conf.split_host_port()
+
+    for event in ("progress", "failed", "started", "completed"):
```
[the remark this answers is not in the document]
akotlar: Tuples are for immutable data.

search2/python/search/utils/beanstalkd.py
akotlar: because this is an immutable collection

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,155 @@
...
+            client.put_job_into(tube_conf["events"], dumps(msg))
+            res = asyncio.get_event_loop().run_until_complete(handler_fn(publisher, job_data))
+
+            base_msg["event"] = events_conf["completed"]
+            msg = completed_msg_fn(base_msg, job_data, res)
+
+            client.put_job_into(tube_conf["events"], dumps(msg))
+            client.delete_job(job.job_id)
+        except BeanstalkError as err:
+            print(f"Received error during execution: {err}")
```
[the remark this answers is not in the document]
akotlar: We can do this in a future PR if desired. This is being run by pm2, and all print statements are captured and logged

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,155 @@
...
+        host = clients[offset][0]
+        port = clients[offset][1]
+        client = clients[offset][2]
+
+        client.watch(tube_conf["submission"])
+
+        try:
+            job = client.reserve_job(5)
+        except BeanstalkError as err:
+            if err.message == "TIMED_OUT":
```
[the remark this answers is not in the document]
akotlar: https://github.com/beanstalkd/beanstalkd/blob/master/doc/protocol.txt is the place where this is defined

search2/python/search/index/handler.py
```diff
@@ -86,17 +106,17 @@ def get_counter(self):
...
+    search_conf: dict,
+    chunk_size=500,
+    paralleleism_chunk_size=5_000,
+    publisher: Optional[Publisher] = None,
+    tar_path: Optional[str] = None,
+    annotation_path: Optional[str] = None,
 ):
     assert not (tar_path is not None and annotation_path is not None)
+    assert not (tar_path is None and annotation_path is None)
```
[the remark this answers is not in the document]
akotlar: > But if so, why is tar_path allowed to be None in the constructor, or annotation_path allowed to be non-none?

This is covered in a TODO at the top of the module.

The rest of your comment has been addressed

search2/python/search/index/listener.py
```diff
@@ -190,15 +26,53 @@ def main():
...
+    conf_dir = args.conf_dir
     with open(args.queue_conf, "r", encoding="utf-8") as queue_config_file:
         queue_conf = YAML(typ="safe").load(queue_config_file)
 
     with open(args.search_conf, "r", encoding="utf-8") as search_config_file:
         search_conf = YAML(typ="safe").load(search_config_file)
 
-    listen(queue_conf, search_conf, config_path_base_dir)
+    def handler_fn(publisher: Publisher, job_details: dict):
+        m_path = get_config_file_path(conf_dir, job_details["assembly"], '.mapping.y*ml')
```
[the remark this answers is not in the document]
akotlar: is this not obvious from context?

akotlar:
I've addressed the comments

search2/python/search/utils/beanstalkd.py
akotlar: We should discuss expectations for docstrings. They are not needed for many of these small, descriptively named methods. Pylint requires them, and I propose we drop that requirement at the linter level.

search2/python/search/save/handler.py
akotlar: It signals intent. The value is boolean logically but can be numerical. int() doesn't signal that.

search2/python/search/save/listener.py
akotlar: What is the business value in doing this for a single use with hardcoded required, optional keys? That config class would almost certainly just replicate this, and we would use it once.

search2/python/search/utils/beanstalkd.py
poneill: gotcha, asking because of this distinction: https://stackoverflow.com/questions/24854139/lists-are-for-homogeneous-data-and-tuples-are-for-heterogeneous-data-why

Not asking for changes to the code but maybe worth a comment

search2/python/search/utils/beanstalkd.py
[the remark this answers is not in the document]
poneill: The link in the other comment explains the distinction in more detail but it's more of a convention than a rule so I'm happy to leave it here.

search2/python/search/utils/beanstalkd.py
poneill: I agree that the English-language meanings of the class names are clear.  I don't think it's as clear from the names what they represent in the context of this code.  For example, Publisher has a host, port, queue and message.  Host and port are clearly enough an address, but what does the queue str represent?  Why does every Publisher have its own Message?  I offer these as examples of questions a good-faith reader might have when reading this code.

It might make sense to disable the docstring requirement for private methods, but would recommend against dropping it for public-facing methods.  Happy to discuss further offline.

search2/python/search/index/listener.py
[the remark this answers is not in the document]
poneill: Would argue for `mapping_path` in that case :)  but am not blocking on it

search2/python/search/save/listener.py
poneill: Yeah,  it's possible its lifetime isn't long enough to justify the lift here.

poneill (approved):
lgtm, thanks for all of the work in review

akotlar:
> lgtm, thanks for all of the work in review

Thanks for the detailed review!

PR #105, opened 2023-05-10
Merge B10 v2

akotlar:
CHANGELOG:
* Merge b10 branch

This is ready to merge. All tests pass, and attempt to run bin/bystro-utils.pl to fetch files and SQL expressions completed successfully. Attempt to run bin/bystro-build.pl also running successfully, currently processing CADD which will take a while.

Passing test suite:

```
All tests successful.
Files=35, Tests=436960, 47 wallclock secs ( 8.43 usr  0.53 sys + 16.28 cusr  1.08 csys = 26.32 CPU)
Result: PASS
```

To verify run `prove -r t` from /bystro


lib/SeqFromQuery.pm
```diff
@@ -41,17 +39,12 @@ extends 'Seq::Base';
...
-
 # Post-processing to run, before commiting the annotation
 has pipeline => (is => 'ro', isa => 'ArrayRef[HashRef]');
 
-# Probably the user id
-has indexName => (is => 'ro', required => 1);
+has inputQueryBody => (is => 'ro', isa => 'HashRef', required => 1);
 
-# The index type; probably the job id
-has indexType => (is => 'ro', required => 1);
```
akotlar, line 54: Recent elastic search versions don't support the type parameter on indices

lib/SeqElastic.pm
```diff
@@ -151,6 +151,8 @@ sub go {
   
   my $fileSize = -s $filePath;
 
+  my $fileSize = -s $filePath;
```
akotlar, line 154: This library should be deleted; obviated by search/index; the corresponding Elasticsearch 5 libraries should also be removed from install-perl-libs.sh

lib/SeqFromQuery.pm
```diff
@@ -31,8 +31,6 @@ use Statistics::Distributions qw(chisqrdistr udistr);
 
 use Math::Round qw/nhimult round/;
 
-use POSIX qw/lround/;
```
akotlar, line 34: This library should be deleted; obviated by search/save; the corresponding Elasticsearch 5 libraries should also be removed from install-perl-libs.sh

akotlar: Also need to make sure we support the additional delimiter in search/save

lib/SeqElastic.pm
akotlar: Also need to make sure we support the additional delimiter in search/index

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,199 @@
...
+    completed_msg_fn: Callable,
+    queue_conf: QueueConf,
+    tube: str,
+):
+    """Listen on a Beanstalkd channel, waiting for work.
+       When work is available call the work handler
+    """
+    hosts, ports = queue_conf.split_host_port()
+
+    for event in ("progress", "failed", "started", "completed"):
```
poneill, line 85: we can push this up into the QueueConf constructor so that it's impossible to even instantiate an invalid one.

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,199 @@
...
+                event_queue=tube_conf["events"]
+            )
+
+            base_msg = {
+                "event": events_conf["started"],
+                "queueID": job.job_id,
+                "submissionID": job_data["submissionID"],
+            }
+
+            msg = submit_msg_fn(base_msg, job_data)
```
poneill, line 124: I didn't really understand what's going on here: I would have expected this function to submit a message rather than package one up, and in some implementations of this interface it just passes the base_msg through.  Could we get some documentation about how the caller should implement `handler_fn, submit_msg_fn` and `completed_msg_fn`?

poneill: we especially want to document the async considerations-- afaict we'll time out if `handler_fn` is non-async.

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,199 @@
...
+        client.watch(tube_conf["submission"])
+
+        try:
+            job = client.reserve_job(5)
+        except BeanstalkError as err:
+            if err.message == BEANSTALK_ERR_TIMEOUT:
+                continue
+            raise err
+        try:
+            job_data = loads(job.job_data)
```
poneill, line 107: this should really be its own try-block.  we're deserializing json that came in over telnet and that could really be anything.  We can validate immediately here and bail if it's malformed, raising an informative exception.

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,199 @@
...
+                job_data['submissionID'],
+                job.job_id,
+                publisher_host=client.host,
+                publisher_port=client.port,
+                progress_event=events_conf["progress"],
+                event_queue=tube_conf["events"]
+            )
+
+            base_msg = {
+                "event": events_conf["started"],
```
poneill, line 119: our handling of event configuration is kind of confusing in general: we pick up the data in `events_conf` from the beanstalkd yaml but it's currently redundant because `events_conf["started"]` is just "started", and so on.  Is there a reason they have to be picked up from a config or could they just be an enum?

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,199 @@
...
+
+            base_msg["event"] = events_conf["completed"]
+            msg = completed_msg_fn(base_msg, job_data, res)
+
+            client.put_job_into(tube_conf["events"], dumps(msg))
+            client.delete_job(job.job_id)
+        except BeanstalkError as err:
+            print(f"Received error during execution: {err}")
+            client.release_job(job.job_id)
+        except NotFoundError as err:
```
poneill, line 137: where does this exception come from?  could we isolate the statement that generates it?

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,199 @@
...
+            client.put_job_into(tube_conf["events"], dumps(msg))
+            res = asyncio.get_event_loop().run_until_complete(handler_fn(publisher, job_data))
+
+            base_msg["event"] = events_conf["completed"]
+            msg = completed_msg_fn(base_msg, job_data, res)
+
+            client.put_job_into(tube_conf["events"], dumps(msg))
+            client.delete_job(job.job_id)
+        except BeanstalkError as err:
+            print(f"Received error during execution: {err}")
```
poneill, line 135: we should figure out a more comprehensive logging strategy at some point, because if you start the annotator with `pm2` afaict these print statemenets just fall into a void.

search2/python/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,199 @@
...
+            }
+
+            traceback.print_exc()
+            client.put_job_into(tube_conf["events"], dumps(msg))
+            client.delete_job(job.job_id)
+        except Exception as err:
+            traceback.print_exc()
+            client.release_job(job.job_id)
+        finally:
+            time.sleep(0.5)
```
poneill, line 152: as a general comment about json parsing, if we're sleeping at the end of this block we probably shouldn't worry too much about perf :)

orjson has the advantages of speed and the ability to handle more complex python datatypes, but it has the disadvantage of way worse error messages.  We're not operating in a regime where the speedup is going to be detectable (parsing is, generously, six orders of magnitude under the sleep time) and we're not passing tricky python objects back and forth, but we are going to have to deal with the error messages :)

search2/python/search/utils/opensearch.py
```diff
@@ -0,0 +1,20 @@
+"""Utility functions for Opensearch 2.*"""
+
+def gather_opensearch_args(search_conf: dict):
+    """
+    Return arguments required to instantiate OpenSearch Client
+    Supports both https and http protocols
+    When hosts are prefixed https://, http_auth, client_cert, client_key, ca_certs,
+         verify_certs, ssl_assert_hostname, and ssl_show_warn will be in effect
+    """
+    return dict(
```
poneill, line 10: this should probably be a dataclass

poneill (changes requested):
This looks like a ton of great work.  Main concerns at this point are:

- tests
- pulling out as much of the business logic in async / networky code as we can into testable methods
- tightening data-containing objects into dataclasses or similar, where reasonable, so we can push runtime errors back to static analysis errors, especially at ser/des boundaries
- documentation, especially of the beanstalkd interface, and docstrings for public methods

Left some comments on the perl code but see what you make of them and let's consider it all legacy and non-blocking unless you think it's worth addressing.

poneill (changes requested):
This looks like a ton of great work.  Main concerns at this point are:

- tests
- pulling out as much of the business logic in async / networky code as we can into testable methods
- tightening data-containing objects into dataclasses or similar, where reasonable, so we can push runtime errors back to static analysis errors, especially at ser/des boundaries
- documentation, especially of the beanstalkd interface, and docstrings for public methods

Left some comments on the perl code but see what you make of them and let's consider it all legacy and non-blocking unless you think it's worth addressing.

bin/bystro-server.pl
```diff
@@ -32,89 +29,50 @@
...
 my $queueConfig = $conf->{beanstalkd}{tubes}{$type};
 
 if(!$queueConfig) {
-  die "$type not recognized. Options are " . ( join(', ', @{keys %{$conf->{beanstalkd}{tubes}}} ) );
+  die "Queue config format not recognized. Options are " . ( join(', ', @{keys %{$conf->{beanstalkd}{tubes}}} ) );
 }
 
 my $beanstalk = Beanstalk::Client->new({
-  server    => $conf->{beanstalkd}{host} . ':' . $conf->{beanstalkd}{port},
+  server    => $conf->{beanstalkd}{addresses}[0],
```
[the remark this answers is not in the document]
akotlar: Yeah, beanstalkd doesn’t have a built in scalability concept, so to scale out you just need multiple servers. In the search PR, we added initial support for that. Unchanged here.

search2/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+use pyo3::prelude::*;
```
poneill, line 1: Do we need this and main.rs at the moment?  If we're including these could we comment what we plan to implement in rust and what we're going to expose to python?

search2/python/search/utils/beanstalkd.py
akotlar: can totally be an enum

search2/python/search/utils/beanstalkd.py
akotlar: my handling of exceptions here is incorrect, mixing exceptions from the handler and from beanstalkd; will fix, good question

search2/src/lib.rs
akotlar: Absolutely. Which option would you prefer; it may be a good exercise to move the save handlers business function (querying a chunk of data, and then writing the result to tsv chunk) in Rust. Small, easy, and a good exercise.

poneill: It's probably worth a separate discussion about how we want to handle the rust rewrite.  Is any rust code load-bearing in this diff?  If not, would prefer to pull it for now.

akotlar: In the comment you're replying to it would be load bearing, yes

akotlar: To be clear, I'm not talking about a rewrite, rather the introduction of functions into lib.rs. This module is intended to serve as a basic test for Rust integration, so ripping out lib.rs (which Maturin probably wants around anyway), would kind of defeat the purpose

poneill: ok, could we just add a comment then stating that it's a stub for future development?  lib.rs and main.rs are clearly stubs but the Cargo.toml has non-trivial dependencies, so a comment would help the reader to understand at a glance what the present state of things is.

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,27 @@
+name: Pylint
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.11"]
```
poneill, line 10: we have to roll back to 3.10 for now :/

bin/bystro-server.pl
```diff
@@ -223,23 +159,20 @@
...
-      # TODO: Improve error handling, this doesn't work reliably
-      if($err->{vars}{body}{status} && $err->{vars}{body}{status} == 400) {
-        $err = "Query failed to parse";
-      } else {
-        $err = "Issue handling query";
-      }
     }
 
-    $beanstalkEvents->put( { priority => 0, data => encode_json({
+    $err = "There was an issue, sorry!";
```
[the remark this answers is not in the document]
poneill: could we just add a #TODO for this?

config/opensearch.yml
```diff
@@ -0,0 +1,10 @@
+connection:
+  max_content_length: 209715200
```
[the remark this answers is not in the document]
poneill: I'm guessing this means 200MB but what's going on here?

install-apt.sh
```diff
@@ -16,7 +16,7 @@ fi
 . ~/.profile;
 
 # Perlbrew simplifies version management
-. ./install/install-perlbrew-linux.sh $INSTALL_DIR perl-5.28.0 ~/.profile;
+. ./install/install-perlbrew-linux.sh $INSTALL_DIR perl-5.30.1 ~/.profile;
```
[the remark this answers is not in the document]
poneill: closing on perl amnesty grounds

install-apt.sh
```diff
@@ -35,4 +35,4 @@ fi
 
 mkdir -p logs;
 
-printf "\n\nREMEMBER TO INCREASE ULIMIT ABOVE 1024 IF RUNNING MANY FORKS\n\n";
\ No newline at end of file
+printf "\n\nREMEMBER TO INCREASE ULIMIT ABOVE 1024 IF RUNNING MANY FORKS\n\nIF RUNNING 1st TIME RUN: `source $PERLBREW_ROOT/etc/bashrc";
```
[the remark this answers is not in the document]
poneill: closing on perl amnesty grounds

install/install-apt-deps.sh
```diff
@@ -1,30 +1,34 @@
 #!/usr/bin/env bash
 
-echo -e "\n\nInstalling Ubuntu/Debian (apt-get) dependencies\n"
+echo -e "\n\nInstalling Ubuntu/Debian (apt-get) dependencies\n";
```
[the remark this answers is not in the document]
poneill: amnesty

lib/Seq.pm
```diff
@@ -179,7 +180,8 @@ sub annotateFile {
     #my ($mce, $slurp_ref, $chunk_id) = @_;
     #    $_[0], $_[1],     $_[2]
     #open my $MEM_FH, '<', $slurp_ref; binmode $MEM_FH, ':raw';
```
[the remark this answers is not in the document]
poneill: perl amnesty

lib/Utils/Fetch.pm
```diff
@@ -133,17 +139,20 @@ sub _fetchFiles {
 
   my $fetchProgram;
 
-  my $isRsync;
+  my $isRsync = 0;
```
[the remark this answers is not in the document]
poneill: amnesty

python/python/bystro/search/utils/messages.py
```diff
@@ -0,0 +1,99 @@
...
+    def keys_with_types(cls) -> dict:
+        return get_type_hints(cls)
+
+
+class FailedJobMessage(Struct, frozen=True):
+    submissionID: SUBMISSION_ID
+    reason: str
+
+
+class InvalideJobMessage(Struct, frozen=True):
```
poneill, line 25: Typo

python/python/bystro/search/utils/messages.py
```diff
@@ -0,0 +1,99 @@
+from enum import Enum
+from typing import get_type_hints
+
+from msgspec import Struct, field
+
+from bystro.search.utils.annotation import AnnotationOutputs
+
+SUBMISSION_ID = str | int
```
poneill, line 8: Type names are CamelCased with an initial capital

config/beanstalk.clean.yml
```diff
@@ -1,6 +1,6 @@
 beanstalkd:
-  host: ~
-  port: ~
+  addresses:
```
poneill, line 2: non-blocking comment: at some point we should probably figure out how to encapsulate the issue of multiple beanstalk servers from the workers.  If this ever gets misconfigured (a worker is watching on one address but its counterpart is submitting on another) it's going to be tough to debug.

python/python/bystro/search/utils/beanstalkd.py
```diff
@@ -0,0 +1,233 @@
+"""TODO: Add description here"""
```
poneill, line 1: it seems that beanstalkd.py provides functionality intended to be shared by all python workers.  If so, it should go in its own top-level directory in the python section of this repo rather than live in `search` code.  Otherwise we'll have `ancestry` (and every other worker) rummaging around in `search` code for its imports.

poneill (approved):
We're amnestying a lot of this, so can we just:

- fix the typo in `InvalideJobMessage` before we bake it in, and
- pull out as much beanstalk functionality as you want to be shared between workers into its own directory outside of `search`?

config/beanstalk.clean.yml
akotlar: Good point. In that process we should also look at alternative message passing frameworks like Kafka that have high availability built in.

akotlar:
@poneill addressed your requests

https://github.com/akotlar/bystro/pull/105#discussion_r1220056050
- that is the maximum size in bytes of data that will be accepted by opensearch in a request. we want to be able to batch many row documents together into one request, since every http call incurs 10's of ms overhead.

config/beanstalk.clean.yml
poneill: sgtm, I'd be curious to hear your thoughts when we cross that bridge but would defer to your expertise on the choice

PR #106, opened 2023-05-25
fix python environment on arm64 mac os

akotlar:


poneill (approved):
lgtm, can install successfully

PR #118, opened 2023-06-07
Switch pylint to ruff, add pytest dep, break out requirements-dev.txt

poneill:
- Add pytest and separate requirements into requirements.txt (holding project deps) and requirements-dev.txt (holding deps necessary only for local dev)
- Configure CI job so that test runner always pulls linting tools from versions specified in requirements-dev.txt
- Sneak in a .gitignore in the config dir as a convenience for local dev

### DIFF
--- a/.initialize_conda_env.sh
+++ b/.initialize_conda_env.sh
@@ -18,5 +18,6 @@ fi
 
 echo -e "\n====                         Installing Python requirements                             ====\n"
 find . -name 'requirements.txt' -exec pip install -q -r {} \;
+echo -e "\n====                         Installing Python development requirements                 ====\n"
 find . -name 'requirements-dev.txt' -exec pip install -q -r {} \;
 echo -e "\n====                                       Done                                         ====\n"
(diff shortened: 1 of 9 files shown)


.initialize_conda_env.sh
```diff
@@ -18,4 +18,6 @@ fi
 
 echo -e "\n====                         Installing Python requirements                             ====\n"
 find . -name 'requirements.txt' -exec pip install -q -r {} \;
+echo -e "\n====                         Installing Python development requirements                 ====\n"
```
poneill, line 21: I'd assume we want conda to install dev requirements by default?

python/requirements-dev.txt
```diff
@@ -0,0 +1,5 @@
+black==23.3.0
+mypy==1.3.0
+pylint==2.17.4
```
poneill, line 3: I notice we're using pylint and ruff atm, for the record I'm all in favor of switching over to ruff-only

akotlar: Let's do it. I had left pylint because we hadn't discussed switching the github workflow. Could you update the workflow definition to lint with ruff, and drop pylint from requirements-dev.txt? Thanks!

akotlar (changes requested):
Switch to ruff in requirements-dev and github workflow, then good to go

akotlar (approved):

PR #122, opened 2023-06-08
Add ancestry record classes

poneill:
Add classes holding common shapes of data for ancestry worker, tests for them.


python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,158 @@
...
+#  the validator needs to be shared between classes. So it just lives
+#  here as a pure python function with a weird type signature in order
+#  to support what attr expects from its validators.
+
+
+def _vcf_validator(_self: object, _attribute: attr.Attribute, value: str) -> None:
+    if not isinstance(value, str):
+        err_msg = f"vcf_path must be of type str, got: {value} instead"
+        raise TypeError(err_msg)
+    if not value.endswith(".vcf"):
```
akotlar, line 25: This will fail with compressed vcfs. We can create a pre-computed regex of
```
regex = re.compile( r'vcf\.{0,1}\w*')

def _vcf_validator(_self: object, _attribute: attr.Attribute, value: str) -> None:
    if not isinstance(value, str):
        err_msg = f"vcf_path must be of type str, got: {value} instead"
        raise TypeError(err_msg)
    if not regex.fullmatch(value):
   ...
```

akotlar: We shouldn't mind accepting a file that doesn't end in .vcf as long as it is a vcf file. But let's wait on relaxing that until we have a use case, the extension check is simple.

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,158 @@
...
+        validator=[
+            instance_of(float),
+            attr.validators.ge(0.0),
+            attr.validators.le(1.0),
+        ]
+    )
+
+    def __attrs_post_init__(self) -> None:
+        """Ensure interval is well-formed."""
+        if not self.lower_bound <= self.upper_bound:
```
akotlar, line 59: mild preference for self.lower_bound > self.upper_bound

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,158 @@
+"""Classes for common shapes of data in ancestry."""
+
+import attr
```
akotlar, line 3: I think we want to import `attrs` not `attr`. The authors recommend it, though I don't get a good sense of why besides performance and sensible defaults.

Reference: https://www.attrs.org/en/stable/names.html

Instead of `attr.s` we would use `attrs.frozen`

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,158 @@
...
+    @results.validator
+    def _is_list_of_ancestry_results(
+        self, _attribute: attr.Attribute, values: list[AncestryResult]
+    ) -> None:
+        for i, value in enumerate(values):
+            if not isinstance(value, AncestryResult):
+                err_msg = f"Expecting list of AncestryResults, at position {i} got {value} instead"
+                raise TypeError(err_msg)
+
+    def __attrs_post_init__(self) -> None:
```
akotlar, line 151: since this also pertains to result validation (at least I think so, could we just fold into is_list_of_ancestry_results?

```
  @results.validator
    def _is_list_of_ancestry_results(
        self, _attribute: attr.Attribute, values: list[AncestryResult]
    ) -> None:
        unique_ids = set()
        for i, value in enumerate(values):
            if not isinstance(value, AncestryResult):
                err_msg = f"Expecting list of AncestryResults, at position {i} got {value} instead"
                raise TypeError(err_msg)
           if value.sample_id in unique_ids:
                raise ValueError(err_msg)
           unique_ids.add(value.sample_id)
```

python/python/bystro/ancestry/tests/test_ancestry_types.py
```diff
@@ -0,0 +1,215 @@
+"""Test ancestry_types.py."""
+import pytest
```
akotlar, line 2: Should we add a test to check that the frozen attrs are indeed frozen?

akotlar (changes requested):
One important change, the rest can be left as is if distracting. If we leave, let's make a tracking issue

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,158 @@
...
+#  the validator needs to be shared between classes. So it just lives
+#  here as a pure python function with a weird type signature in order
+#  to support what attr expects from its validators.
+
+
+def _vcf_validator(_self: object, _attribute: attrs.Attribute, value: str) -> None:
+    if not isinstance(value, str):
+        err_msg = f"vcf_path must be of type str, got: {value} instead"
+        raise TypeError(err_msg)
+    if "vcf" not in value:
```
poneill, line 25: @akotlar loosened up the validator according to my interpretation of the intent of the regex, lmk if this should rather be a list of predefined extensions?

(fyi testing for string inclusion is ~5x faster than the compiled regex in this case :))

python/python/bystro/ancestry/ancestry_types.py
poneill: yeah I think the grand strategy for doing type- vs value validation could be a style guide topic, but can do here.

python/python/bystro/ancestry/tests/test_ancestry_types.py
poneill: good catch!

python/python/bystro/ancestry/ancestry_types.py
akotlar: hah, man regex is slower than I thought

python/python/bystro/ancestry/tests/test_ancestry_types.py
```diff
@@ -53,11 +53,24 @@
 
 def test_AncestrySubmission():
     AncestrySubmission("foo.vcf")
+    AncestrySubmission("foo.vcf.gz")
+    AncestrySubmission("foo.vcf.gz")
```
akotlar, line 57: this is entered 2x

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,179 @@
+"""Classes for common shapes of data in ancestry."""
+import re
+from collections import Counter
+from collections.abc import Sequence
+from typing import TypeVar
+
+import attrs
+from attrs import field
```
cristinaetrv, line 8: Since you're importing attrs here but use attrs.field, maybe importing field could be skipped here? And then uses of field converted to attrs.field for consistency? Will mark uses of field without attrs below.

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,179 @@
...
+    ge(0.0),
+    le(1.0),
+]
+
+
+@attrs.frozen()
+class ProbabilityInterval:
+    """Represent an interval of probabilities."""
+
+    lower_bound: float = field(validator=unit_float_validator)
```
cristinaetrv, line 65:
```suggestion
    lower_bound: float = attrs.field(validator=unit_float_validator)
```

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,179 @@
...
+    upper_bound: float = field(
```
cristinaetrv, line 66:
```suggestion
    upper_bound: float = attrs.field(
```

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,179 @@
...
+    superpops: SuperpopVector = field(validator=instance_of(SuperpopVector))
+    missingness: float = field(validator=unit_float_validator)
+
+
+@attrs.frozen(kw_only=True)
+class AncestryResponse:
+    """An outgoing response from the ancestry worker."""
+
+    vcf_path: str = attrs.field(validator=_vcf_validator)
+    results: list[AncestryResult] = field(validator=instance_of(list))
```
cristinaetrv, line 157:
```suggestion
    results: list[AncestryResult] = attrs.field(validator=instance_of(list))
```

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,179 @@
...
+    EAS: ProbabilityInterval = attrs.field(validator=ProbIntValidator)
+    EUR: ProbabilityInterval = attrs.field(validator=ProbIntValidator)
+    SAS: ProbabilityInterval = attrs.field(validator=ProbIntValidator)
+
+
+@attrs.frozen(kw_only=True)
+class AncestryResult:
+    """An ancestry result from a sample."""
+
+    sample_id: str = field(validator=instance_of(str))
```
cristinaetrv, line 146:
```suggestion
    sample_id: str = attrs.field(validator=instance_of(str))
```

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,179 @@
...
+    populations: PopulationVector = field(validator=instance_of(PopulationVector))
```
cristinaetrv, line 147:
```suggestion
    populations: PopulationVector = attrs.field(validator=instance_of(PopulationVector))
```

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,179 @@
...
+    superpops: SuperpopVector = field(validator=instance_of(SuperpopVector))
```
cristinaetrv, line 148:
```suggestion
    superpops: SuperpopVector = attrs.field(validator=instance_of(SuperpopVector))
```

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -0,0 +1,179 @@
...
+    missingness: float = field(validator=unit_float_validator)
```
cristinaetrv, line 149:
```suggestion
    missingness: float = attrs.field(validator=unit_float_validator)
```

python/python/bystro/ancestry/tests/test_ancestry_types.py
poneill: tyty

python/python/bystro/ancestry/tests/test_ancestry_types.py
```diff
@@ -0,0 +1,276 @@
...
+from ..ancestry_types import (
+    AncestryResponse,
+    AncestryResult,
+    AncestrySubmission,
+    AttrValidationError,
+    PopulationVector,
+    ProbabilityInterval,
+    SuperpopVector,
+)
+
```
cristinaetrv, line 14: Could you add a brief overview of how these types are being tested at the beginning? I saw a little description under some of them but just to get a quick idea could be useful, or if you wanted to instead add a little description to every subtype for more clarity.

python/python/bystro/ancestry/tests/test_ancestry_types.py
```diff
@@ -0,0 +1,276 @@
...
+    with pytest.raises(AttrValidationError):
+        ProbabilityInterval(lower_bound=1, upper_bound=0)
+
+    prob_int = ProbabilityInterval(lower_bound=0.0, upper_bound=1.0)
+    with pytest.raises(FrozenInstanceError):
+        prob_int.lower_bound = 0.5  # type: ignore [misc]
+
+
+prob_int = ProbabilityInterval(lower_bound=0.0, upper_bound=1.0)
```
cristinaetrv, line 99: prob_int is defined right above as the same thing - ignore this if it's included here as well for a separate reason but otherwise one of them could be removed.

cristinaetrv (approved):
Overall lgtm!

python/python/bystro/ancestry/ancestry_types.py
poneill: 🐢

python/python/bystro/ancestry/tests/test_ancestry_types.py
poneill: I added some more expansive docstrings and renamed some of the tests for clarity.  Going to land this now so we can keep the ball rolling but pls lmk if you'd like additional docs

PR #129, opened 2023-06-09
add rough train code

poneill:
Asking for very light code review on this one because this is not the final form of the ancestry model-- I just want to get it on GH so @cristinaetrv can start iterating on it (and in case 🚌)

closes https://github.com/bystrogenomics/bystro/issues/127


python/python/bystro/ancestry/train_utils.py
```diff
@@ -0,0 +1,24 @@
+"""Utilities for ancestry model training."""
+
+from typing import Any
+
+import numpy as np
+
+
+def get_variant_ids_from_callset(callset: dict[str, Any]) -> np.ndarray:
```
akotlar, line 8: definitely not something to do in this PR, but I may think about making a custom Struct or attrs slotted class for `callset` instead of a dict.

akotlar (approved):

akotlar:
Thanks for the PR! Let's get this in as is, to make sure you aren't blocked. I'll provide a more complete review when you feel it is time.

python/python/bystro/ancestry/train.py
```diff
@@ -0,0 +1,480 @@
...
+from sklearn.metrics import accuracy_score
+from sklearn.model_selection import train_test_split
+
+from bystro.ancestry.asserts import assert_equals, assert_true
+from bystro.ancestry.train_utils import get_variant_ids_from_callset
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+DATA_ROOT_DIR = Path.home() / "emory/human_ancestry_project/data/1000_genome_VCF"
```
akotlar, line 36: In the final version, are you planning on specifying an acceptable project structure, or else config file that you read these paths / parameters from?

poneill: yeah this is definitely not the way we want to do this in prod.  Given that this is a training job we'll probably end up passing it in as a command line arg.

PR #131, opened 2023-06-13
[ancestry] add ancestry infra code

### LINKED ISSUE #126: Beanstalk listener and handler
Set up listener and handler to integrate ancestry worker into beanstalk framework.

poneill:
Block out nonce helper functions (handler_fn, submit_msg_fn, completed_msg_fn) for beanstalk listener.  Once we're happy with these, we can do live model inference inside handler_fn.

(It may seem like this PR is not doing much, but it will allow the ancestry worker to run end-to-end while we're revising the model in parallel.)

Closes #126


python/python/bystro/ancestry/listener.py
```diff
@@ -0,0 +1,105 @@
...
+) -> AncestryResponse:
+    """Do ancestry job, wraping _infer_ancestry for beanstalk."""
+    # Separating _handler_fn from _infer_ancestry in order to separate ML from infra concerns,
+    # and especially to keep _infer_ancestry eager.
+    return _infer_ancestry(ancestry_job_data.ancestry_submission, publisher)
+
+
+def submit_msg_fn(ancestry_job_data: AncestryJobData) -> SubmittedJobMessage:
+    """Acknowledge receipt of AncestryJobData."""
+    return SubmittedJobMessage(ancestry_job_data.submissionID)
```
poneill, line 61: Here and throughout, are we handling submissionIDs correctly?  Would there be any way to make that harder to mess up?

python/python/bystro/ancestry/listener.py
```diff
@@ -0,0 +1,105 @@
...
+            "doesn't match response filename {ancestry_response.vcf_path}: this is a bug."
+        )
+        raise AssertionError(err_msg)
+    logger.debug("completed ancestry inference for: %s", ancestry_response)
+    return AncestryJobCompleteMessage(
+        submissionID=ancestry_job_data.submissionID, results=ancestry_response
+    )
+
+
+def main(queue_conf: QueueConf, ancestry_tube: str) -> None:
```
poneill, line 81: This function is a sort of legacy of a failed attempt at integration testing.  We can take it out, but we should also sort out the story about integration testing python workers.

Right now we're unit testing the three functions that `listen` requires, but it would be good to test the whole thing end to end.

One way might be to refactor `listen` so that we can dependency-inject a beanstalk client (and hence a mock for it in unit testing).

python/python/bystro/ancestry/listener.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def _load_queue_conf(queue_conf_path: str) -> QueueConf:
+    with Path(queue_conf_path).open(encoding="utf-8") as queue_config_file:
+        raw_queue_conf = YAML(typ="safe").load(queue_config_file)
+    beanstalk_conf = raw_queue_conf["beanstalkd"]
+    return QueueConf(addresses=beanstalk_conf["addresses"], tubes=beanstalk_conf["tubes"])
+
+
+def _infer_ancestry(
```
akotlar, line 36: I would probably do the business logic in Ray, so that we can scale out to N machines

python/python/bystro/ancestry/listener.py
```diff
@@ -0,0 +1,105 @@
...
+    logger.debug("Inferring ancestry for: %s", ancestry_submission)
+
+    # not doing anything with this reporter at the moment, we're
+    # simply threading it through for later.
+    _reporter = get_progress_reporter(_publisher)
+    vcf_path = ancestry_submission.vcf_path
+    return AncestryResponse(vcf_path=vcf_path, results=[])
+
+
+async def handler_fn(
```
akotlar, line 50: This doesn't need to be an async function, necessarily, but feel free to leave it. We may need to update beanstalkd to handle non-async functions.

python/requirements.txt
```diff
@@ -1,4 +1,6 @@
 Cython==0.29.35
+attrs==23.1.0
+cattrs==23.1.2
```
akotlar, line 3: Do we need cattrs?

akotlar (approved):

python/python/bystro/ancestry/listener.py
```diff
@@ -0,0 +1,105 @@
...
+    """Acknowledge receipt of AncestryJobData."""
+    return SubmittedJobMessage(ancestry_job_data.submissionID)
+
+
+def completed_msg_fn(
+    ancestry_job_data: AncestryJobData, ancestry_response: AncestryResponse
+) -> AncestryJobCompleteMessage:
+    """Send job complete message."""
+    ancestry_submission = ancestry_job_data.ancestry_submission
+    if ancestry_submission.vcf_path != ancestry_response.vcf_path:
```
akotlar, line 69: Don't we already validate this in the attrs model?

python/python/bystro/ancestry/listener.py
```diff
@@ -0,0 +1,105 @@
...
+        err_msg = (
+            f"Ancestry submission filename {ancestry_submission.vcf_path} "
+            "doesn't match response filename {ancestry_response.vcf_path}: this is a bug."
+        )
+        raise AssertionError(err_msg)
```
akotlar, line 74: probably a ValueError?

python/python/bystro/ancestry/listener.py
akotlar: With Ray, the reason to have async is if the main function is doing a bunch of non-blocking work.

akotlar:
None of my comments are blocking

python/python/bystro/ancestry/listener.py
poneill: ok, IIUC that sounds like a task for the next PR unless there's anything currently actionable?

python/python/bystro/ancestry/listener.py
poneill: Yeah I'd love to make this an eager function if we can get away with it.  When I remove the async, though, I get the following error when running the script and handling a message:

```
Traceback (most recent call last):
  File "/Users/patrickoneil/bystro/python/python/bystro/beanstalkd/worker.py", line 153, in listen
    res = asyncio.get_event_loop().run_until_complete(handler_fn(publisher, job_data))
  File "/opt/homebrew/Cellar/python@3.10/3.10.11/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 628, in run_until_complete
    future = tasks.ensure_future(future, loop=self)
  File "/opt/homebrew/Cellar/python@3.10/3.10.11/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/tasks.py", line 615, in ensure_future
    return _ensure_future(coro_or_future, loop=loop)
  File "/opt/homebrew/Cellar/python@3.10/3.10.11/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/tasks.py", line 630, in _ensure_future
    raise TypeError('An asyncio.Future, a coroutine or an awaitable '
TypeError: An asyncio.Future, a coroutine or an awaitable is required
```

Is there a way to use the existing `listen` API with an eager handler_fn?

python/python/bystro/ancestry/listener.py
poneill: Ok, I just wanted to be sure the worker was just supposed to return the original submissionID, and wasn't responsible for altering / replacing it in any way.

At some point it would be a big help to have some docs on submissionID / queueID semantics for future reference, maybe in the beanstalkd directory.

As an aside, I do think that long-term we should be running with `mypy --strict` tbh, or at least a stricter config containing the rules we think give us most bang for our buck, including `disallow_untyped_def`, but I guess that's out of scope here.

python/python/bystro/ancestry/listener.py
poneill: ah yeah these used to be private methods but now they're public so it's technically a ValueError, ty.  It's sort of an edge case whether the listener should have any public methods at all, but given that they're interfacing with `listen` it's probably best to think of them as public.

python/requirements.txt
poneill: nope thanks, not in this one, it was a holdover from the next PR

Issue #132, opened 2023-06-14
enforce unit tests on CI


poneill:
looks like we're not running unit tests on CI atm, we should make them blocking in the CI runner.

PR #133, opened 2023-06-15
make beanstalkd.worker expect sync functions

akotlar:


Makefile
```diff
@@ -7,4 +7,4 @@ develop:
 	cd python && maturin develop && cd ../
 
 serve-dev: develop
-	pm2 delete all 2> /dev/null && pm2 start startup.yml
+	pm2 delete all 2> /dev/null || true && pm2 start startup.yml
```
akotlar, line 10: to prevent error when nothing is running

config/dm6.mapping.yml
```diff
@@ -16,8 +16,6 @@ sort:
       number_of_replicas: 0
       number_of_shards: 6
       codec: best_compression
-      mapping:
```
akotlar, line 19: single_type was removed in recent elastic search versions

python/python/bystro/search/utils/annotation.py
```diff
@@ -39,7 +39,6 @@ class AnnotationOutputs(Struct, frozen=True):
             statistics: Optional[StatisticsOutputs]
                 Basenames of the statistics files, inside the archive
     """
-    output_dir: str
```
akotlar, line 42: output_dir on this caused issues during indexing because the bakend does not pass this property, and since we don't need it elsewhere, removing.

poneill:
The main thrust of this PR looks good to me, though I would ask to rename `go`.  

As for the changes unrelated to the async issue, are they strictly necessary to go in on this PR?  If so, we should probably give a brief explanation in the final git log message for the squashed commit.

But if they're not necessary for the async fix, can we put them in on a separate PR? (I haven't been perfect about this by any means but hope to get better at it...)

And if we changed the ES version recently and needed to update the yamls to comply with it, is there a test we could add to catch that sort of thing earlier?

(Just commenting here and deferring to @cristinaetrv on this one for approval)

poneill:
This change will break ancestry, so that will need to be updated too.  We'll be able to catch these things more easily after addressing #132

cristinaetrv:
> The main thrust of this PR looks good to me, though I would ask to rename `go`.

I agree w/ Pat that 'go' should be changed to something a little more specific especially because there seem to be two versions and now one uses async the other one doesn't (search/save/handler.py and search/index/handler.py). 

Is there a place where we're internally documenting what is using async and what isn't at the moment? I think that would make it easier to think about the workflows.

cristinaetrv:
> I agree with you guys. Will change go to something else. What should we name those?

Maybe something like handler_save handler_index if I'm understanding correctly that 'go' serves as the handler function?

poneill:
> > I agree with you guys. Will change go to something else. What should we name those?
> 
> Maybe something like handler_save handler_index if I'm understanding correctly that 'go' serves as the handler function?

IME, if we're having trouble naming a function it's a good sign that it has too many responsibilities and should be refactored.  But for now, maybe something like `run_query_and_save_output` would work?

akotlar:
> run_query_and_save_output

Sure.

akotlar:
Merging, will address comments/suggestions in followup PR

PR #134, opened 2023-06-15
Run tests during CI

### LINKED ISSUE #132: enforce unit tests on CI

poneill:
Add a pytest job to lint.yml to ensure that we're running unit tests during CI.

Closes #132

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,12 +21,13 @@ jobs:
         pip install -r **/requirements-dev.txt
         echo "Final dependency list:"
         pip freeze
-    - name: Linting with ruff
-      run: |
-        ruff $(git ls-files '*.py')
-    - name: Static analysis with mypy
-      run: |
-        mypy --config-file python/mypy.ini $(git ls-files '*.py')
+    # - name: Linting with ruff
+    #   run: |
+    #     ruff $(git ls-files '*.py')
+    # - name: Static analysis with mypy
+    #   run: |
+    #     mypy --config-file python/mypy.ini $(git ls-files '*.py')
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -16,14 +16,22 @@ jobs:
...
+        python -m venv ~/.venv/ci-venv
+        source ~/.venv/ci-venv/bin/activate
         python -m pip install --upgrade pip
         pip install -r **/requirements.txt
         pip install -r **/requirements-dev.txt
-        echo "Final dependency list:"
-        pip freeze
     - name: Linting with ruff
       run: |
+        source ~/.venv/ci-venv/bin/activate
```
poneill, line 26: having to reactivate the venv on each step is not so nice, lmk if there's a better way to do this with github actions?

python/requirements-dev.txt
```diff
@@ -2,5 +2,7 @@ black==23.3.0
 mypy==1.3.0
 pandas-stubs==2.0.2.230605
 pytest==7.3.1
+pytest-cov==4.1.0
```
poneill, line 5: These deps were missing before because tests weren't running in CI, so we add them here

.github/workflows/lint.yml
akotlar: Hm, I'm surprised Maturin needs venv (I run without venv locally; use conda). Let me check

akotlar:
let's get this in for now, can always adjust/improve

akotlar (approved):

PR #135, opened 2023-06-16
Adds rough train code to incorporate gnomad PCs

cristinaetrv:
Adds code to use PCs generated from 36,882 gnomad samples instead of generating PCs from 1KGP alone for ML model to existing train code

### DIFF
--- a/python/python/bystro/ancestry/train.py
+++ b/python/python/bystro/ancestry/train.py
@@ -37,6 +37,11 @@ DATA_ROOT_DIR = Path.home() / "emory/human_ancestry_project/data/1000_genome_VCF
 VCF_PATH = DATA_ROOT_DIR / "1KGP_final_variants_1percent.vcf.gz"
 ANCESTRY_MODEL_PRODUCTS_DIR = Path("ancestry_model_products")
 
+#Loads in gnomad PC loadings
+loadings=pd.read_csv('gnomadloadings.tsv',sep='\t')
+#Temporary placeholder to mark file used for testing, will load internally in future
+TKGP_VCF = DATA_ROOT_DIR / '1kgpGnomadList.vcf'
+
 ANCESTRY_INFO_PATH = DATA_ROOT_DIR / "20130606_sample_info.txt"
 ROWS, COLS = 0, 1
 QUALITY_CUTOFF = 100  # for variant quality filtering
@@ -306,6 +311,64 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
         pi = np.mean(gs) / PLOIDY
         total += ci * pi * (1 - pi)
     return (p * (1 - p) - total) / (p * (1 - p))
+
+
+def load_1kgp_vcf(TKGP_VCF):
+    """This is a temporary placeholder method to load in 1kgp vcf 
+    filtered down to the same variants as the gnomad loadings for testing
+    using plink2"""
+    headervcf=pd.read_csv(TKGP_VCF,delimiter='\t',comment='#')
+    dosagevcf = headervcf.replace('0|0', 0)
+    dosagevcf = dosagevcf.replace('0|1', 1)
+    dosagevcf = dosagevcf.replace('1|0', 1)
+    dosagevcf = dosagevcf.replace('1|1', 2)
+    dosagevcf=dosagevcf.rename(columns={"#CHROM": "Chromosome", "POS": "Position"}, errors="raise")
+    return dosagevcf
+    
+
+def load_pca_loadings(loadings, dosage_vcf):
+    """This loads in the gnomad PCs, reformats, and filters them to match
+    the format and variant list of the 1kgp vcf"""
+    loadings[['Chromosome', 'Position']] = loadings['locus'].str.split(':', expand=True)
+    #Match variant format with vcf
+    loadings['variant']=loadings.apply(lambda x: ':'.join([x['locus'][3:],x['alleles'][2],x['alleles'][6]]),axis=1)
+    #Remove variants that are missing from IGSR version of 1kgp
+    missing_variants = set(loadings.variant) - set(dosagevcf.ID)
+    loadings=loadings[~loadings['variant'].isin(missing_variants)]
+    #Remove brackets
+    loadings['loadings'] = loadings['loadings'].str[1:]
+    loadings['loadings'] = loadings['loadings'].str[:-1]
+    #Split PCs and join back
+    gnomadPCs = loadings['loadings'].str.split(',', expand=True)
+    loadings=loadings.join(gnomadPCs)
+    loadings=loadings.reset_index()
+    return loadings
+
+
+def apply_pca_transform(loadings, dosagevcf):
+    """Transforms vcf with genotypes in dosage format with PCs loadings from gnomad PCA"""
+    #Both files need to be np arrays for pca transformation
+    pc_loadings = loadings.iloc[:, 8:28]
+    genos = dosagevcf.iloc[:, 9:]
+    scaler = StandardScaler()
+    TGP_scaled = scaler.fit_transform(genos).T
+    TGP_scaled = np.array(TGP_scaled)
+    pc_loadings_array = pc_loadings.values.astype(float)
+    #Ensure the number of features matches the original dataset
+    if TGP_scaled.shape[1] != pc_loadings.shape[0]:
+        raise ValueError("Number of features in the new data doesn't match the number of features in the PCA loadings.")
+    #Scale 1kgp vcf
+    scaler = StandardScaler()
+    TGP_scaled = scaler.fit_transform(Genos).T
+    #Apply the loadings to the scaled 1kGP data
+    transformed_data = np.dot(TGP_scaled, pc_loadings_array)
+    #Add the IDs back on to PCs
+    IDs = genos.iloc[1]
+    IDs = IDs.T.reset_index()
+    IDsascol = IDs.iloc[:,0]
+    px = pd.DataFrame(transformed_data)
+    ids_with_pcs = pd.concat([IDs, px], axis = 1)
+    return ids_with_pcs
 
 
 def _perform_pca(train_X: pd.DataFrame, test_X: pd.DataFrame) -> tuple[np.ndarray, np.ndarray, PCA]:


cristinaetrv:
Still need to remove portions that may not be relevant after this change and decide on how to harmonize incoming vcf format for this step.

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,64 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    gnomadPCs = loadings['loadings'].str.split(',', expand=True)
+    loadings=loadings.join(gnomadPCs)
+    loadings=loadings.reset_index()
+    return loadings
+
+
+def apply_pca_transform(loadings, dosagevcf):
+    """Transforms vcf with genotypes in dosage format with PCs loadings from gnomad PCA"""
+    #Both files need to be np arrays for pca transformation
+    pc_loadings = loadings.iloc[:, 8:28]
```
poneill, line 351: do we need the rest of `loadings` elsewhere, or should this step go into `load_pca_loadings`?

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,64 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    genos = dosagevcf.iloc[:, 9:]
+    scaler = StandardScaler()
+    TGP_scaled = scaler.fit_transform(genos).T
+    TGP_scaled = np.array(TGP_scaled)
+    pc_loadings_array = pc_loadings.values.astype(float)
+    #Ensure the number of features matches the original dataset
+    if TGP_scaled.shape[1] != pc_loadings.shape[0]:
+        raise ValueError("Number of features in the new data doesn't match the number of features in the PCA loadings.")
+    #Scale 1kgp vcf
```
poneill, line 360: What's happening here?  A few questions:

- Should we be scaling at all?  If we're going to scale, we need to hang onto the scaler object as well so that we can apply it to new genotypes at inference time, otherwise train and inference data will be on the wrong scale.  Given that we're using tree-based classification downstream anyway (and so don't have to worry about the relative scales of our features), we should be ok just to drop it.
- It looks like we're scaling twice here
- `Genos` is undefined, so ruff should be warning about this locally.  lmk if you want a hand getting your local linting config set up

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,64 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    """Transforms vcf with genotypes in dosage format with PCs loadings from gnomad PCA"""
+    #Both files need to be np arrays for pca transformation
+    pc_loadings = loadings.iloc[:, 8:28]
+    genos = dosagevcf.iloc[:, 9:]
+    scaler = StandardScaler()
+    TGP_scaled = scaler.fit_transform(genos).T
+    TGP_scaled = np.array(TGP_scaled)
+    pc_loadings_array = pc_loadings.values.astype(float)
+    #Ensure the number of features matches the original dataset
+    if TGP_scaled.shape[1] != pc_loadings.shape[0]:
```
poneill, line 358: this is a good check but we should also guarantee that the order of variants in pc_loadings agrees with the order of variants in the TGP file.  (Otherwise we'd get a silent failure here.)

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,64 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+        raise ValueError("Number of features in the new data doesn't match the number of features in the PCA loadings.")
+    #Scale 1kgp vcf
+    scaler = StandardScaler()
+    TGP_scaled = scaler.fit_transform(Genos).T
+    #Apply the loadings to the scaled 1kGP data
+    transformed_data = np.dot(TGP_scaled, pc_loadings_array)
+    #Add the IDs back on to PCs
+    IDs = genos.iloc[1]
```
poneill, line 366: Much as with pie crust, the secret to manipulating dataframes (in my experience) is to do as little work as possible :)  If we have a genotype dataframe `genotypes` with dimensions (samples, variants) and a PCA loadings dataframe `loadings` with dimensions (variants, pc_dimensions), then we can just multiply them with the dot method of pd.DataFrame and get a type-checked matrix multiplication.  For example:

```
genotypes = pd.DataFrame(np.random.random(size=(5, 3)), columns=['variant1', 'variant2', 'variant3'], index=['alice', 'bob', 'carol', 'dave', 'eve'])
loadings = pd.DataFrame([[1, 2], [3, 4], [5, 6]], index=['variant1', 'variant2', 'variant3'], columns=['pc1', 'pc2'])
transformed_genotypes = genotypes.dot(loadings)

assert (transformed_genotypes.index == genotypes.index).all()
assert (transformed_genotypes.columns == loadings.columns).all()
```

The virtue of this approach is that if we try to multiply genotypes and loadings with mismatching sets of variants, we'll get a ValueError, and if we pass the right variants in the wrong order, `dot` will join them appropriately under the hood so that we get the right answer either way.

poneill (changes requested):
This is a good first cut, my main issue is just ensuring that we're always lining up variants correctly when we transform into PC space-- I sketched one idea for this in the comments.  

We should also make sure that we're checking in all the intermediate data products somewhere, along with code to generate them.

We should also test these functions on trivially-sized datasets.

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,64 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    dosagevcf=dosagevcf.rename(columns={"#CHROM": "Chromosome", "POS": "Position"}, errors="raise")
+    return dosagevcf
+    
+
+def load_pca_loadings(loadings, dosage_vcf):
+    """This loads in the gnomad PCs, reformats, and filters them to match
+    the format and variant list of the 1kgp vcf"""
+    loadings[['Chromosome', 'Position']] = loadings['locus'].str.split(':', expand=True)
+    #Match variant format with vcf
+    loadings['variant']=loadings.apply(lambda x: ':'.join([x['locus'][3:],x['alleles'][2],x['alleles'][6]]),axis=1)
```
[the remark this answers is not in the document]
cristinaetrv: Changed it to:
```
get_chr_pos= lambda x: x['locus'][3:]
get_allele1=  lambda x: x['alleles'][2]
get_allele2= lambda x: x['alleles'][6]
loadings['variant'] = loadings.apply(lambda x: ':'.join([get_chr_pos(x), get_allele1(x), get_allele2(x)]), axis=1)
```
Is this more of what you had in mind?

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,78 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
     return (p * (1 - p) - total) / (p * (1 - p))
 
 
+def load_1kgp_vcf(vcf_filepath: str) -> tuple[pd.DataFrame, pd.DataFrame]:
+    """Temporary placeholder method to load in 1kgp vcf
+    filtered down to the same variants as the gnomad loadings for testing
+    using plink2.
+    """
+    #This line will have to be altered if testing other vcfs
+    header_vcf=pd.read_csv(vcf_filepath,delimiter='\t',skiprows=107)
```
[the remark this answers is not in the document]
cristinaetrv: For the current version, comment='#' cuts out the header row which includes sample IDs but after final input file is decided, it should be a more generalizable version

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,78 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    #If testing other vcfs this line will also need to be altered
+    dosage_vcf=dosage_vcf.rename(
+        columns={"#CHROM": "Chromosome", "POS": "Position"}, 
+        errors="raise"
+    )
+    genos = dosage_vcf.iloc[:, 9:]
+    genos["variant"] = dosage_vcf["ID"]
+    genos=genos.set_index("variant")
+    genos=genos.sort_index()
+    return dosage_vcf, genos
```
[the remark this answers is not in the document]
cristinaetrv: Previously genos was processed in apply_pca_transform - main difference is that it removes all the non-genotype columns for the transformation, but the columns are used in processing the pca loadings in the next section.

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,78 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    
+
+def load_pca_loadings(loadings: pd.DataFrame, dosage_vcf: pd.DataFrame) -> pd.DataFrame:
+    """Load in the gnomad PCs, reformats, and filters them to match
+    the format and variant list of the 1kgp vcf.
+    """
+    loadings[["Chromosome", "Position"]] = loadings["locus"].str.split(":", expand=True)
```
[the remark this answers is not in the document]
cristinaetrv: Would you recommend renaming 'load_pca_loadings' to something like 'process_pca_loadings'? Or something even more specific? This function is more about processing the pca loadings to get them in the correct formatting for pca transformation so maybe that would make the most sense.

python/python/bystro/ancestry/train.py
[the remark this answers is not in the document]
poneill: the 1KGP vcf and the gnomad pca loadings are two conceptually separate objects-- we should probably find a way to load the gnomad loadings without changing the signature of `load_1kgp_vcf`.

If someone writes `kgp_vcf = load_1kgp_vcf()`, they're going to be surprised by what they get, namely that `kgp_vcf` will be the tuple: `(dosage_vcf, genos)`

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,78 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+
+
+def apply_pca_transform(pc_loadings: pd.DataFrame, genos: pd.DataFrame) -> pd.DataFrame:
+    """Transform vcf with genotypes in dosage format with PCs loadings from gnomad PCA."""
+    #Transpose genos so it is in the correct configuration for dot product
+    genos_transpose=genos.T
+    #Ensure that genos_transpose and pc_loadings have the same variants in same order
+    assert (genos_transpose.columns == pc_loadings.index).all()
+    assert genos_transpose.shape[1] == pc_loadings.shape[0]
+    #Convert to np for faster dot product
```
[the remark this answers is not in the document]
cristinaetrv: So far it's still much faster running it with numpy (less than a second compared to several minutes) so I think it's worth it.

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,78 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    genos_np = np.array(genos_transpose)
+    pc_loadings_array = pc_loadings.to_numpy.astype(float)
+    #Apply the loadings to the transposed 1kGP data
+    transformed_data = np.dot(genos_np, pc_loadings_array)
+    #Add the IDs back on to PCs
+    KGP_index = genos_transpose.index
+    transformed_data_with_ids = pd.DataFrame(transformed_data, index=KGP_index)
+    #Add PC labels
+    transformed_data_with_ids.columns = ["PC" + str(i) for i in range(1, 21)]
```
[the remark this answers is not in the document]
cristinaetrv: Will add the full set - I think I subsetted for testing but hadn't added them all back in yet.

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,78 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    #Split PCs and join back
+    gnomadPCs = loadings["loadings"].str.split(",", expand=True)
+    loadings=loadings.join(gnomadPCs)
+    loadings=loadings.reset_index()
+    pc_range = range(8, 28)
+    pc_loadings = loadings.iloc[:, pc_range].copy()
+    pc_loadings["variant"] = loadings["variant"]
+    pc_loadings=pc_loadings.set_index("variant")
+    pc_loadings=pc_loadings.sort_index()
+    return pc_loadings
```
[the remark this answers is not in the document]
cristinaetrv: Added intersection of sets step to assert we have the correct number of variants

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,88 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    dosage_vcf = dosage_vcf.replace("1|1", 2)
+    #If testing other vcfs this line will also need to be altered
+    dosage_vcf=dosage_vcf.rename(
+        columns={"#CHROM": "Chromosome", "POS": "Position"}, 
+        errors="raise"
+    )
+    return dosage_vcf
+    
+
+def load_pca_loadings(GNOMAD_PC_PATH: str, dosage_vcf: pd.DataFrame) -> pd.DataFrame:
```
poneill, line 335: it looks like `dosage_vcf` is no longer used in this function?  (fyi ruff can spot these things for you)

python/python/bystro/ancestry/train.py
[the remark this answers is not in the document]
poneill: is this comment still accurate?  Could we just assert that the shape of pc_loadings is what we expect?

(And also maybe a check that it's of dtype float, since we got bitten hard by that today?

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,88 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    dosage_vcf = header_vcf.replace("0|0", 0)
+    dosage_vcf = dosage_vcf.replace("0|1", 1)
+    dosage_vcf = dosage_vcf.replace("1|0", 1)
+    dosage_vcf = dosage_vcf.replace("1|1", 2)
+    #If testing other vcfs this line will also need to be altered
+    dosage_vcf=dosage_vcf.rename(
+        columns={"#CHROM": "Chromosome", "POS": "Position"}, 
+        errors="raise"
+    )
+    return dosage_vcf
```
poneill, line 332: What would you think about setting dosage_vcf so that its index is sample_ids and its columns are variants?  That could simplify the overlap calculation (see below)

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,88 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    loadings = loadings.reset_index()
+    pc_range = range(8, 38)
+    pc_loadings = loadings.iloc[:, pc_range].copy()
+    pc_loadings["variant"] = loadings["variant"]
+    pc_loadings = pc_loadings.set_index("variant")
+    pc_loadings = pc_loadings.sort_index()
+    return pc_loadings
+
+
+def loadings_var_overlap(pc_loadings: pd.DataFrame, dosage_vcf: pd.DataFrame) -> tuple[pd.DataFrame, pd.DataFrame]:
```
poneill, line 363: if we defined dosage_vcf so that its index is sample_ids and columns are variants, we could do something like the following here:
```
def loadings_var_overlap(pc_loadings: pd.DataFrame, dosage_vcf: pd.DataFrame) -> tuple[pd.DataFrame, pd.DataFrame]:
    common_variants = pc_loadings.index.intersection(dosage_vcf.columns)
    return pc_loadings.loc[common_variants]
```

wdyt?

poneill, line 363: I'm also having a little trouble following the name of this function, is there a good verb to start the function name with, e.g. `restrict_pc_loadings_to_variants` or something?

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,88 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    genos["variant"] = dosage_vcf["ID"]
+    genos=genos.set_index("variant")
+    genos=genos.sort_index()
+    #Transpose genos so it is in the correct configuration for dot product
+    genos_transpose=genos.T
+    #Ensure that genos_transpose and pc_loadings have the same variants in same order
+    assert (genos_transpose.columns == pc_loadings_overlap.index).all()
+    assert genos_transpose.shape[1] == pc_loadings_overlap.shape[0]
+    #Dot product
+    transformed_data = genos_transpose @ pc_loadings_overlap
```
poneill, line 389: if we preprocess the dfs a bit more upfront, the whole function could be this line, right?

poneill (changes requested):

python/python/bystro/ancestry/train.py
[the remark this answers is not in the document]
cristinaetrv: Yes but moved that section to the function below it - we check shape in the transformation function but float check here sounds like a good idea.

python/python/bystro/ancestry/train.py
cristinaetrv: The current format of dosage_vcf has variant IDs as rows and sample IDs in the header/columns, with additional informational columns. It's not until genos is processed and transposed that its index is sample IDs and columns are variant IDs. If I move genos processing to a separate function above this step (as you suggested below) that would make sense to do a genos_transpose columns/pc_loadings index intersection or maybe even easier conceptually genos/pc_loadings index intersection. Worth enforcing genos[overlapping_variants] as well in some step (maybe this one) since in the current version the dosage_vcf is pre-processed filtered on the gnomad variants.

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,93 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    genos = genos.sort_index()
+    return genos
+
+def restrict_loadings_variants_to_vcf(
+    pc_loadings: pd.DataFrame,
+    genos: pd.DataFrame
+) -> tuple[pd.DataFrame, pd.DataFrame]:
+    """Restrict variant list to overlap between gnomad loadings and reference vcf."""
+    #IGSR version of 1kgp is current reference vcf
+    loadings_var_set = set(pc_loadings.index)
```
poneill, line 377: afaict this is actually unnecessary because pandas Indexes have an intersection method already defined.

python/python/bystro/ancestry/train.py
```diff
@@ -308,6 +313,93 @@ def _calc_fst(variant_counts: pd.Series, samples: pd.DataFrame) -> float:
...
+    genos_var_set = set(genos.index)
+    var_overlap = loadings_var_set.intersection(genos_var_set)
+    pc_loadings_overlap = pc_loadings[pc_loadings.index.isin(var_overlap)]
+    genos_overlap = genos[genos.index.isin(var_overlap)]
+    #Ensure that genos transpose and pc_loadings have corresponding shape and vars
+    assert genos_overlap.T.shape[1] == pc_loadings_overlap.shape[0]
```
poneill, line 383: nit: `genos_overlap.T.shape[1]` is just `genos_overlap.shape[0]`, right?

poneill (approved):
lgtm, left two non-blocking comments about minor simplifications

poneill (changes requested):
sorry I just noticed that CI isn't running on this PR: is it still in draft mode?  

- Also the lint.yml is out of date so you'll want to merge / rebase master before merging this branch.

- let's also unit test this in this PR.

cristinaetrv:
Continued in #148 to incorporate CI changes

PR #137, opened 2023-06-17
add multivariate-tada folder

IlhaH:


contrib/multivariate-tada/genunivdata.py
```diff
@@ -0,0 +1,50 @@
+import pandas as pd
+import torch
+
+#Simulates univariate data based on demo mutation rates in original TADA code
+
+def genUnivData(gamma, q, N_family, N_case, N_control, mut_1, mut_2):
```
poneill, line 6: Prefer `snake_case` to `camelCase` for method and variable names: https://pep8.org/#function-names

poneill: Also, it's fine to use single character variables in order to maintain consistency with notation used in a paper, but if so we'll want to gloss them somewhere, probably in the module docstring.

poneill, line 6: Let's also add type annotations for this method

contrib/multivariate-tada/genunivdata.py
```diff
@@ -0,0 +1,50 @@
...
+    #Rates for poisson distributions that the counts follow
+
+    X_d1_rates = 2 * mut_1 * gamma * N_family
```
poneill, line 9: Here and throughout, let's pull out numerical constants into global constants so the reader can tell at a glance what `2` means in this context: is it ploidy or something else?

contrib/multivariate-tada/genunivdata.py
```diff
@@ -0,0 +1,50 @@
...
+    X_d2_rates = 2 * mut_2 * gamma * N_family
+    X_ctrl_rates = 5000 * [q * N_control]
+    X_case_rates = 5000 * [q * gamma * N_case]
+    X_t_rates = 5000 * [q * gamma * N_family]
+    X_nt_rates = 5000 * [q * N_family]
+
+    #Stacks poisson rates and samples
+    cls1 = torch.stack([X_d1_rates,torch.tensor(X_t_rates),torch.tensor(X_nt_rates),torch.tensor(X_case_rates),torch.tensor(X_ctrl_rates)])
```
poneill, line 17: what's happening here?  AFAICT this will fail unless `X_d1_rates` is a tensor,  which means that at least one of the quantities used to define it must also be a tensor.  Type annotations would help to clarify this for the reader

contrib/multivariate-tada/genunivdata.py
```diff
@@ -0,0 +1,50 @@
...
+    cls2 = torch.stack([X_d2_rates,torch.tensor(X_t_rates),torch.tensor(X_nt_rates),torch.tensor(X_case_rates),torch.tensor(X_ctrl_rates)])
+    sample_1 = torch.poisson(cls1)
+    sample_2 = torch.poisson(cls2)
+
+    #Combines simulated data with mutation rates into table
+    table = torch.transpose(torch.cat((mut_1.unsqueeze(0),sample_1,mut_2.unsqueeze(0),sample_2),dim=0),0,1)
```
poneill, line 23: What's happening here?  Do any of these intermediate expressions have domain-level interpretations that we could pull out and name?

contrib/multivariate-tada/genunivdata.py
poneill, line 6: This method needs a docstring too, preferably a multi-line one that glosses the parameters and return value :)

contrib/multivariate-tada/genunivdata.py
```diff
@@ -0,0 +1,50 @@
...
+    df.columns = ['mut.cls1','dn.cls1', 'trans.cls1', 'ntrans.cls1', 'case.cls1', 'ctrl.cls1', 'mut.cls2', 'dn.cls2', 'trans.cls2', 'ntrans.cls2', 'case.cls2', 'ctrl.cls2']
+    return df
+
+
+#Read mutation data
+tada_file = "TADA_demo_counts_de-novo_and_inherited.txt"
+tada_data = pd.read_table(tada_file)
+
+#Specify the number of families and the number of cases and control samples included in the simulation
+N_family = 4500
```
poneill, line 34: What exactly do the three `N_` values represent, and how do they relate to one another?  Are there any constraints we would expect to hold between them?

contrib/multivariate-tada/genunivdata.py
poneill, line 17: is `cls` an existing convention from the paper?  If so we should gloss it and if not we should probably choose something more descriptive

contrib/multivariate-tada/genunivdata.py
```diff
@@ -0,0 +1,50 @@
...
+
+#Extract mutation rates
+mut_cls1 = torch.tensor(tada_data["mut.cls1"])
+mut_cls2 = torch.tensor(tada_data["mut.cls2"])
+
+#Varying gamma and q parameters
+g_1,g_2,g_3 = 15, 20, 25
+q_1,q_2,q_3 = 5* 10 ** -5, 10**-4, 2*10**-4
+
+#Test gamma_1 and q_1
```
poneill, line 46: How do we plan to test this?  Could we implement that test in python and check it in along with this code?

contrib/multivariate-tada/genunivdata.py
```diff
@@ -0,0 +1,50 @@
...
+N_family = 4500
+N_case = 1000
+N_ctrl = 3000
+
+#Extract mutation rates
+mut_cls1 = torch.tensor(tada_data["mut.cls1"])
+mut_cls2 = torch.tensor(tada_data["mut.cls2"])
+
+#Varying gamma and q parameters
+g_1,g_2,g_3 = 15, 20, 25
```
poneill, line 43: if `gamma` is the varname we chose above, we should prefer `gamma` throughout, so `gamma1, gamma2, gamma3`.

poneill, line 43: also, not a big deal but the formatting is inconsistent between the left and right hand sides of this assignment.  You can fix that once and for all by running `black` on your code, and ideally configuring your editor to run it for you on save so you never have to think about it (the bystro black config is stored at bystro/python/pyproject.toml).

poneill (changes requested):
Hey Ilha, this looks like a great start.  My main ask is for some additional documentation of the function / purpose of this code and tests to ensure that it's working as intended.  Happy to chat offline about any part of it, though.

PR #141, opened 2023-06-28
add more generic dummy AncestryResponse to ancestry listener

poneill:
We were previously returning an empty AncestryResponse in the ancestry listener-- make the response more generic by returning AncestryResults for three dummy samples.


python/python/bystro/ancestry/listener.py
```diff
@@ -55,21 +60,21 @@ async def handler_fn(
     """Do ancestry job, wraping _infer_ancestry for beanstalk."""
     # Separating _handler_fn from _infer_ancestry in order to separate ML from infra concerns,
     # and especially to keep _infer_ancestry eager.
-    logger.debug("entering handler_fn with: {}", ancestry_job_data)
+    logger.debug("entering handler_fn with: %s", ancestry_job_data)
```
akotlar, line 63: at some point should we settle on f-string vs formatted strings?

dlin30 (approved):

python/python/bystro/ancestry/listener.py
poneill: `logging` requires `%s`, the previous commit was actually just a bug.  

These kinds of errors can be caught by ruff (https://beta.ruff.rs/docs/rules/logging-too-many-args/) and usually are: I run ruff locally with a stricter config than we currently enforce in CI.  But I was live-debugging something the other day and this slipped through.

The long-term fix I'd recommend is to enable (among other things) the `PLE` error group (https://beta.ruff.rs/docs/rules/#error-ple) in the ruff config.

PR #145, opened 2023-07-04
Fix lingering annotation & utility issues

akotlar:


lib/Seq.pm
```diff
@@ -134,15 +134,15 @@ sub annotateFile {
 
   # To avoid the Moose/Mouse accessor penalty, store reference to underlying data
   my $db = $self->{_db};
-  my $refTrackGetter = $self->tracksObj->getRefTrackGetter();
-  my @trackGettersExceptReference = @{$self->tracksObj->getTrackGettersExceptReference()};
+  my $refTrackGetter = $self->{_tracks}->getRefTrackGetter();
```
akotlar, line 137: $self->{_tracks} is initialized in the BUILD method, and is done so because tracksObj is lazy and its lazy init method sets the database to read only mode. We want that to happen as early in the program as possible, because otherwise LMDB will be in write mode, and we do not respect write transaction semantics in annotation (it must be read only). This gives us a large throughput benefit.

python/python/bystro/search/utils/annotation.py
```diff
@@ -94,7 +94,7 @@ def from_path(
     "field": "\t",
     "allele": "/",
     "position": "|",
-    "overlap": "\\",
+    "overlap": chr(31),
```
akotlar, line 97: overlap delimiter is used in master annotator, and this matches the Seq::Output::Delimiters value (it is a control character; we should discuss better ways of handling overlapping values and nested relationships in the future).

README.md
```diff
@@ -3,6 +3,7 @@
 TLDR; 1,000x+ faster than VEP, more complete annotation + online search (https://bystro.io) for datasets of up to 47TB (compressed) online, or petabytes offline.
```
akotlar, line 3: I've made some improvements, but lets save other improvements for a future PR

README.md
```diff
@@ -39,17 +40,115 @@ Please read [FIELDS.md](FIELDS.md)
...
+                GROUP_CONCAT(DISTINCT(NULLIF(x.spDisplayID, '')) SEPARATOR ';') FROM kgXref
+                x WHERE x.refseq=r.name) AS spDisplayID, (SELECT GROUP_CONCAT(DISTINCT(NULLIF(x.protAcc,
+                '')) SEPARATOR ';') FROM kgXref x WHERE x.refseq=r.name) AS protAcc, (SELECT
+                GROUP_CONCAT(DISTINCT(NULLIF(x.mRNA, '')) SEPARATOR ';') FROM kgXref x WHERE
+                x.refseq=r.name) AS mRNA, (SELECT GROUP_CONCAT(DISTINCT(NULLIF(x.rfamAcc,
+                '')) SEPARATOR ';') FROM kgXref x WHERE x.refseq=r.name) AS rfamAcc FROM
+                refGene r WHERE chrom=%chromosomes%;
+      ```
+
+      Running `bin/bystro-utils.pl --config <path/to/this/config> ` will results in the following config:
```
poneill, line 84: typo: will result

lib/Seq.pm
poneill: makes sense to me, but could this explanation (and the others on this PR giving context / clarifying intent) go into the code as comments?

python/python/bystro/search/utils/annotation.py
poneill: If this piece of configuration is shared between perl and python code, what would you think about pulling it out into a config file that both can draw from so they can't get out of sync?

Failing that, we should probably at least document that this is tied to a definition in the annotator.

akotlar: Will do in a followup PR. We can't afaik put non-rendered characters in yaml, but that may be the wrong control character to use. Will need to this more about this. I think this points out the need to move away from .tsv

PR #146, opened 2023-07-04
Cinch ruff config to strictest ruleset bystro currently satisfies

poneill:
Update ruff config to enforce the coding disciplines we're currently already observing.  (This requires no change to the codebase.)


akotlar (commented):
LGTM!

PR #150, opened 2023-07-14
Lay groundwork to add ancestry chip model training code

poneill:
Do some ground-clearing to prepare the next PR: training a new ancestry model on the intersection of the Illumina and Affymetrix callsets.  This PR is an admitted hodge-podge of:

- Directory reorganization
- Refactoring of VCF-loading code for performance
- Refactoring of various functions to improve testability
- Various other chores

There's more that could be done in this diff: refactoring `make_rfc` and increasing test coverage, for instance, but the PR's already getting a little long: happy to include those in this PR or do a separate one, reviewer's choice?

Closes https://github.com/bystrogenomics/bystro/issues/152


python/python/bystro/ancestry/tests/test_train.py
```diff
@@ -1,10 +1,65 @@
...
+from pandas.testing import assert_frame_equal
 
-from bystro.ancestry import train
+from bystro.ancestry.train import VARIANT_REGEX, _parse_vcf_from_file_stream
 
-## this is not a very useful test, really just testing that train
-## loaded successfully, didn't fail on import
 
+def test_VARIANT_REGEX():
+    assert VARIANT_REGEX.match("chr1:123456:A:A")
```
cristinaetrv, line 9: Can we include match tests where ref and alt are different instead of the same? We probably want to throw an error when ref=alt in the variant ID maybe as an assert in the train code for the processed vcf if there isn't one yet.

python/python/bystro/ancestry/train.py
```diff
@@ -454,27 +662,28 @@ def _get_mi_df(train_X: pd.DataFrame, train_y_pop: pd.Series) -> pd.DataFrame:
...
+    skops_dump(rfc, rfc_fpath)
 
 
 def main() -> None:
     """Train global ancestry model."""
     err_msg = "This module isn't production-ready yet!"
     raise NotImplementedError(err_msg)
     genotypes, labels = _load_dataset()
-    train_X, test_X, train_y, test_y = _make_train_test_split(
+    train_X, test_X, train_y, test_y = make_train_test_split(
```
cristinaetrv, line 682: Do you need to repeat making the test/train split here since there's a version on line 465? Or is it set up differently because it's a different model?

python/python/bystro/ancestry/train.py
```diff
@@ -376,39 +501,125 @@ def sample(cls) -> "RFCParamChoices":
...
+
+    return AccuracyReport(
+        accuracy_score(train_y.population, train_yhat_pops),
+        accuracy_score(test_y.population, test_yhat_pops),
+        accuracy_score(train_y.superpop, train_yhat_superpops),
+        accuracy_score(test_y.superpop, test_yhat_superpops),
+    )
+
+
+def make_rfc(
```
cristinaetrv, line 537: Will the final version of train include multiple models (the RFC and the latest model)?

python/python/bystro/ancestry/train.py
```diff
@@ -280,8 +406,8 @@ def _filter_variants_for_fst(genotypes: pd.DataFrame, labels: pd.DataFrame) -> p
 def _load_dataset() -> tuple[pd.DataFrame, pd.DataFrame]:
     """Return the final dataset consisting of genotype matrix, labels."""
     genotypes = _load_genotypes()
-    labels = _load_label_data(genotypes.index)
-    genotypes, labels = _filter_samples_for_relatedness(genotypes, labels)
+    labels = load_label_data(genotypes.index)
+    genotypes, labels = filter_samples_for_relatedness(genotypes, labels)
```
cristinaetrv, line 410: Are we filtering samples for relatedness with the idea that a future reference dataset will include related individuals? If not, do we still need this step?

python/python/bystro/ancestry/train.py
```diff
@@ -376,39 +501,125 @@ def sample(cls) -> "RFCParamChoices":
...
+class AccuracyReport:
+    """Represent model accuracy scores at population and superpopulation levels."""
+
+    train_pop_accuracy: float
+    test_pop_accuracy: float
+    train_superpop_accuracy: float
+    test_superpop_accuracy: float
+
+
+def _compute_accuracy_report(
```
cristinaetrv, line 514: Annotation here and for the following function would be helpful

python/python/bystro/ancestry/train.py
```diff
@@ -280,8 +406,8 @@ def _filter_variants_for_fst(genotypes: pd.DataFrame, labels: pd.DataFrame) -> p
 def _load_dataset() -> tuple[pd.DataFrame, pd.DataFrame]:
     """Return the final dataset consisting of genotype matrix, labels."""
     genotypes = _load_genotypes()
-    labels = _load_label_data(genotypes.index)
-    genotypes, labels = _filter_samples_for_relatedness(genotypes, labels)
+    labels = load_label_data(genotypes.index)
+    genotypes, labels = filter_samples_for_relatedness(genotypes, labels)
     genotypes = _filter_variants_for_maf(genotypes)
```
cristinaetrv, line 411: MAF filtering is already a filter in the preprocess step - is there a reason we want to repeat that preprocess step here?

cristinaetrv (approved):
Overall lgtm, have some non-blocking clarification questions that might make more sense to discuss on a higher bandwidth medium.

python/python/bystro/ancestry/tests/test_train.py
poneill: good catch!

python/python/bystro/ancestry/train.py
poneill: yes, train.py is a bit of a junk drawer at the moment-- a lot of it was experimental code that I amnestied just to get it on github.  After deploying the chip model, glad to come back and clean up what we no longer need.

NB: we're publicizing this method because we'll need to import it into a separate script for training the chip model (to come shortly)

python/python/bystro/ancestry/train.py
poneill: we'll probably never want to include related individuals in an ancestry model, so maybe we should canonicalize the sample set and push this step further upstream.  Added TODO to this effect.

python/python/bystro/ancestry/train.py
poneill: We're going to aim for one model for now and see if a use case for a second emerges.  for now we're just publicizing the method so we can import it into a separate chip model training script later.

python/python/bystro/ancestry/train.py
poneill: Do you mean a docstring?  Added, but which other function?  Looks like make_rfc's already covered

python/python/bystro/ancestry/train.py
poneill: I think it may be easier to discuss offline?

poneill:
> Overall lgtm, have some non-blocking clarification questions that might make more sense to discuss on a higher bandwidth medium.

Thanks, addressed what I could, happy to discuss the rest offline but will merge for now

PR #154, opened 2023-07-17
Add code to calculate shared illumina / affymetrix callset

poneill:
Add code to:

- load illumina, affymetrix callsets
- liftover variants from 37 to 38
- calculate intersection

### DIFF
--- a/python/python/bystro/ancestry/define_callset.py
+++ b/python/python/bystro/ancestry/define_callset.py
@@ -26,25 +26,43 @@ def get_watson_crick_complement(base: str) -> str:
 def liftover_38_from_37(variant: str) -> str | None:
     """Liftover a variant to genome build 38 from 37."""
     chrom, pos, ref, alt = variant.split(":")
-    locations = converter[int(chrom.lstrip("chr"))][int(pos)]
+    # liftover doesn't deal gracefully with MT variants, but we don't need them anyway
+    if "MT" in chrom:
+        return None
+    locations = converter[chrom][int(pos)]
     if locations is None or len(locations) != 1:
         logger.info("Variant %s had a non-unique location, couldn't lift over", variant)
         return None
     chrom38, pos38, _strand38 = locations[0]
-    variant38 = ":".join([chrom38, str(pos), ref, alt])
+    variant38 = ":".join([chrom38, str(pos38), ref, alt])
     assert_true("lifted-over variant starts with 'chr'", variant38.startswith("chr"))
     return variant38
 
 
-def load_illumina_callset() -> pd.Series:
-    """Load list of variants for illumina chip."""
-    illumina_df = pd.read_csv(ILLUMINA_FILEPATH, skiprows=7)
+def _load_illumina_df() -> pd.DataFrame:
+    comment_rows = 7
+    columns_to_keep = ["Chr", "MapInfo", "SNP", "RefStrand"]
+    illumina_df = pd.read_csv(ILLUMINA_FILEPATH, skiprows=comment_rows)
     assert_equals(
         "Set of genome builds",
         {37.1},
         "actual set of genome builds",
         set(illumina_df.GenomeBuild.dropna()),
     )
+    return illumina_df[columns_to_keep]
+
+
+def load_illumina_callset() -> pd.Series:
+    """Load list of variants for illumina chip."""
+    illumina_df = _load_illumina_df()
+    illumina_variants = _process_illumina_df(illumina_df)
+    assert_equals(
+        "number of illumina variants", 578822, "recovered number of variants", len(illumina_variants)
+    )
+    return illumina_variants
+
+
+def _process_illumina_df(illumina_df: pd.DataFrame) -> pd.Series:
     variants = []
     liftover_attempts = 0
     liftover_failures = 0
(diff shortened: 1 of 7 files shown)


python/python/bystro/ancestry/define_callset.py
```diff
@@ -0,0 +1,141 @@
...
+        "Set of genome builds",
+        {37.1},
+        "actual set of genome builds",
+        set(illumina_df.GenomeBuild.dropna()),
+    )
+    return illumina_df[columns_to_keep].dropna()
+
+
+def load_illumina_variants() -> pd.Series:
+    """Load list of variants for illumina chip."""
```
cristinaetrv, line 57: I know the chip name is in the file path but I think it might be helpful to put the names of the chips in the doc strings so it gets captured for future reference and in autodocs.

python/python/bystro/ancestry/define_callset.py
```diff
@@ -0,0 +1,141 @@
...
+    illumina_df = _load_illumina_df()
+    illumina_variants = _get_variants_from_illumina_df(illumina_df)
+    assert_equals(
+        "number of illumina variants", 578822, "recovered number of variants", len(illumina_variants)
```
cristinaetrv, line 61: Just to check I'm following this correctly, 578822 is the number of total autosomal SNVs after liftover?

python/python/bystro/ancestry/train_utils.py
```diff
@@ -30,3 +30,30 @@ def get_variant_ids_from_callset(callset: dict[str, Any]) -> np.ndarray:
...
+    [ACGT]  # ref allele
+    :
+    [ACGT]  # alt allele
+    $
+    """,
+    re.VERBOSE,
+)
+
+
+def is_autosomal_variant(potential_variant: str) -> bool:
```
cristinaetrv, line 51: Do you want to add a more explicit check whether the chr isn't X,Y,MT here and then you could add this into the liftover step instead of having that more spread out across the loading snp chip and liftover steps? Will mark in define_callset

python/python/bystro/ancestry/define_callset.py
```diff
@@ -0,0 +1,141 @@
...
+def get_watson_crick_complement(base: str) -> str:
+    """Calculate Watson-Crick Complement."""
+    wc_dict = {"A": "T", "T": "A", "G": "C", "C": "G"}
+    return wc_dict[base]
+
+
+def liftover_38_from_37(variant: str) -> str | None:
+    """Liftover a variant to genome build 38 from 37."""
+    chrom, pos, ref, alt = variant.split(":")
+    if "MT" in chrom:
```
cristinaetrv, line 30: It looks like there already shouldn't be MT in either illumina.variants or affymetrix.variants workflow when it gets to this part so I think this step might be redundant? Or it might make sense to include the is_autosomal_variant step here.

python/python/bystro/ancestry/train_utils.py
poneill: yeah you've def picked up on one of the fault lines in this design.  Conceptually, I think we actually want to make two distinct checks: 

1.  Is it a variant we want to keep?  (i.e. an autosomal variant)
2. Is it a variant that would break liftover?  (an MT variant)

The reason we want to keep these checks distinct is that liftover will fail for a small fraction (perhaps 0.05%) of variants just because they can't be mapped to a unique location in the 38 build-- we monitor the proportion of variants that fail liftover and report it to the user as a sanity check.  Liftover can handle X and Y variants just fine, we simply don't want them in our model.  So if we report X, Y variants as failing in liftover, we'll artificially inflate the liftover failure rate to something that looks alarmingly high, when it doesn't actually signal a problem with liftover.  

But as it is, is_autosomal_variant no longer needs to handle MT variants, so we can take the check out in the interest of simplicity.

python/python/bystro/ancestry/define_callset.py
poneill: yes, took the redundant check out as per https://github.com/bystrogenomics/bystro/pull/154#discussion_r1265805259, thanks!

python/python/bystro/ancestry/define_callset.py
poneill: that's correct, is there a way of writing this assertion that would make it easier to read at a glance?

python/python/bystro/ancestry/define_callset.py
```diff
@@ -0,0 +1,141 @@
...
+        if not is_autosomal_variant(variant37):
+            continue
+        variants38.append(liftover_38_from_37(variant37))
+    illumina_variants = pd.Series(variants38)
+    liftover_failure_rate = illumina_variants.isna().mean()
+    logger.info("liftover failure rate: %1.2f%%", liftover_failure_rate * 100)
+    return illumina_variants.dropna()
+
+
+def _load_affymetrix_df() -> pd.DataFrame:
```
cristinaetrv, line 89: This may be something for QC we develop later that integrates the annotator, but did we want to include something for the affymetrix manifest that includes checking whether or not the rsID and position matches since we identified incorrect positions for some of them?

python/python/bystro/ancestry/define_callset.py
cristinaetrv: maybe just something a little more specific like 'autosomal SNVs after liftover' vs 'recovered number of variants' for quick clarity on what the final number is representing

python/python/bystro/ancestry/define_callset.py
```diff
@@ -0,0 +1,141 @@
...
+    illumina_df = _load_illumina_df()
+    illumina_variants = _get_variants_from_illumina_df(illumina_df)
+    assert_equals(
+        "number of illumina variants", 578822, "recovered number of variants", len(illumina_variants)
+    )
+    return illumina_variants
+
+
+def _get_variants_from_illumina_df(illumina_df: pd.DataFrame) -> pd.Series:
+    """Extract affymetrix variants and lift over."""
```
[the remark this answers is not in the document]
cristinaetrv:
```suggestion
    """Extract illumina variants and lift over."""
```

cristinaetrv (approved):
Looks good to me - I suggested a change on a minor typo but otherwise overall looks to be in good shape!

python/python/bystro/ancestry/define_callset.py
poneill: Agreed in general, think it's out of scope for this PR (and we have accuracy checks on the model in place so we know that the discordance between rsids and coordinates can't be *that* bad).  Added a todo.

python/python/bystro/ancestry/define_callset.py
poneill: good point!

PR #155, opened 2023-07-17
add train_chip_model.py

poneill:
commit training code for PCA + RF model built off of shared illumina / affy variants.

### DIFF
--- a/python/python/bystro/ancestry/train_chip_model.py
+++ b/python/python/bystro/ancestry/train_chip_model.py
@@ -1,0 +1,52 @@
+"""Implement PCA + RF for shared Illumina + Affymetrix variants."""
+
+import pandas as pd
+from sklearn.decomposition import PCA
+
+from bystro.ancestry.train import (
+    INTERMEDIATE_DATA_DIR,
+    filter_samples_for_relatedness,
+    load_callset_for_variants,
+    load_label_data,
+    make_rfc,
+    make_train_test_split,
+    serialize_model_products,
+)
+
+
+def load_illumina_affy_variants() -> set[str]:
+    """Get previously computed intersection of illumina and affymetrix variants."""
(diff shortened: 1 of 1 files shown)


python/python/bystro/ancestry/train_chip_model.py
```diff
@@ -0,0 +1,52 @@
...
+    """PCA transform dataframe, retaining index and labeling columns appropriately."""
+    pc_columns = ["pc" + str(i) for i in range(1, pca.num_components_ + 1)]
+    return pd.DataFrame(pca.transform(X), index=X.index, columns=pc_columns)
+
+
+def main() -> None:
+    """Train PCA, RF for Illumina and Affymetrix variants, save model products to disk."""
+    kgp_genotypes = load_kgp_genotypes_for_shared_variants()
+    labels = load_label_data(kgp_genotypes.index)
+    kgp_genotypes, labels = filter_samples_for_relatedness(kgp_genotypes, labels)
```
cristinaetrv, line 38: I think we already agreed to discuss this further offline, but just want to mention here again for the record filtering on relatedness might make more sense as a preprocess step for any reference data further upstream - by this step I think all the processing of these files should be completed and at the moment we're not expecting reference data that includes related individuals.

cristinaetrv (approved):
Lgtm!

python/python/bystro/ancestry/train_chip_model.py
poneill: Agreed, we have a todo for this in the code-- there are a few options about where to put this, depending on how strongly we want to bake this assumption into the code base.  If we're very comfortable with a canonical sample list we could push it as far upstream as parsing the vcf file itself.  Going to leave this as a todo for now because it's not bottlenecking the deployment but happy to revisit it once the dust has settled.

PR #158, opened 2023-07-20
relax ruff to allow named expressions before returns

### LINKED ISSUE #157: Stop blocking PRs based on linter
We use ruff for linting, which does catch genuine errors but some aspects of ruff are unnecessarily opinionated and block legitimate coding choices.

poneill:
This rule interferes with readability, so let's get rid of it.

closes #157


cristinaetrv (approved):
Lgtm!

PR #159, opened 2023-07-20
Add CONTRIBUTING.md and docs/style_guide.md

poneill:
Add rough cuts of both of these docs just to get the ball rolling.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,15 @@
+# Contributing to Bystro
+
+[WIP]
+
+Welcome to Bystro! We're glad you're interested in contributing to our project. Before you start,
+please take a moment to familiarize yourself with our approach and guidelines.
+
+## Installing Bystro
+[TK]
+
+## Contributing a PR
+We welcome 3rd-party contributions.  Out of respect for your time,
+please open an issue and confirm the basic approach / design with the
+core team before submitting a PR.  Please also see our code [style
+guide](docs/style_guide.md).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,6 @@ please take a moment to familiarize yourself with our approach and guidelines.
 [TK]
 
 ## Contributing a PR
-We welcome 3rd-party contributions.  Out of respect for your time,
-please open an issue and confirm the basic approach / design with the
-core team before submitting a PR.  Please also see our code [style
-guide](docs/style_guide.md).
+We welcome 3rd-party contributions.  Out of respect for your time, please open an issue and confirm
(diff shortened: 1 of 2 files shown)


poneill:
Please suggest edits liberally.

docs/style_guide.md
```diff
@@ -0,0 +1,60 @@
+# Contributing to Bystro
+
+[WIP]
+
+Welcome to Bystro! We're glad you're interested in contributing to our project. Before you start,
+please take a moment to familiarize yourself with our approach and guidelines.
+
+## Installing Bystro
+[TK]
```
cristinaetrv, line 9: I think the installation guide should be its own doc - maybe just put a link here? Same in contributing doc.

poneill: added link to pre-existing installation guide in CONTRIBUTING.md

CONTRIBUTING.md
```diff
@@ -0,0 +1,14 @@
+# Contributing to Bystro
+
+[WIP]
```
cristinaetrv, line 3: I think we can remove this WIP marker for now, and just add changes as needed in future

CONTRIBUTING.md
```diff
@@ -0,0 +1,14 @@
...
+
+Welcome to Bystro! We're glad you're interested in contributing to our project. Before you start,
+please take a moment to familiarize yourself with our approach and guidelines.
+
+## Installing Bystro
+Please see our [installation guide](INSTALL.md).
+
+## Contributing a PR
```
cristinaetrv, line 11: Do we want to describe the forking model for all contributions since we decided on that as a generalized approach? Or add something like that to the style guide contribution portion?

poneill: The forking model can be taken for granted for 3rd party contributors.  For 1st party contributions I would prefer to wait for the dust to settle.  We (or I, am least) don't think I've yet found all the snags that can occur when one forks a repo one has commit rights to.

docs/style_guide.md
```diff
@@ -0,0 +1,48 @@
+# Style Guide
+
+## Notes on Tooling
```
cristinaetrv, line 3: Should we mention the python version we're using somewhere in here?

cristinaetrv (approved):
No blocking comments on my end - I'll leave it unmerged for now in case anyone else wants to make any other suggestions/adjustments before it goes through. Otherwise, I think it's a good start on the style guide! If we want to include things like very specific conventions, testing guidance, etc we agree on in future, we can just add it to this doc.

akotlar (approved):

akotlar:
LGTM

PR #162, opened 2023-07-21
add s3 upload script for ancestry model products

### LINKED ISSUE #161: Upload model products to s3

poneill:
upload ancestry model products (to wit: PCA, RFC objects) to s3://bystro-ancestry.

Closes #161

### DIFF
--- a/python/python/bystro/ancestry/upload_model.py
+++ b/python/python/bystro/ancestry/upload_model.py
@@ -1,0 +1,49 @@
+"""Upload PCA and RFC models to ancestry s3 bucket."""
+
+import logging
+import os
+import sys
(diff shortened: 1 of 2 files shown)


python/python/bystro/ancestry/upload_model.py
```diff
@@ -0,0 +1,49 @@
+"""Upload PCA and RFC models to ancestry s3 bucket."""
```
poneill, line 1: This script is basically pure IO so it's not a great candidate for unit testing (but change my view?).  The script works and the model products are already uploaded-- I'm just checking it in for posterity / automating future deployments.

akotlar (approved):

PR #172, opened 2023-07-23
Clean up bystro-utils

akotlar:
* Update help message
* Error early if the config file is defines a utils section with a member that doesn't have a `name` property
* Use all CPU threads by default
* In FilterCadd compress by default, to match SortCadd behavior


poneill (changes requested):
I'm confused about what's happening with CI here: we're not running any checks.  This, in turn, seems like two distinct issues:

- we're not running perl tests
- we're not running python tests.  Even though the python tests aren't relevant to this PR, I don't think the github actions runner knows that yet?  So I think we should iron out the CI issues here first.

akotlar:
@poneill https://github.com/bystrogenomics/bystro/pull/177 will fix the GitHub actions issue. Nice catch

PR #173, opened 2023-07-24
remove nonce file for testing branch protection rules

poneill:
Previously we had branch protection rules in place for master, but apparently they didn't apply by default to administrators (which is most of us).  I've gone ahead and applied the rule to administrators too.  So now the nonce file used to test the branch protection rules must be removed via PR.


cristinaetrv (approved):
Looks like this concludes this test.

PR #176, opened 2023-07-24
Add s3 handling to pull down ancestry model

poneill:
- pull down models from s3
- pick up vcf files from EFS
- minor refactoring of inference code
- more tests for listener.py
- drop MAG population from ancestry_types.py


python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -122,7 +122,6 @@ class PopulationVector:
     JPT: ProbabilityInterval = field(validator=ProbIntValidator)
     KHV: ProbabilityInterval = field(validator=ProbIntValidator)
     LWK: ProbabilityInterval = field(validator=ProbIntValidator)
-    MAG: ProbabilityInterval = field(validator=ProbIntValidator)
```
cristinaetrv, line 125: Is MAG being removed because we're no longer expecting it in any datasets or is it being removed here for a different reason?

poneill: no longer expecting it in any datasets

python/python/bystro/ancestry/listener.py
```diff
@@ -38,30 +106,99 @@ def _load_queue_conf(queue_conf_path: str) -> QueueConf:
     return QueueConf(addresses=beanstalk_conf["addresses"], tubes=beanstalk_conf["tubes"])
 
 
+def _fill_missing_data(genotypes: pd.DataFrame) -> tuple[pd.DataFrame, pd.Series]:
```
cristinaetrv, line 109: I'm a little confused why this imputation step is happening in the listener script - I think it should be part of preprocess or train?

python/python/bystro/ancestry/listener.py
```diff
@@ -38,30 +106,99 @@ def _load_queue_conf(queue_conf_path: str) -> QueueConf:
...
+    sample_missingnesses = genotypes.isna().mean(axis="columns")  # average over columns, leaving indices
+    # todo: much better imputation strategy to come, but we're stubbing out for now.
+    # if col completely missing in all samples, just fill as heterozygote for now
```
cristinaetrv, line 112: Is the result exactly the same if all samples are all homozygous ref or all hets?

poneill: If there's at least one sample for which genotyping data for a given variant is present, we'll take the mean dosage over all samples where we have data.  If it's completely absent from the dataset, however, we're arbitrarily filling it in as heterozygote just to get a complete input matrix.

This isn't a great imputation strategy-- in fact, it's about the worst one.  I'm preparing another PR to do imputation in a more sophisticated way-- this is just a placeholder to get the model running end to end and will be replaced as soon as we can demonstrate that capability.

Meanwhile lmk if there's any way I can make that clearer in the code?

python/python/bystro/ancestry/listener.py
poneill: This is for the inference step, so the genotype matrix we're filling out here won't be available at train time.  Certainly open to suggestions about how to make that distinction clearer in the code.

cristinaetrv (approved):
Overall lgtm, especially after separating out inference steps to their own model as we discussed offline.

PR #177, opened 2023-07-25
Update lint.yml

akotlar:
* Updates GitHub actions to run on pull_request in addition to push.

We may want to limit to only running tests on pull request, because GitHub actions cost us money and will have a lot of noise (every commit push to the repo), but leaving for now until we know that we don't need them.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,6 @@
 name: lint
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:


poneill (approved):
It seems there's a question of whether we want `pull_request` or `pull_request_target` here, the issue being that allowing 3rd party contributors to submit PRs with CI checks that run according to infra code in their own branch, rather than the infra code in their target branch, is something of a security risk.  (see: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/). 

But this is an improvement over the status quo, so approving for now.

If we are concerned about costs I'd rather configure CI so that it doesn't run on every push, but still keep the full pipeline in place and require passing checks on the last commit before merging.  Ruff and mypy will still catch things that a less-than-perfect test suite won't.

PR #184, opened 2023-07-27
Add preprocess 1KGP using gnomad loadings for Ancestry

cristinaetrv:


python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
```diff
@@ -0,0 +1,37 @@
...
+
+#TODO convert this process to one using our own vcf parser
+
+#Download plink2 - can be downloaded with example below or at https://www.cog-genomics.org/plink/2.0/
+#wget https://s3.amazonaws.com/plink2-assets/plink2_linux_avx2_20230707.zip
+#unzip plink2_linux_avx2_20230707.zip 
+
+#Download 1kgp genomes - make sure this is the most recent version
+wget 'ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/data_collections/1000G_2504_high_coverage/working/20220422_3202_phased_SNV_INDEL_SV/1kGP_high_coverage_Illumina.chr'*
+
```
poneill, line 11: let's add a checksum to ensure that we've got the data we think we do

python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
```diff
@@ -0,0 +1,37 @@
+#Extract gnomad loadings variant list from 1kgp genomes using plink
+
+#TODO convert this process to one using our own vcf parser
+
+#Download plink2 - can be downloaded with example below or at https://www.cog-genomics.org/plink/2.0/
```
poneill, line 5: The user won't necessarily see this comment if they try to run the script without plink2 installed-- we could write something like:

```bash
# Define the program name
program_name="plink2"

# Check if the program is installed and present on $PATH
if ! command -v "$program_name" &> /dev/null; then
    echo "Error: '$program_name' is not installed or not present on the system's PATH.
    echo "Please install '$program_name' or make sure it's added to the PATH.  You can do this by blah blah blah"
    exit 1
fi

echo "'$program_name' is installed and present on the system's PATH."
```

python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
```diff
@@ -0,0 +1,37 @@
...
+    #Input and output paths for each chromosome
+    input_vcf="1kGP_high_coverage_Illumina.chr${chr}.filtered.SNV_INDEL_SV_phased_panel.vcf.gz"
+    output_base="chr${chr}"
+
+    #Run plink2 with current chromosome
+    ./plink2 --extract gnomadvariantlist.txt \
+             --make-pgen \
+             --out "$output_base" \
+             --vcf "$input_vcf"
+     # Append output file name to the merge list, excluding 'chr1'
```
poneill, line 26: this seems a little more error-prone than writing each file to a tmp dir and concatenating them all at the end with `vcfcat`-- what happens to the header lines for chrs2-22?

poneill (changes requested):

python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
cristinaetrv: Should I add one for each of the files? There's I think 46 that come through at this point. Also to that end, should I add a step to delete them all at the end?

python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
cristinaetrv: The vcf commented headers aren't all kept through the plink processing as it's converted to plink format to do so, but I'm not sure exactly what you mean here. Are you suggesting not using the plink process at all here? I just wanted to document this method for this preprocess step that is used for this version of 1kgp until this step is converted to our own vcf parser module.

python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
cristinaetrv: Ok I found the checksums in the manifest so was able to easily include from there

python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
poneill: ok, if there are no header lines I withdraw my comment :)

poneill (approved):
lgtm!

PR #186, opened 2023-07-28
enable automatic discovery of tests by pytest

### LINKED ISSUE #185: Enable automatic discovery of tests by pytest
Currently some test files are prefixed by `t_` instead of `test_`, meaning that these tests are skipped during CI.  Rename them so that pytest can discover them automatically.

poneill:
Rename some test dirs and files so that they get picked up by pytest during CI.

Closes #185


akotlar (approved):
Ok to merge, pytest uses test or tests as the default directory it searches for, recursively.

PR #187, opened 2023-07-28
Add vcf simulator for testing analysis modules

cristinaetrv:


python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
poneill, line 1: looks like this was dragged in from #184 ?

python/python/bystro/vcf_utils/sim_data_generator.py
```diff
@@ -0,0 +1,61 @@
+"""VCF index and data simulator for testing analysis modules."""
+
+import random
+from io import StringIO
+
+import pandas as pd
+
+HEADER_COLS = ["#CHROM", "POS", "ID", "REF", "ALT", "QUAL", "FILTER", "INFO", "FORMAT"]
```
poneill, line 8: nit: probably should be `"CHROM` and then the comment syntax goes in separately

python/python/bystro/vcf_utils/sim_data_generator.py
```diff
@@ -0,0 +1,61 @@
...
+    header = HEADER_COLS + sample_ids
+    vcf_data = []
+    simulated_indices = []
+    vcf_data.append("\t".join(header))
+    for position in range(num_vars):
+        chrom = "chr1"
+        pos = position
+        variant_id = generate_random_vcf_index()  # Using simulated index as variant ID
+        simulated_indices.append(variant_id)  # Store the simulated index
+        ref = random.choice(["A", "T", "C", "G"])
```
poneill, line 35: I think I'm lost here-- why are we regenerating info that we've already created with `generated_random_vcf_index`?

python/python/bystro/vcf_utils/sim_data_generator.py
```diff
@@ -0,0 +1,61 @@
...
+        alt = random.choice([letter for letter in ["A", "T", "C", "G"] if letter != ref])
+        qual = random.randint(0, 100)
+        filter_ = "PASS"
+        info = "."
+        format_ = "GT"
+        samples = []
+        for _ in range(num_samples):
+            genotype = random.choice(["0|0", "0|1", "1|0", "1|1"])
+            samples.append(genotype)
+        record = [chrom, str(pos), variant_id, ref, alt, str(qual), filter_, info, format_]
```
poneill, line 45: nit: slightly easier to follow if this line comes before L42-44, as it is we're generating the line from right to left (first the body and then the header)

python/python/bystro/vcf_utils/sim_data_generator.py
```diff
@@ -0,0 +1,61 @@
...
+        record.extend(samples)
+        vcf_data.append("\t".join(record))
+    # Convert to pandas DataFrame for further processing
+    vcf_str = "\n".join(vcf_data)
+    vcf_df = pd.read_csv(StringIO(vcf_str), delimiter="\t")
+    return vcf_df, simulated_indices
```
poneill, line 51: we're duplicating info here, because vcf_df already contains a column for variant_ids, which is what simulated_indices contains

python/python/bystro/vcf_utils/sim_data_generator.py
```diff
@@ -0,0 +1,61 @@
...
+
+import random
+from io import StringIO
+
+import pandas as pd
+
+HEADER_COLS = ["#CHROM", "POS", "ID", "REF", "ALT", "QUAL", "FILTER", "INFO", "FORMAT"]
+
+
+def generate_random_vcf_index():
```
poneill, line 11: here and throughout, type annotations pls :)

python/python/bystro/vcf_utils/sim_data_generator.py
```diff
@@ -0,0 +1,61 @@
...
+    random_chr = random.randint(1, 22 + 1)
+    random_pos = random.randint(1, 1000000)
+    random_ref = random.choice(["A", "T", "C", "G"])
+    random_alt = random.choice([letter for letter in ["A", "T", "C", "G"] if letter != random_ref])
+    return f"{random_chr}:{random_pos}:{random_ref}:{random_alt}"
+
+
+def generate_simulated_vcf(
+    num_samples:int, num_vars:int
+) -> tuple[pd.DataFrame, str]:
```
poneill, line 22: I would push back against this return type, actually.  If we're simulating VCFs, seems like we'd get more mileage out of a textual type (probably just `str`) rather than a df so we can test df parsing.

python/python/bystro/vcf_utils/sim_data_generator.py
```diff
@@ -0,0 +1,61 @@
...
+        format_ = "GT"
+        samples = []
+        for _ in range(num_samples):
+            genotype = random.choice(["0|0", "0|1", "1|0", "1|1"])
+            samples.append(genotype)
+        record = [chrom, str(pos), variant_id, ref, alt, str(qual), filter_, info, format_]
+        record.extend(samples)
+        vcf_data.append("\t".join(record))
+    # Convert to pandas DataFrame for further processing
+    vcf_str = "\n".join(vcf_data)
```
poneill, line 49: we should probably just return `vcf_str` here

python/python/bystro/vcf_utils/sim_data_generator.py
```diff
@@ -0,0 +1,61 @@
...
+    return vcf_df, simulated_indices
+
+
+def write_sim_vcf_with_comments(vcf_data: pd.DataFrame):
+    """Add a specific number of comment lines to test header is removed properly."""
+    num_comment_lines = 107
+    comment_lines = [f"# Comment line {i}" for i in range(1, num_comment_lines + 1)]
+    sim_vcf_with_comments = "\n".join(comment_lines) + "\n" + "\n".join(vcf_data)
+    # Save the simulated VCF as a TSV file
+    with open("simulated.vcf", "w") as file:
```
poneill, line 60: this function is two different ideas: adding comment lines, and doing disk IO.  We want to push disk IO out to the edges of the API as much as possible so that the core remains easy to test.

We also seem to be passing the data from str to df back to str, whereas if we made this part of the vcf generation code we could do it in one pass.

python/python/bystro/vcf_utils/tests/test_sim_data_generator.py
```diff
@@ -0,0 +1,35 @@
+"""Test simulated data."""
+
+import pandas as pd
+from bystro.vcf_utils.sim_data_generator import (
+    generate_random_vcf_index, 
+    generate_simulated_vcf
+)
+
+HEADER_COLS = ["#CHROM", "POS", "ID", "REF", "ALT", "QUAL", "FILTER", "INFO", "FORMAT"]
```
poneill, line 9: reduplicated from sim_data_generator.py?

python/python/bystro/vcf_utils/tests/test_sim_data_generator.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_generate_random_vcf_index():
+    """Test expected index components have correct parameters."""
+    random_vcf_index = generate_random_vcf_index()
+    assert isinstance(random_vcf_index, str)
+    parts = random_vcf_index.split(":")
```
poneill, line 16: could write this as something like `chrom, pos, ref, alt = random_vcf_index.split(":")` to make assertions below more readable

python/python/bystro/vcf_utils/tests/test_sim_data_generator.py
```diff
@@ -0,0 +1,35 @@
...
+    assert len(parts) == 4
+    assert parts[0].isdigit()
+    assert 1 <= int(parts[0]) <= 22
+    assert parts[1].isdigit()
+    assert 1 <= int(parts[1]) <= 1000000
```
poneill, line 21: nit: probably what we really care about is the fact that it's >= 1, the 1_000_000 is an implementation detail

python/python/bystro/vcf_utils/tests/test_sim_data_generator.py
```diff
@@ -0,0 +1,35 @@
...
+    assert parts[3] in ["A", "T", "C", "G"]
+    assert parts[2] != parts[3]
+
+
+def test_generate_simulated_vcf():
+    """Test generate_simulated_vcf."""
+    num_samples = 10
+    num_vars = 10
+    simulated_vcf, simulated_indices = generate_simulated_vcf(num_samples, num_vars)
+    assert isinstance(simulated_vcf, pd.DataFrame)
```
poneill, line 32: mypy will catch these for you, but it won't (at least not without a lot of ceremony) check whether the genotypes are valid, whether the variants and samples are unique &c. (in fact technically speaking the variants don't have to be unique in this implementation).

Generally, I think we want to let mypy handle `isinstance`-type checks and let the unit tests focus on higher-level properties that we can't verify statically.

python/python/bystro/vcf_utils/tests/test_sim_data_generator.py
```diff
@@ -0,0 +1,35 @@
...
+    assert isinstance(simulated_indices, list)
+    assert len(simulated_indices) == num_vars
```
poneill, line 34: See above comment: because simulated_indices is reduplicating info from the df, we have to write this check-- if we just return the df then there's no way for this to go wrong because it's impossible to construct a df with a differing number of rows per column.  

This is a particular case of the general principle that whenever we have two lists (or list-like collections) of corresponding data, it's better to express that as a list of tuples than two separate lists, because in the latter case we can't even write the error of the lists not being of the same length.  There are certain exceptions where the contrary convention has the weight of tradition behind it (like writing `X, y` for the features and labels of a dataset in ML) but it's a useful habit to keep in mind.

python/python/bystro/vcf_utils/sim_data_generator.py
poneill, line 1: consider renaming as something more specific, e.g. `simulate_random_vcf.py` or some such.

poneill (changes requested):

python/python/bystro/ancestry/preprocess_1kgp_using_gnomad_loadings.sh
cristinaetrv: Yep, this sh script shouldn't be part of this PR (all follow-up edits to this file are in 184)

python/python/bystro/vcf_utils/sim_data_generator.py
cristinaetrv: I initially included this part to test loading the exact 1kgp code, but I'm not sure I need this function at all anymore after separating that function out (in upcoming PR) unless I want to replicate a certain vcf and save it. Thoughts on removing this function or making it a private function? It's not part of the unit tests anymore.

python/python/bystro/vcf_utils/simulate_random_vcf.py
```diff
@@ -0,0 +1,65 @@
...
+    vcf_str = "\n".join(vcf_data)
+    vcf_df = pd.read_csv(StringIO(vcf_str), delimiter="\t")
+    return vcf_df
+
+
+def write_sim_vcf_with_comments(vcf_data: pd.DataFrame):
+    """Add a specific number of comment lines to test header is removed properly."""
+    num_comment_lines = 107
+    comment_lines = [f"# Comment line {i}" for i in range(1, num_comment_lines + 1)]
+    sim_vcf_with_comments = "\n".join(comment_lines) + "\n" + "\n".join(vcf_data)
```
poneill, line 62: is this line under test?

if we call `"\n".join(df)` on a dataframe, it won't crash but it will silently concatenate the columns, i.e.:

```python
df = pd.DataFrame({"foo":[1], "bar":[2]})
"\n".join(df)
```

gives you: `'foo\nbar'`

python/python/bystro/vcf_utils/simulate_random_vcf.py
```diff
@@ -0,0 +1,65 @@
...
+        info = "."
+        format_ = "GT"
+        samples = []
+        record = [str(chrom), str(pos), variant_id, ref, alt, str(qual), filter_, info, format_]
+        for _ in range(num_samples):
+            genotype = random.choice(["0|0", "0|1", "1|0", "1|1"])
+            samples.append(genotype)
+        record.extend(samples)
+        vcf_data.append("\t".join(record))
+    return vcf_data, simulated_indices
```
poneill, line 48: same question as before: is there a reason we need to return `simulated_indices` if it's reduplicating data contained in vcf_data?

python/python/bystro/vcf_utils/simulate_random_vcf.py
```diff
@@ -0,0 +1,65 @@
+"""VCF index and data simulator for testing analysis modules."""
+
+import random
+from io import StringIO
+
+import pandas as pd
+
+HEADER_COLS = ["#CHROM", "POS", "ID", "REF", "ALT", "QUAL", "FILTER", "INFO", "FORMAT"]
```
poneill, line 8: could we move the comment syntax out of HEADER_COLS?

python/python/bystro/vcf_utils/simulate_random_vcf.py
```diff
@@ -0,0 +1,65 @@
...
+        record = [str(chrom), str(pos), variant_id, ref, alt, str(qual), filter_, info, format_]
+        for _ in range(num_samples):
+            genotype = random.choice(["0|0", "0|1", "1|0", "1|1"])
+            samples.append(genotype)
+        record.extend(samples)
+        vcf_data.append("\t".join(record))
+    return vcf_data, simulated_indices
+
+
+def convert_sim_vcf_to_pd(vcf_data: str) -> pd.DataFrame:
```
poneill, line 51: we want to test round-tripping the vcf to df conversion to make sure the composition of vcf -> df  and df -> vcf is the identity.

python/python/bystro/vcf_utils/simulate_random_vcf.py
```diff
@@ -0,0 +1,65 @@
...
+    # Convert to pandas DataFrame for further processing
+    vcf_str = "\n".join(vcf_data)
```
poneill, line 53: if it makes sense to call "\n".join here then vcf_data is `list[str]` (rather than `str` as annotated).  Probably makes more sense to have it as `str`, meaning that you want to convert it to `str` around L48 or so

python/python/bystro/vcf_utils/simulate_random_vcf.py
```diff
@@ -0,0 +1,65 @@
...
+    """VCF simulator for testing analysis modules."""
+    # This is a first pass - could include more sophisticated choices for possible values in future
+    sample_ids = [f"SampleID{i+1}" for i in range(num_samples)]
+    header = HEADER_COLS + sample_ids
+    vcf_data = []
+    simulated_indices = []
+    vcf_data.append("\t".join(header))
+    for variant in range(num_vars):
+        # Use index simulator to generate random chrom,pos,ref,alt
+        random_chr, random_pos, random_ref, random_alt = generate_random_vcf_index()
```
poneill, line 30: if we just name these vars like we want them here, we don't have to reassign them in L32-36

python/python/bystro/vcf_utils/simulate_random_vcf.py
```diff
@@ -0,0 +1,65 @@
...
+    return vcf_df
+
+
+def write_sim_vcf_with_comments(vcf_data: list[str]):
+    """Add a specific number of comment lines to test header is removed properly."""
+    num_comment_lines = 107
+    comment_lines = [f"# Comment line {i}" for i in range(1, num_comment_lines + 1)]
+    sim_vcf_with_comments = "\n".join(comment_lines) + "\n" + "\n".join(vcf_data)
+    # Save the simulated VCF as a TSV file
+    with open("simulated.vcf", "w") as file:
```
poneill, line 64: this function is complecting business with logic with IO; we want to have them separated as much as possible.  Also we'll probably want to be able to pass in a filename rather than have it hardcoded

poneill (changes requested):

python/python/bystro/vcf_utils/simulate_random_vcf.py
cristinaetrv: I think '#CHROM' is how VCF is usually coded? The preprocess depends on that designation. Or did you mean something else by this?

python/python/bystro/vcf_utils/simulate_random_vcf.py
cristinaetrv: I'm returning simulated_indices here for downstream analysis (using the same index to generate a matching gnomad loadings index test)

python/python/bystro/vcf_utils/simulate_random_vcf.py
cristinaetrv: I think the final update didn't go in quickly enough but now vcf_data is list[str]

python/python/bystro/vcf_utils/simulate_random_vcf.py
cristinaetrv: Do you think it'd be worth separating into two functions? Add comment lines and write out vcf separately?

cristinaetrv:
Continued in #195

PR #197, opened 2023-08-02
ancestry bugfixes: handle prefixless chromosome names, fix pandas performance warning, make AncestryResponse msgspec compatible

### LINKED ISSUE #199: [Ancestry] Makes sure that ancestry model gives intelligible answers for corrupted VCFs
VCFs are required to have either 8 columns, or 9 to inf in the case that we have sample genotypes. The number of columns is fixed for all rows.

The ancestry module will rely on those genotypes to build a dosage matrix. If somehow the VCF file has on line 2 a different number of genotypes than on line 1, ancestry module will attempt to construct a ragged matrix and (presumably) loudly error.

We should provide a nicer error message.

poneill:
- canonicalize variants by prepending `chr` if not present
- remove now-redundant autosomal variant check
- we were previously adding dosage columns for missing variants in a casual way, causing a performance warning.  Do this in a more principled way.
- ensure all floats in AncestryResponse are raw floats, not np.float64

Closes #199

### DIFF
--- a/python/python/bystro/ancestry/ancestry_types.py
+++ b/python/python/bystro/ancestry/ancestry_types.py
@@ -62,7 +62,7 @@ unit_float_validator = [
 class ProbabilityInterval:
     """Represent an interval of probabilities."""
 
-    # we need these to be literal floats for msgpack serialization, not numpy floats or anything else.
+    # we need these to be literal floats for msgspec serialization, not numpy floats or anything else.
     lower_bound: float = field(converter=float, validator=unit_float_validator)
     upper_bound: float = field(converter=float, validator=unit_float_validator)
 
@@ -159,7 +159,7 @@ class AncestryResult:
     sample_id: str = field(validator=instance_of(str))
     populations: PopulationVector = field(validator=instance_of(PopulationVector))
     superpops: SuperpopVector = field(validator=instance_of(SuperpopVector))
-    # needs to be literal float for msgpack
+    # needs to be literal float for msgspec
     missingness: float = field(converter=float, validator=unit_float_validator)
 
 
--- a/python/python/bystro/ancestry/ancestry_types.py
+++ b/python/python/bystro/ancestry/ancestry_types.py
@@ -62,20 +62,17 @@ unit_float_validator = [
 class ProbabilityInterval:
     """Represent an interval of probabilities."""
 
-    lower_bound: float = field(validator=unit_float_validator)
-    upper_bound: float = field(
-        validator=[
-            instance_of(float),
-            attrs.validators.ge(0.0),
-            attrs.validators.le(1.0),
-        ]
-    )
+    # we need these to be literal floats for msgpack serialization, not numpy floats or anything else.
+    lower_bound: float = field(converter=float, validator=unit_float_validator)
+    upper_bound: float = field(converter=float, validator=unit_float_validator)
 
     def __attrs_post_init__(self) -> None:
         """Ensure interval is well-formed."""
         if self.lower_bound > self.upper_bound:
-            err_msg = f"""Lower bound must be less than or equal to upper bound.
-            Got: lower_bound={self.lower_bound}, upper_bound={self.upper_bound} instead."""
+            err_msg = (
+                f"Lower bound must be less than or equal to upper bound.  "
+                f"Got: lower_bound={self.lower_bound}, upper_bound={self.upper_bound} instead."
+            )
             raise ValueError(err_msg)
 
 
@@ -162,7 +159,8 @@ class AncestryResult:
     sample_id: str = field(validator=instance_of(str))
     populations: PopulationVector = field(validator=instance_of(PopulationVector))
     superpops: SuperpopVector = field(validator=instance_of(SuperpopVector))
-    missingness: float = field(validator=unit_float_validator)
+    # needs to be literal float for msgpack
+    missingness: float = field(converter=float, validator=unit_float_validator)
 
 
 @attrs.frozen(kw_only=True)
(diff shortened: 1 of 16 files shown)


python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -62,20 +62,17 @@ class AncestrySubmission:
...
 
-    lower_bound: float = field(validator=unit_float_validator)
-    upper_bound: float = field(
-        validator=[
-            instance_of(float),
-            attrs.validators.ge(0.0),
-            attrs.validators.le(1.0),
-        ]
-    )
+    # we need these to be literal floats for msgpack serialization, not numpy floats or anything else.
```
akotlar, line 65: msgpack -> msgspec :)

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -162,7 +159,8 @@ class AncestryResult:
     sample_id: str = field(validator=instance_of(str))
     populations: PopulationVector = field(validator=instance_of(PopulationVector))
     superpops: SuperpopVector = field(validator=instance_of(SuperpopVector))
-    missingness: float = field(validator=unit_float_validator)
+    # needs to be literal float for msgpack
```
akotlar, line 162: msgspec

python/python/bystro/ancestry/tests/test_ancestry_types.py
```diff
@@ -46,7 +46,9 @@ def test_AncestrySubmission_is_frozen():
 
 def test_ProbabilityInterval_accepts_valid_bounds() -> None:
     """Ensure we can instantiate, validate ProbabilityInterval correctly."""
-    ProbabilityInterval(lower_bound=0.1, upper_bound=0.9)
+    prob_int = ProbabilityInterval(lower_bound=0.1, upper_bound=0.9)
+    assert type(prob_int.lower_bound) is float  # for msgpack serialization
```
akotlar, line 50: msgspec

python/python/bystro/ancestry/train.py
```diff
@@ -136,12 +139,13 @@ def _parse_vcf_line_for_dosages(
     line: str, variants_to_keep: Container[Variant]
 ) -> tuple[Variant, list[int]] | None:
     # will throw ValueError if "PASS" not found, which is good
-    fields = line[: line.index("PASS")].split()
+    fields = line.split()
     variant = ":".join([fields[0], fields[1], fields[3], fields[4]])
-    assert_true("variant is a valid variant string", is_autosomal_variant(variant))
+    variant = variant if variant.startswith("chr") else "chr" + variant
+    assert_true(f"variant {variant} is a valid variant string", is_autosomal_variant(variant))
```
akotlar, line 145: Wondering whether we want to fold in is_autosomal_variant updates (can be a separate PR as well):

Wondering:
- Do we want to raise an assertion in this error or just return None? This function already has a way of communicating invalid sites: that is a None, so I'm just wondering if we need 2 mechanisms.

- We currently do a lot of processing on variant: splitting the line, rejoining the first field fields (skipping the id field) into a string, splitting it again in is_autosomal to check the components, again splitting it in _get_chr_from_variant, . Could you help me understand the benefit?
   - If we keep variant as a tuple (fields[0], fields[1], fields[3], fields[4]), or even better a named tuple, I think we get much better performance, with no downside. Pandas will happily take a tuple as a column name, since tuples are serializable.

akotlar (changes requested):
A comment change, the need for PASS / . check, and a note about the autosomal check

python/python/bystro/ancestry/train.py
poneill: The short answer is that the autosomal variant check no longer actually make sense here, so the easiest thing to do is simply take it out.  (It made sense in the original context that the function was written, but now that we're applying it to parsing VCFs in the wild it's redundant: if a variant isn't one we want to keep, checking that against `variants_to_keep` is enough to determine that.

To the broader point, the benefit of having this check operate on strings was that strings are, for better or worse, currently the canonical representation of variants in this module.  It is not a great representation, for reasons discussed offline, and I'm happy to change it at some point: a namedtuple or the like is probably fine.  But, conceptually trivial as it is, it's major implementation change that will affect every part of the ancestry module, even affecting serialized model artifacts.  Like the old joke about finding half a worm in your apple, the only thing worse than a bad implementation is good one that half-replaces a bad one.  So if/when we change it I'd like to change it once and for all after we figure out exactly what it needs to handle: SNPs or arbitrary variants, genome build info, 0- or 1-based indexing, rsids, &c.

Most broadly speaking, I'd prefer to postpone this entire question of how quickly we're checking for the validity of a variant until we know that the ancestry service is basically functioning.  If we cut the time for this check down to zero we'd save approximately 1% of the total processing time, which we're doing in pure python to begin with.  As it is, we'd get a ~2x improvement over the regex implementation, so 0.5% overall.  (Recall that to check the validity of the `pos` value we either have to convert from str to int or handle the str with a regex anyway.). I thought we had consensus earlier that current VCF parsing performance was broadly acceptable, and that if we need generic performance, as opposed to relief from specifically targeted bottlenecks, then python was the wrong tool for the job to begin with and we should just rewrite (or better, find a library, cf. https://docs.rs/vcf/latest/vcf/)

The aim of this PR is really just to get ancestry back up and running so we can proceed with the end-to-end testing.  Once we're happy with what the model is giving us, I'm happy to make it go blazing fast.  But there are always going to be miscellaneous plumbing issues like the msgspec thing the first time we stand any new kind of service up and I'd like to try to see the end of the tunnel before optimizing.

poneill:
> A comment change, the need for PASS / . check, and a note about the autosomal check

much obliged for catching the typo, PASS/. filter check implemented, autosomal check removed, see other comment

akotlar:
Confirmed that this addresses the observed issues.

PR #205, opened 2023-08-07
Add gnomad PC loadings into ancestry train

cristinaetrv:
Load and process 1KGP and gnomad PC loadings into ancestry train. This is a copy of a previous version to address additional commits being pulled into this PR.


poneill (changes requested):

poneill (approved):
lgtm!

PR #209, opened 2023-08-07
Fix performance regression in parse_vcf

### LINKED ISSUE #204: Benchmark Ancestry Performance
Currently, the ancestry module appears about 1-2 order of magnitude slower than either indexing or annotation, with annotation of 13,000 variants and 3 genotypes taking approximately 2 seconds, indexing about 10 seconds, and ancestry 45 seconds. Given the much lower IO requirements for ancestry vs either annotation or indexing, this is surprising. This performance gap also makes for a meaningfully worse user experience, and will also limit the kinds of queries users can execute as the user waits on ancestry to complete. 

If we scale this to even 1000 genomes size data, which is small by today's standards, how does ancestry calculation scale? What if we have Wingo-scale data (~20,000 genomes). Where is the performance bottleneck, VCF parsing, data frame manipulation, or RandomForestClassifier?
 - If VCF parsing, we should move to Rust vcf parser (we need this for Bystro anyway)
 - If pandas, we should try polars or stick with numpy directly
 - If RandomForestClassifier we should try XGBoost

What we learn here will be used to improve ancestry runtime performance and make future architectural decisions.

poneill:
We introduced a performance regression in parse_vcf by splitting the sample fields before we knew that we needed the variant represented by that row.  

- Fix this by parsing just enough of the line to check the variant and proceeding only if we need it.
- Add explicit timing info to logging to more easily diagnose future regressions.

Closes #204

### DIFF
--- a/python/python/bystro/ancestry/inference.py
+++ b/python/python/bystro/ancestry/inference.py
@@ -1,4 +1,5 @@
 """Classify genotypes at inference time."""
+import logging
 from dataclasses import dataclass
 from pathlib import Path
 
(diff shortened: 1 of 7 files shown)


akotlar (approved):

akotlar:
Approved modulo test errors

PR #210, opened 2023-08-14
Add combining gnomad pcs to 1kgp to ancestry code

cristinaetrv:


python/python/bystro/ancestry/tests/test_train.py
```diff
@@ -286,3 +288,72 @@ def test_1kgp_vcf_to_dosage():
...
+    """Tests restriction of gnomad loadings variants to vcf variants."""
+    num_samples = 10
+    num_vars = 10
+    simulated_vcf = generate_simulated_vcf(num_samples, num_vars)
+    sim_vcf_df = convert_sim_vcf_to_df(simulated_vcf)
+    loaded_vcf = convert_1kgp_vcf_to_dosage(sim_vcf_df)
+    processed_sim_vcf = process_vcf_for_pc_transformation(loaded_vcf)
+    # Using same indices, generate sim loadings
+    num_pcs = 30
+    sim_pcs = np.random.random((num_samples, num_pcs))
```
poneill, line 303: FYI technically sim_pcs is a random semi-orthogonal matrix.  You can simulate this by generating random data, PCAing it, and taking the `pca.components_` matrix.  I don't think it really matters here for what you're doing but if you extend this test later, you should be aware that a PC matrix has special structure that this matrix won't have.

python/python/bystro/ancestry/tests/test_train.py
```diff
@@ -286,3 +288,72 @@ def test_1kgp_vcf_to_dosage():
...
+        data=sim_pcs, index=processed_sim_vcf.index, columns=[f"PC{i+1}" for i in range(num_pcs)]
+    )
+    # Run restrict_loadings with sim data
+    pc_loadings_overlap, genos_overlap, num_var_overlap = restrict_loadings_variants_to_vcf(
+        sim_loadings, processed_sim_vcf
+    )
+    # Check for expected columns
+    expected_columns_loadings = ["PC1", "PC2"]
+    expected_columns_genos = ["SampleID1", "SampleID2"]
+    for column in expected_columns_loadings:
```
poneill, line 314: why not just check that the columns are equal up to permutation by doing something like:

`set(expected_columns_loadings) == set(pc_loadings_overlap.columns)`,

as you do below?

python/python/bystro/ancestry/train.py
```diff
@@ -560,6 +560,38 @@ def get_variant(x):
...
+    var_overlap = pc_loadings.index.intersection(genos.index)
+    pc_loadings_overlap = pc_loadings[pc_loadings.index.isin(var_overlap)]
+    genos_overlap = genos[genos.index.isin(var_overlap)]
+    # Ensure that genos transpose and pc_loadings have corresponding shape and vars
+    print("genos_overlap.index", genos_overlap.index)
+    print(pc_loadings_overlap.index)
+    assert genos_overlap.T.shape[1] == pc_loadings_overlap.shape[0]
+    assert set(genos_overlap.index) == set(pc_loadings_overlap.index)
+    assert (genos_overlap.columns == genos.columns).all()
+    assert "SampleID1" in genos_overlap.columns
```
poneill, line 577: will this always be true?

python/python/bystro/ancestry/train.py
```diff
@@ -560,6 +560,38 @@ def get_variant(x):
...
+    pc_loadings: pd.DataFrame, genos: pd.DataFrame
+) -> tuple[pd.DataFrame, pd.DataFrame, int]:
+    """Restrict variant list to overlap between gnomad loadings and reference vcf."""
+    # IGSR version of 1kgp is current reference vcf
+    var_overlap = pc_loadings.index.intersection(genos.index)
+    pc_loadings_overlap = pc_loadings[pc_loadings.index.isin(var_overlap)]
+    genos_overlap = genos[genos.index.isin(var_overlap)]
+    # Ensure that genos transpose and pc_loadings have corresponding shape and vars
+    print("genos_overlap.index", genos_overlap.index)
+    print(pc_loadings_overlap.index)
```
poneill, line 573: this should be a logging.debug rather than a print, (I think ruff should have caught this for you?)

python/python/bystro/ancestry/tests/test_train.py
```diff
@@ -286,3 +288,72 @@ def test_1kgp_vcf_to_dosage():
...
+        assert column in pc_loadings_overlap.columns
+    for column in expected_columns_genos:
+        assert column in genos_overlap.columns
+    # Check that the output DataFrames have the expected number of rows
+    expected_num_rows = num_var_overlap
+    assert pc_loadings_overlap.shape[0] == expected_num_rows
```
poneill, line 320: I think this test is a sign that we don't really need num_var_overlap in the return value of this function, because it's redundantly encoding the lengths of these dataframes.

python/python/bystro/ancestry/train.py
```diff
@@ -560,6 +560,38 @@ def get_variant(x):
     return pc_loadings
 
 
+def restrict_loadings_variants_to_vcf(
+    pc_loadings: pd.DataFrame, genos: pd.DataFrame
+) -> tuple[pd.DataFrame, pd.DataFrame, int]:
+    """Restrict variant list to overlap between gnomad loadings and reference vcf."""
```
poneill, line 566: we want to describe the return value here as well

python/python/bystro/ancestry/train.py
```diff
@@ -560,6 +560,38 @@ def get_variant(x):
...
+    assert "SampleID1" in genos_overlap.columns
+    # Record amount of overlap
+    num_var_overlap = len(var_overlap)
+    return pc_loadings_overlap, genos_overlap, num_var_overlap
+
+
+def apply_pca_transform(pc_loadings_overlap: pd.DataFrame, genos_overlap: pd.DataFrame) -> pd.DataFrame:
+    """Transform vcf with genotypes in dosage format with PCs loadings from gnomad PCA."""
+    # Dot product
+    transformed_data = genos_overlap.T @ pc_loadings_overlap
```
poneill, line 586: if we're doing this correctly then transformed_data should have the right index and columns by default, is there a reason why we have to set its metadata manually below?

poneill: this also appears to be the only time we're really using genos_overlap (as opposed to checking its properties) so would it make more sense to define genos_overlap as genos_overlap.T instead?

poneill (changes requested):

python/python/bystro/ancestry/train.py
cristinaetrv: Interestingly ruff didn't catch this for me and I ran ruff a few times in the process - is it a config option that maybe isn't set up? I could add it to my local dev version at the very least.

python/python/bystro/ancestry/tests/test_train.py
cristinaetrv: I think I wanted to have a readout of num_var_overlap at some point, but wasn't sure where to put it - probably best as part of a log instead though?

python/python/bystro/ancestry/train.py
cristinaetrv: There's a few places where genos_overlap is tested so it would either end up requiring a transpose step again or changing all those tests (which is easy to do but just wondering whether that's conceptually a better move than leaving the transpose step only in the place where it is needed as more of a general thought). I think setting the metadata was just a holdover from when it got converted to numpy first but no longer needed.

python/python/bystro/ancestry/tests/test_train.py
cristinaetrv: I think this is a little further than I'd like to go for this particular test - wasn't planning on making a super realistic simulated gnomad pc loadings since it's limited to just this one part of one module, as compared to a vcf sim that's closer to the real data to test various other analysis modules. Will re-evaluate if it ends up extending further.

python/python/bystro/ancestry/train.py
poneill: There's a pretty strong convention in statistics / ML to orient matrices so that observed entities correspond to rows, and features correspond to columns.  I'd try to honor that convention.

poneill (approved):

PR #212, opened 2023-08-21
Add module for loading / prepping fragpipe tandem mass tag datasets

poneill:
Add module to load and preprocess:

- abundance files
- annotation files

arising from fragpipe TMT experiments.

### DIFF
--- a/python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
+++ b/python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
@@ -1,0 +1,49 @@
+from dataclasses import dataclass
+import pandas as pd
+from typing import TypeVar, Tuple
+
+ABUNDANCE_COLS = ["Index", "NumberPSM", "ProteinID", "MaxPepProb", "ReferenceIntensity"]
+ANNOTATION_COLS = ["plex", "channel", "sample", "sample_name", "condition", "replicate"]
+
+T = TypeVar("T")
+
+
+@dataclass(frozen=True)
+class TMTDataset:
+    abundance_df: pd.DataFrame
+    annotation_df: pd.DataFrame
+
+
+def is_prefix(xs: list[T], ys: list[T]) -> bool:
+    for x, y in zip(xs, ys, strict=False):
+        if x != y:
+            return False
+    return True
+
+
+def parse_abundance_df(abundance_df: pd.DataFrame) -> pd.DataFrame:
+    if not is_prefix(ABUNDANCE_COLS, abundance_df.columns):
+        raise ValueError
+    abundance_df = abundance_df.set_index("Index")
+    first_sample_column_idx = abundance_df.columns.to_list().index("ReferenceIntensity") + 1
+    sample_columns = abundance_df.columns[first_sample_column_idx:]
+    for sample_column in sample_columns:
+        abundance_df[sample_column] -= abundance_df.ReferenceIntensity
+    return abundance_df
+
+
+def parse_annotation_df(annotation_df: pd.DataFrame) -> pd.DataFrame:
+    if not is_prefix(ANNOTATION_COLS, annotation_df.columns):
+        raise ValueError
+
+    return annotation_df.set_index("sample")
+
+
+def parse_tandem_mass_tag_files(
+    abundance_filename: str, annotation_filename: str
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    raw_abundance_df = pd.read_csv(abundance_filename, sep="\t")
+    raw_annotation_df = pd.read_csv(annotation_filename, sep="\t")
+    abundance_df = parse_abundance_df(raw_abundance_df)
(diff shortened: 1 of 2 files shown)


python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
```diff
@@ -0,0 +1,53 @@
...
+
+    abundance_df: pd.DataFrame
+    annotation_df: pd.DataFrame
+
+
+def _check_df_cols(df: pd.DataFrame, expected_cols: list[str]) -> None:
+    actual_cols = df.columns
+    if not all(x == y for x, y in zip(expected_cols, actual_cols, strict=False)):
+        err_msg = (
+            f"expected dataframe to begin with cols: {expected_cols}, got cols: {actual_cols} instead."
```
cristinaetrv, line 23: General question - is there an overlapping core set of columns between outputs from different fragpipe pipelines or will we have to be very specific about what is allowed as input?

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def _prep_abundance_df(abundance_df: pd.DataFrame) -> pd.DataFrame:
+    """Prep abundance_df, setting index and normalizing abundances by ReferenceIntensity."""
+    _check_df_cols(abundance_df, ABUNDANCE_COLS)
+    abundance_df = abundance_df.set_index("Index")
+    first_sample_column_idx = abundance_df.columns.to_list().index("ReferenceIntensity") + 1
+    sample_columns = abundance_df.columns[first_sample_column_idx:]
+    for sample_column in sample_columns:
+        abundance_df[sample_column] -= abundance_df.ReferenceIntensity
```
cristinaetrv, line 35: Is this a standard normalization step? I thought they were normalized in a different way, but just checking. As a side note, is it possible we want to support different possible normalization schemes as a small analysis step and leave them unnormalized for the basic report people will search through?

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
poneill: The annotation files look to be conserved between pipelines, but the abundance filenames look to be pretty sui generis

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
```diff
@@ -0,0 +1,53 @@
...
+
+import pandas as pd
+
+ABUNDANCE_COLS = ["Index", "NumberPSM", "ProteinID", "MaxPepProb", "ReferenceIntensity"]
+ANNOTATION_COLS = ["plex", "channel", "sample", "sample_name", "condition", "replicate"]
+
+
+@dataclass(frozen=True)
+class TandemMassTagDataset:
+    """Represent a Fragpipe Tandem Mass Tag dataset."""
```
cristinaetrv, line 13: It seems like there are several TMT outputs from FragPipe - do they all have the same type of output files and headings or should we be more specific here (i.e. TMT-10, TMT16)?

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
poneill: If we don't normalize in this way, then ReferenceIntensity explains something like 90% of the variance of each column.  I'm of course open to doing it in some other way, but what would we do instead?

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
poneill: The TMT dataset we have access to is described as 4-plex, but in the paper, it seems, it's described as TMT-10 (https://www.sciencedirect.com/science/article/pii/S0092867419311237?via%3Dihub).  I'm not sure how to resolve that apparent tension, but in any case this is the only fragpipe TMT dataset we have to work with at present.  

I can doublecheck this.

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
cristinaetrv: Looks like there's quite a few normalization methods - just wanted to be sure we were going with a standard choice, would be worth conferring with someone else about this. Might also make more sense as a downstream analysis step that happens after annotation? Are we reporting variance in the annotation?

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
cristinaetrv: Is it possible we'll be able to capture at least most/many TMT pipeline outputs with these headers? May be a question we send out to our contacts.

cristinaetrv (approved):
Overall, lgtm! I think we should be careful about which specific calculations/analysis to include during the annotation but that's my only concern at this point. Let's chat about this in another medium.

PR #213, opened 2023-08-22
Add loading / prep code for fragpipe DIA datasets

poneill:
- Load and prep pg_matrix files from Fragipipe DIA workflow
- Refactor annotation_df loading, which is shared by all fragpipe workflows


python/python/bystro/proteomics/fragpipe_utils.py
```diff
@@ -0,0 +1,21 @@
+"""Provide various utilities for Fragpipe analysis."""
```
cristinaetrv, line 1: nit: Are these for Fragpipe analysis or for ingesting/processing Fragpipe output files?

python/python/bystro/proteomics/fragpipe_utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+import pandas as pd
+
+ANNOTATION_COLS = ["plex", "channel", "sample", "sample_name", "condition", "replicate"]
+
+
+def check_df_cols(df: pd.DataFrame, expected_cols: list[str]) -> None:
+    """Check that df cols contain expected cols as a prefix."""
+    actual_cols = df.columns
+    if not all(x == y for x, y in zip(expected_cols, actual_cols, strict=False)):
```
cristinaetrv, line 11: Will expected_cols contain the minimal set of column headers that we're expected for a given proteomics 'annotation' file? I may be missing context but how/where will we handle columns specific to an annotation type file that are experiment-specific (will this type of information be separated into subject-level information for instance)?

python/python/bystro/proteomics/tests/test_fragpipe_utils.py
```diff
@@ -0,0 +1,25 @@
+import re
+import pytest
+
+import pandas as pd
+
+from bystro.proteomics.fragpipe_utils import check_df_cols
+
+
+def test_check_df_cols_happy_path():
```
cristinaetrv, line 9: Could you add docstrings here and throughout?

python/python/bystro/proteomics/tests/test_annotation_df.py
```diff
@@ -0,0 +1,27 @@
+import pandas as pd
```
cristinaetrv, line 1: Could you describe how this is testing annotation_df or change the name of it to improve clarity?

cristinaetrv (approved):
No blocking changes, but would appreciate more clarity/docstrings in some areas. Otherwise looks good as a starting place for ingesting these files!

akotlar:
Are we ok to merge this?

python/python/bystro/proteomics/tests/test_fragpipe_utils.py
poneill: So while I agree in general about docstrings, I'd be specifically wary of making them mandatory in tests, because the primary problem with testing is getting people to write them at all, and so things that increase the friction of writing tests can backfire.  

Happy to add them in this PR but I think it's worth getting general consensus on about whether to require them.

python/python/bystro/proteomics/tests/test_annotation_df.py
poneill: yep, that's pretty unclear now that I look at it, refactored away

python/python/bystro/proteomics/fragpipe_utils.py
poneill: That's a good question and the short answer is I don't know yet, but I think it falls out of scope for this method, which is doing something more modest.  I've renamed the method though to try to make it clearer what it's supposed to be doing, though.

poneill: We definitely don't have the subject level info story completely worked out yet: for now everything is more or less its own independent sample

PR #215, opened 2023-08-27
Supervised ppca

austinTalbot7241993:


poneill:
This looks like a lot of great work.  My one concern is that it's about 20x the recommended PR size, and is going to be difficult to give it the attention it deserves in review.  Can we break this up into chunks so that we can review it piecemeal?  

I looked a bit into solutions for automating type-hinting-- you might find this package useful: https://pypi.org/project/sphinx-autodoc-typehints/.  Since you already mostly have type info in the docstrings, the package can help to automate the work of sticking that into type annotations as well so they can be statically verified.  As it stands, it looks like all of the mypy error messages pertain to missing stubs for libraries-- once those errors are resolved by installing the relevant typeshed packages it should be easier to see how much remaining mypy work is required to get this passing.

austinTalbot7241993:
@poneill

akotlar:
Working on actual code, but noticing that we have a dependency incompatibility. torch==1.13.1 needs to be torch==2.0.*, :

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
xformers 0.0.20 requires torch==2.0.1, but you have torch 1.13.1 which is incompatible.
torchvision 0.15.2+cu118 requires torch==2.0.1, but you have torch 1.13.1 which is incompatible.
torchaudio 2.0.2+cu118 requires torch==2.0.1, but you have torch 1.13.1 which is incompatible.
```

python/python/bystro/_template_npr.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+
+        """
+        assert self._model is not None, "Fit model first"
```
akotlar, line 59: @poneill what do you prefer for user-facing errors in situations like this? Pep suggests RuntimeError afaik?

python/python/bystro/_template_npr.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        default_options = {
+            "num_chains": 1,
+            "num_warmup": 500,
+            "num_samples": 2000,
+        }
+        mcmc_opts = deepcopy(default_options)
+        mcmc_opts.update(mcmc_options)
+        if mcmc_opts["num_chains"] > 1:
+            print("You tried to set num_chains>1. This has been ignored.")
```
akotlar, line 107: I think we may not want this to be generically true for all numpyro methods.

austinTalbot7241993: I would rather manually force an override after model creation.

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+    ----------
+    covariance : np.array-like(p,p)
+        The covariance matrix
+
+    Returns
+    -------
+    precision : np.array-like(sum(p),sum(p))
+        The inverse of the covariance matrix
+    """
+    precision = la.inv(covariance)
```
poneill, line 180: I can tell from the documentation above that you've already thought about this, but are we sure we need to invert a (p, p) matrix?  What are typical values of p in this case?  Depending on what we're doing with the covariance matrix downstream, we may be able to get away with doing something else that's more numerically stable.

Not a blocking comment rn, just curious

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+    ||A||_F^2/||A||^2
+
+    Parameters
+    ----------
+    covariance : np.array-like(p,p)
+        The covariance matrix
+
+    Returns
+    -------
+    srank : float
```
poneill, line 196: nit: is this abbreviation an established convention in numerical linear algebra?  my own preference would be to spell it out for the benefit of reading speed.

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+        The stable rank
+    """
+    singular_values = la.svd(covariance, compute_uv=False)
+    srank = np.sum(singular_values) / singular_values[0]
+    return srank
+
+
+def _predict(covariance, Xobs, idxs):
+    """
+    Predicts missing data using observed data.
```
poneill, line 206: I think it would be really helpful here if we could reference a specific equation in a paper that we're implementing here, or otherwise gloss what we're doing.  I think I recognize terms from the calculation of the conditional density of a mv gaussian, but I'm getting lost at la.solve-- an additional breadcrumb would really seal the deal.

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+    covariance : np.array-like(p,p)
+        The covariance matrix
+
+    X : np.array-like,(N,sum(idxs))
+        The data
+
+    idxs: np.array-like,(sum(p),)
+        The observation locations
+
+    weights : np.array-like,(N,),default=None
```
poneill, line 266: could we specify what we expect in terms of weights here?  if the user passes np.ones(N) vs np.ones(N) / N they'll get different answers

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+    covariance_22 = covariance_sub[:, idxs == 1]
+    covariance_21 = covariance_sub[:, idxs == 0]
+
+    covariance_nonsub = covariance[idxs == 0]
+    covariance_11 = covariance_nonsub[:, idxs == 0]
+
+    beta_bar = la.solve(covariance_22, covariance_21)
+    Second_part = np.dot(covariance_21.T, beta_bar)
+
+    covariance_11 - Second_part
```
poneill, line 234: It looks like this line isn't doing anything?

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+
+    idxs: np.array-like,(p,)
+        The observation locations
+
+    Returns
+    -------
+    scores : float
+        Log likelihood for each sample
+    """
+    covariance_sub = covariance[idxs == 1]
```
poneill, line 305: should this block be refactored in light of the `_predict` method above?

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+
+    covariance_bar = covariance_11 - Second_part
+
+    mu_ = np.dot(X[:, idxs == 1], beta_bar)
+    scores = _score_samples(covariance_bar, X[:, idxs == 0] - mu_)
+
+    return scores
+
+
+def _conditional_parameters(covariance, idxs):
```
poneill, line 323: unless a method is a @property, it's generally preferred to begin them with a verb, e.g. `_get_conditional_parameters` or similar

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+    Returns
+    -------
+    beta_bar : array-like
+        The predictive covariates
+
+    covariance_bar : array-like
+        Conditional covariance
+    """
+
+    covariance_sub = covariance[idxs == 1]
```
poneill, line 344: looks like we should really factor this block out

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+    scores : float
+        Log likelihood for each sample
+    """
+    p = covariance.shape[0]
+    term1 = -p / 2 * np.log(2 * np.pi)
+
+    _, logdet = la.slogdet(covariance)
+    term2 = -0.5 * logdet
+
+    quad_init = la.solve(covariance, np.transpose(X))
```
poneill, line 460: could we just get a brief gloss of what's going on in these three lines?  AFAICT we're just calculating the log-density of X, but I think the modal engineer would have to stop and think for a while about why this works.

Also, here and throughout, are we guaranteed that X is already centered?

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+    ----------
+    covariance : np.array-like(p,p)
+        The covariance matrix
+
+    Returns
+    -------
+    entropy : float
+        The differential entropy of the distribution
+    """
+    covariance.shape[1]
```
poneill, line 492: this line looks like it's a no-op?  FYI ruff should be able to detect these for you, I'll have to doublecheck why it doesn't seem to be picking this one up

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+    cov_new = 2 * np.pi * np.e * covariance
+    _, logdet = la.slogdet(cov_new)
+    entropy = 0.5 * logdet
+    return entropy
+
+
+def _entropy_subset(covariance, idxs):
```
poneill, line 499: should this be `calculate_marginal_entropy` or something instead?

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+        The mutual information between the two variables
+    """
+    idxs = idxs1 + idxs2
+    cov_sub = covariance[np.ix_(idxs == 1, idxs == 1)]
+    idxs1_sub = idxs1[idxs == 1]
+    Hy = _entropy(cov_sub[np.ix_(idxs1_sub == 1, idxs1_sub == 1)])
+
+    _, covariance_conditional = _conditional_parameters(cov_sub, 1 - idxs1_sub)
+    Hyx = _entropy(covariance_conditional)
+    mutual_information = Hy - Hyx
```
poneill, line 548: This line ended up being surprising to me because I instinctively read `Hyx` as the joint entropy of X and Y rather than the conditional entropy H(y|x).  Could we rename to `H_y_given_x` or similar?

python/python/bystro/covariance/_covariance_np.py
```diff
@@ -0,0 +1,96 @@
...
+implements MAP estimation using several common priors.
+
+Objects
+-------
+EmpiricalCovariance(BaseCovariance)
+    
+
+"""
+import numpy as np
+from ._base_covariance import BaseCovariance
```
poneill, line 13: I believe the house style is to use absolute imports, i.e. `from bystro.covariance._base_covariants import BaseCovariance`

python/python/bystro/covariance/_covariance_np.py
```diff
@@ -0,0 +1,96 @@
...
+        This fits a covariance matrix using samples X.
+
+        Parameters
+        ----------
+        X : np.array-like,(n_samples,n_covariates)
+            The data
+        """
+        self.N, self.p = X.shape
+        XTX = np.dot(X.T, X)
+        self.covariance = XTX / self.N
```
poneill, line 36: Would np.cov not work as a drop-in replacement here?  (And do we want to divide by N or N-1 here?)  Does this implementation work for non-centered matrices?  

I think I'm just confused about what X is and what properties we can expect it to have, could we gloss that at the beginning of the module

python/python/bystro/covariance/_covariance_np.py
```diff
@@ -0,0 +1,96 @@
...
+        self.N, self.p = X.shape
+        p_opts = self.prior_options
+        covariance_empirical = np.dot(X.T, X)
+        if p_opts["type"] == "inverse-wishart":
+            nu = p_opts["iw_params"]["pnu"] + self.p
+            cov_prior = p_opts["iw_params"]["sigma"] * np.eye(self.p)
+            posterior_cov = cov_prior + covariance_empirical
+            posterior_nu = nu + self.N
+            self.covariance = posterior_cov / (posterior_nu + self.p + 1)
+        else:
```
poneill, line 69: if we're only implementing the inverse-wishart type, can we just drop the type key from the prior_options dict altogether?  Is there some existing convention we have to honor?

python/python/bystro/covariance/_covariance_np.py
```diff
@@ -0,0 +1,96 @@
...
+
+        Parameters
+        ----------
+        prior_options : dict
+            The original prior options passed as a dictionary
+
+        Options
+        -------
+        type : str, default='inverse-wishart'
+            The type of prior. Currently one lol
```
poneill, line 84: I think this comment speaks to the fact that this option is kind of crying out to be removed :)

python/python/bystro/covariance/_precision_np.py
```diff
@@ -0,0 +1,24 @@
...
+    def fit(self, X):
+        """
+        This fits a precision matrix using samples X.
+
+        Parameters
+        ----------
+        X : np.array-like,(n_samples,n_covariates)
+            The data
+        """
+        self.N, self.p = X.shape
```
poneill, line 22: same questions as before about np.cov

poneill (changes requested):
A lot of great work here!  LMK if it would be helpful to discuss the comments offline

python/python/bystro/_template_npr.py
poneill: Yes, I left a comment about this too but it looks like github ate it.  

In general we should code as if our asserts could be turned off at any time, because they can.  That means that asserts are basically only good for catching errors that should *never* happen (as an aid to development / debugging) and communicating invariants to the reader.  If we have to check for an error that could happen for reasons beyond the code's control, such as ValueErrors or a user using a class incorrectly, we should check explicitly and raise the appropriate exception (here, a RuntimeError).

python/python/bystro/_template_npr.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+
+        """
+        self.mcmc_options = self._fill_mcmc_options(mcmc_options)
+        self.hp_options = self._fill_hp_options(hp_options)
```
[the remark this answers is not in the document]
austinTalbot7241993: hyperparameters

python/python/bystro/covariance/_covariance_np.py
austinTalbot7241993: OK will change

python/python/bystro/covariance/_covariance_np.py
austinTalbot7241993: Yeah I think so. The reason for X is it allows for some stochasticity, which I like in tests like this.

python/python/bystro/covariance/_covariance_np.py
austinTalbot7241993: I just need to add more options then.

python/python/bystro/_template_npr.py
```diff
@@ -0,0 +1,121 @@
...
+            "num_samples": 2000,
+        }
+        mcmc_opts = deepcopy(default_options)
+        mcmc_opts.update(mcmc_options)
+        if mcmc_opts["num_chains"] > 1:
+            print("You tried to set num_chains>1. This has been ignored.")
+            mcmc_opts["num_chains"] = 1
+        return mcmc_opts
+
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
austinTalbot7241993: Because aspects of MCMC inference are broadly applicable across different types of models but hyperparameters are not.

python/python/bystro/_template_npr.py
```diff
@@ -0,0 +1,121 @@
...
+    This is the template Numpyro model
+
+"""
+import abc
+import cloudpickle # type: ignore
+import numpyro # type: ignore
+from copy import deepcopy
+
+
+class _BaseNPRModel(object):
```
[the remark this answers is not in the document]
austinTalbot7241993: What would you want to change for that?

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: What do you mean?

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Yes

python/python/bystro/_template_npr.py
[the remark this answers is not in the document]
austinTalbot7241993: Removed

python/python/bystro/_template_sgd_np.py
```diff
@@ -0,0 +1,133 @@
...
+        -------
+        training_opts : dict
+        """
+        default_options = {"n_iterations": 5000}
+        training_opts = deepcopy(default_options)
+        training_opts.update(training_options)
+        return training_opts
+
+    @abc.abstractmethod
+    def _save_variables(self, training_variables):
```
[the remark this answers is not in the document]
austinTalbot7241993: To make life easier for me I save out parameters as numpy arrays rather than the pytorch internal state.

python/python/bystro/covariance/_base_covariance.py
```diff
@@ -0,0 +1,549 @@
...
+import pytz
+
+class BaseCovariance(object):
+    """
+    This object basically contains all the methods asides for fitting
+    a covariance matrix.
+    """
+
+    def __init__(self):
+        self.creationDate = dt.now(pytz.timezone("US/Pacific"))
```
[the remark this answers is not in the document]
austinTalbot7241993: Yes thanks to ruff.

python/python/bystro/_template_sgd_np.py
```diff
@@ -0,0 +1,133 @@
...
+    def pickle(self, path):
+        """
+        Method for saving the model
+
+        Parameters
+        ----------
+        path : str
+            The directory to save the model to
+        """
+        mydict = {"model": self}
```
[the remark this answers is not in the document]
austinTalbot7241993: No because it's convenient to save the model object, but what we'll need to load in are the saved variables and set them as model attributes.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Unfortunately in statistics often times the explicit estimates of the precision matrix are of interest for interpretation. Certainly though if you want to evaluate the likelihood this is a bad idea.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Done.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Done.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Non negative and made those examples equivalent.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: No,  pred method just returns means, not covariance which is needed for likelihood.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Done.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Can't for reason mentioned above. Unfortunately inverses don't work out the way you think that would allow this.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Or more accurately that's the requirement, maybe should state explicitly.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: I like the name but if we insist on changing it I will.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Done.

python/python/bystro/covariance/_covariance_np.py
austinTalbot7241993: Done.

python/python/bystro/covariance/_covariance_np.py
austinTalbot7241993: Just removed.

python/python/bystro/covariance/_precision_np.py
austinTalbot7241993: I guess I like doing things myself.

python/python/bystro/_template_npr.py
```diff
@@ -0,0 +1,121 @@
...
+object extending the template, namely
+    fit
+    _fill_hp_options
+
+These are the methods for running the samples given data and providing 
+hyperameter selections respectively.
+
+Objects
+-------
+_BaseNPRModel(object)
```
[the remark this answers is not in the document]
austinTalbot7241993: Done.

python/python/bystro/covariance/_covariance_np.py
poneill: Sorry, could you unpack that a bit more? I'm not sure I'm following.  My questions (consolidating this thread and the other one on the same topic) are:

1.  This implementation only computes the covariance matrix if X is already centered, i.e. E[X_i] = E[X_j] 0 for all rows i and columns j.  Are we guaranteed that X will always be centered?  If so, can we indicate that somehow in the code, like with a pre-condition check or even a descriptive variable name?  If X is not guaranteed to be centered but this is the implementation we want, should we call it something like `non_centered_covariance_matrix` instead?  Maybe it's a terminological confusion between fields but I've always heard of covariance matrices referring to the centered version, and if the opposite convention holds here it seems noteworthy enough to document.

2.  If we do want the centered version, why not just use `np.cov`?  I'm not saying that we should use libraries for everything, but numpy seems like a safe assumption as a dependency for an ML / stats module.  If we're doing something by hand when there's a very natural approach with a library function, it raises a question in the reader's mind why we're doing something by hand instead and what the difference might be between our approach and the library function.  In this case, there is in fact a difference (the non-centered vs. centered issue) and if that difference is intentional I think we want to signpost it to the reader.
3.  We're dividing by N rather than (N-1), meaning that our estimates of the covariance will be biased (though not by much if N is large).  I'm not even worried so much about the bias as about the difficulty the next engineer might have in interpreting the intent behind the choice of biased estimator.

Are these questions roughly checking out?  I'm not as familiar with covariance matrix estimation so I might just be misunderstanding the intent here, but I think these are questions the modal engineer might also have when reading this code.

austinTalbot7241993: np.std uses the biased estimator as a default, I assumed that np.cov did too but apparently I was wrong. Either way, by writing stuff out there's no ambiguity. I'm making it clear that we should expect centered data. Should I make an assert statement? Also for (3), since we are using maximum likelihood for the remaining approaches might as well stay consistent. The biased version is the ML version.

Also centering is only applicable in the columns.

python/python/bystro/_template_sgd_np.py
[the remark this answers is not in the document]
poneill: Sorry, I'm still not sure I'm following here: if we're dumping the model to disk this way, why not just pick up with `cloud_pickle.load` and return `my_dict['model']` on analogy to the way that we're dumping it?  I might be missing something here but why would that be a reasonable default implementation?  Is there a handy example of when we wouldn't want that behavior?

python/python/bystro/_template_sgd_np.py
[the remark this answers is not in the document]
poneill: Thanks, I get that idea and like it, I think there's something else that's confusing me here, maybe I can try to restate it another way: `_save_variables` just takes a list of variables to save, but it doesn't take a path, so this method can't know where to save `training_variables` unless that path is somewhere else in its global state (but I don't see it in the instance attrs, which is only `training_options`?).  Shouldn't the implementor of this method in a subclass need a filepath to tell the subclass where to save them?

python/python/bystro/covariance/_base_covariance.py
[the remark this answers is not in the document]
poneill: Ah I see what's going on: https://beta.ruff.rs/docs/rules/call-datetime-now-without-tzinfo/

I think the preferred solution (from the docs) is: `datetime.datetime.now(tz=datetime.timezone.utc)`.

Also not to nitpick too much but `creationDate` should be `creation_date`

python/python/bystro/covariance/_base_covariance.py
poneill: Appreciate it, but have you pushed the latest commit?  I'm still seeing `srank`on my end

python/python/bystro/covariance/_base_covariance.py
poneill: Here are six different places in the code where we're repeating the same pattern line for line:

```python

    covariance_sub = covariance[idxs == 1]
    covariance_22 = covariance_sub[:, idxs == 1]
    covariance_21 = covariance_sub[:, idxs == 0]
    beta_bar = la.solve(covariance_22, covariance_21)
###
    covariance_sub = covariance[idxs == 1]
    covariance_22 = covariance_sub[:, idxs == 1]
    covariance_21 = covariance_sub[:, idxs == 0]
    beta_bar = la.solve(covariance_22, covariance_21)

###

    covariance_sub = covariance[idxs == 1]
    covariance_22 = covariance_sub[:, idxs == 1]
    covariance_21 = covariance_sub[:, idxs == 0]
    beta_bar = la.solve(covariance_22, covariance_21)

####
    csub = cov[idxs == 1]
    c22 = csub[:, idxs == 1]
    c21 = csub[:, idxs == 0]
    csub = cov[idxs == 0]
    beta_bar = la.solve(c22, c21)

####
    csub = cov[idxs == 1]
    c22 = csub[:, idxs == 1]
    c21 = csub[:, idxs == 0]
    beta_bar = la.solve(c22, c21)


####
    csub = cov[idxs == 1]
    c22 = csub[:, idxs == 1]
    c21 = csub[:, idxs == 0]
    beta_bar = la.solve(c22, c21)

```

My concerns are that:

1. it will be easy for one of these instances of the pattern to get out of synch with the others, and once it gets out of synch it will be difficult to tell that it's out of synch precisely because it's repeated so many times.

2.  A pattern repeated that many times probably has a pre-existing name.  If we pull it out into a helper method and call it by its proper name, we can help the reader out.  I don't think it's quite self-evident to the modal engineer what the intent is here (using `la.solve` in order to compute a log normal density isn't really elementary), so a named function would give them something to google.

python/python/bystro/covariance/_base_covariance.py
poneill: Could you say more about that (maybe in the thread above for consolidation): https://github.com/bystrogenomics/bystro/pull/215/files#r1315087573?  I'm worried that we're talking past each other here (e.g. I'm not claiming that inverses work in any particular way, I'm just noticing a pattern of code duplication).

python/python/bystro/covariance/_base_covariance.py
poneill: I think it would really help to state that explicitly somewhere, because we're making a lot of assumptions that X is centered throughout the code but AFAICT never checking it explicitly.

python/python/bystro/covariance/_covariance_np.py
```diff
@@ -0,0 +1,90 @@
...
+
+class EmpiricalCovariance(BaseCovariance):
+    def __init__(self):
+        """
+        This object just fits the covariance matrix as the standard sample
+        covariance matrix
+        """
+        super().__init__()
+
+    def fit(self, X):
```
poneill, line 28: for all public methods where we accept an `X`, what would you think about checking that X is centered and raise an informative ValueError if not?

austinTalbot7241993: What should that check be? standardScalar for example only gets it to 1e-16 within 0 usually, and we dont' assume normalization. If the mean is .0001 for a single covariate does it fail?

austinTalbot7241993: Thought I responded but didn't save for some reason. What check do you have in mind, since sklearn's standardscaler often gives values like 1e-16 for covariates. Is mean of .0001 too much if it's in a single covariate? What if the variance in that covariate is .001 instead of approximately 1? In my strong opinion this is one of those things where we have to trust the user to properly use the product.

poneill: I think whether normal machine precision is acceptable for this application is an implementation detail I'd defer to your judgment on, my concern is more about gross departures from centeredness.  A reasonable compromise might be checking that `X` is `np.close` to its StandardScaled version.

Want to discuss this stuff on a call tomorrow?  I'm worried I just don't have enough context yet to understand the approach.

austinTalbot7241993: Sure. But here's a straightforward application where np.close will fail, if you standardize a dataset before dividing it into a training and test set.

python/python/bystro/covariance/_covariance_np.py
poneill: I think where I land here is that if we have a class called EmpiricalCovariance, that class needs to compute empirical covariances.

Since empirical covariance is a well-defined mathematical concept that is unlikely to change over the lifetime of the project, my recommendation (as I understand the intent of your code) would be use `np.cov` or to implement the more or less one-line formula as it appears in stats textbooks, so that callers of this class don't get surprised.

If this class only computes empirical covariances for centered matrices and not for others, and this is correct and necessary for your purposes, then:

1.  The restricted domain should be reflected in the class name, because it's no longer computing the empirical covariance as every textbook defines it, but some other quantity.  Maybe we could call it `CenteredEmpiricalCovariance` or something.
2. As per a previous comment, if we're writing things out that are almost but not exactly equivalent to common library functions, that actually introduces ambiguity because it invites the reader to wonder, as I did, why we're not either using a common library functions or fully implementing the well-known formula for empirical covariance ourselves.  But given all that we've written back and forth here trying to understand each other, I think it's at least worth a comment in the code.  The next engineer will not even have the benefit of this record of correspondence.
3. If the `fit` method of the `EmpiricalCovariance` class receives non-centered input, it will not compute the empirical covariance of that input, but silently return some other output instead.  I would recommend it just refuse to calculate anything in those cases.  The conventional way to do that is to raise a ValueError.  If we check for value errors with asserts, we risk silent failures as discussed in another thread on this PR.

I want to leave open the possibility that I've just entirely misunderstood the intent of this code, in which case I apologize, but please consider this as a good faith attempt to understand the intent here and consider that future engineers may have similar questions.

Thank you for noting the typo, corrected above.

poneill (approved):
I am unblocking this PR with all remaining comments reserved to the author's best judgment.

python/python/bystro/_template_sgd_np.py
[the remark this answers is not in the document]
austinTalbot7241993: It's just easier that way, so you can say my_model = Model();my_model.unpickle('path') and all the attributes of my_model will be set.

python/python/bystro/covariance/_base_covariance.py
austinTalbot7241993: Sigma_11^{-1} is not the same as Phi_11 where Phi=Sigma^{-1}. Can test out with a matrix generated by ortho_group where Sigma_11^{-1} != I = Phi_11

PR #216, opened 2023-08-28
Feature/tada mixture model

austinTalbot7241993:
This provides a basic implementation of multivariate tada using a Poisson likelihood with a mixture model (also includes a ZIP option). There are two inference methods, maximizing the marginal likelihood using stochastic gradient descent, and expectation maximziation.


python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+
+        N, p = X.shape
+
+        model = GaussianMixture(K)
+        model.fit(X)
+        Lambda_list_l = []
+        for i in range(K):
+            Lambda_list_l.append(
+                torch.tensor(
+                    1.2 * model.means_[i].astype(np.float32) + 0.1,
```
akotlar, line 221: Could you help me understand this part; should we have 1.2 as a variable / model hyperparameter? I think I understand the 0.1 to be a way of getting away from P(X=0|lambda=0) = 1

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+        data = torch.tensor(X.astype(np.float32))
+        pi_ = 1 / K * np.ones(K)
+        myloss = PoissonNLLLoss(log_input=False, full=True, reduction="none")
+
+        for i in myrange(td["n_iterations"]):
+            # E step compute most likely categories
+            Zprobs = np.zeros((N, K))
+            for k in range(K):
+                Lambda_k = Lambda_list_l[k]
+                alpha_k = 0.8 * sigmoid(alpha_list_l[k])  # Prob of being 0
```
akotlar, line 247: could you help me understand the need to use sigmoid here? just a knowledge gap on my part

akotlar (approved):
Austin, in the spirit of iterations and not holding you up, approving this. I will test more in depth later, and bring up any issues if I find them

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+        tops : dict
+        """
+        default_options = {"n_iterations": 100}
+        tops = deepcopy(default_options)
+        tops.update(training_options)
+        return tops
+
+
+class MVTadaZPoissonEM(object):
+    def __init__(self, K=4, training_options={}):
```
poneill, line 182: `training_options={}` is a bit of a footgun / language wart in python, because if it's modified in one object it will be modified by default in all others, e.g.:

```python
class Foo:
    def __init__(self, my_dict={}):
        self.my_dict = my_dict


a = Foo()
a.my_dict["hello"] = "world"


b = Foo()
print(b.my_dict)
# {'hello': 'world'}
```

a safer way would be to write something like:
```python
def __init__(training_options=None):
    if training_options is None:
        training_options = {}
```

I'm not sure if this actually causes a problem in this class at the moment, but avoiding initializing vars to mutable collections like lists or dicts will help to keep it safe in the future.

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+            GMM on the data and using those.
+
+        pi_init : np.array-like,shape=(k),default=None
+            Initialize the class weights of the model.
+
+        Returns
+        -------
+        self
+        """
+        td = self.training_options
```
poneill, line 82: nit: I think it's fine to use one letter variables for math modeling, but what's `td` short for here?

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+                Zprobs[:, j] = log_pmf_sum + np.log(pi_[j])
+            Ez = np.argmax(Zprobs, axis=1)
+
+            # M step, maximize parameters
+            for j in range(K):
+                pi_[j] = np.mean(Ez == j)
+                Lambda_[j] = np.mean(X[Ez == j], axis=0)
+
+            eps = 0.001
+            pi_ = pi_ * (1 - K * eps) + np.ones(K) * eps
```
poneill, line 115: it looks like this is a kind of "drag term" to regularize `pi_` and prevent it from ever assigning zero probability mass to any component?  might be worth a comment to that effect if so

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+            The number of clusters, i.e. number of different types of 
+            genes
+
+        training_options : dict,default={}
+            The parameters for the inference scheme
+        """
+        self.K = int(K)
+        self.training_options = self._fill_training_options(training_options)
+
+    def fit(self, X, progress_bar=True, Lamb_init=None, pi_init=None):
```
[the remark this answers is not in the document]
poneill: what is it in the paper, though?  It looks like Lamb is matrix-valued, in which case I'd say it's actually conventional to capitalize it tbh

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+                torch.tensor(
+                    1.2 * model.means_[i].astype(np.float32) + 0.1,
+                    requires_grad=True,
+                )
+            )
+
+        myrange = trange if progress_bar else range
+
+        pct_0 = np.mean(X == 0, axis=0)
+        pct_0 = 0.98 * pct_0 + 0.01
```
poneill, line 229: if the intent here is to round `pct_0` away from 0 or 1 so it doesn't blow up the logit function, could we make that more explicit by writing `pct_0 = np.clip(pct_0, 0.01, 0.99)` (or whatever)?  (Asking because it looks syntactically similar to other transformations of params that have statistical interpretations)

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+        pct_0_inv = np.log(pct_0 / (1 - pct_0))
+
+        alpha_list_l = [
```
poneill, line 232: what does `l` stand for?  nit: also, it's conventional just to pluralize the variable name to denote a list

python/python/bystro/tada_mixture/mixture_tada_em.py
poneill: also where does 0.8 come from?

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+
+            # M step, maximize parameters
+            for k in range(K):
+                pi_[k] = np.mean(Ez == k)
+                data_sub = data[Ez == k]  # Selects relevant data
+
+                trainable_variables = [Lambda_list_l[k], alpha_list_l[k]]
+
+                optimizer = torch.optim.SGD(
+                    trainable_variables, lr=td["learning_rate"], momentum=0.8
```
poneill, line 280: do we want to pull momentum out as a parameter?

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+                )
+
+                for j in range(td["n_inner_iterations"]):
+                    Lambda_k = Lambda_list_l[k]
+                    alpha_k = 0.8 * sigmoid(alpha_list_l[k])  # Prob of being 0
```
poneill, line 285: same question about constants

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+                    optimizer.zero_grad()
+                    loss.backward()
+                    optimizer.step()
+
+                    Lambda_k.requires_grad_(False)
+                    Lambda_k[Lambda_k < 0] = 0
+                    Lambda_k.requires_grad_(True)
+
+        self.pi = pi_
+        self.Lambda = -1000 * np.ones((K, p))
```
poneill, line 317: why -1000?

python/python/bystro/tada_mixture/test/test_em.py
```diff
@@ -0,0 +1,33 @@
...
+    Lambda[0, int(p / 2) :] = 0
+    Lambda[1, : int(p / 2)] = 0
+    for k in range(K):
+        Lambda[k] = Lambda[k] / np.mean(Lambda[k]) * 10
+
+    X = np.zeros((N, p))
+    for i in range(N):
+        idx = int(Z_true[i])
+        X[i] = rng.poisson(Lambda[idx])
+        zerod_out = rng.binomial(1, Alphas[idx])
```
poneill, line 26: zeroed_out?

python/python/bystro/tada_mixture/test/test_em.py
```diff
@@ -0,0 +1,33 @@
...
+        X[i, zerod_out == 1] = 0
+
+    model = MVTadaZPoissonEM(K=K)
+    model.fit(X)
+
+    model = MVTadaPoissonEM(K=K)
```
poneill, line 32: why do we call these lines twice?

python/python/bystro/tada_mixture/test/test_em.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+from ..mixture_tada_em import MVTadaPoissonEM, MVTadaZPoissonEM
+
+
+def test_fit():
```
poneill, line 5: I'm not sure I'm fully getting the idea behind this test.  Is there anything we want to check after fitting?  If not, would it make more sense to just fit the simplest matrix `X` that could possibly work?  It seems like we're setting X up with a lot of care that seems surprising relative to what we're testing [?]

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+
+        alpha_list_l = [
+            torch.tensor(pct_0_inv, requires_grad=True) for i in range(K)
+        ]
+
+        sigmoid = nn.Sigmoid()
+
+        data = torch.tensor(X.astype(np.float32))
+        pi_ = 1 / K * np.ones(K)
+        myloss = PoissonNLLLoss(log_input=False, full=True, reduction="none")
```
poneill, line 240: nit: maybe clearer if just named `poisson_nll_loss` or similar?

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
+"""
+This uses the expectation maximization algorithm to fit a latent variable
+model for Tada, with two flavors, a ZIP and a standard poisson. The
```
poneill, line 3: somewhere (maybe not here but somewhere) we should have a link to the papers we're implementing / extending

poneill (changes requested):
Very nice, I appreciate the google-style docstrings :)

Left some comments, mostly minor, lmk what you think.  The one change I'm requesting is a one liner to avoid initializing with`training_options={}`, see comment below.

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -0,0 +1,394 @@
...
+model for Tada, with two flavors, a ZIP and a standard poisson. The 
+generative formulation is 
+
+p(z) = Dir(alpha_k)
+p(x_ij|z=k) = (ZI)Poisson(Lambda_{jk})
+
+Obviously the ZIP will have a parameter pi for probability of 0.
+
+ZIP does not have an analytic maximum likelihood formula. This is not a 
+problem for EM,  the M step only requires that we be able to maximize
```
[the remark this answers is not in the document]
austinTalbot7241993: It applies here as well as we need to maximizize the likelihood of each ZIP conditional on the cluster assignments.

python/python/bystro/tada_mixture/mixture_tada_em.py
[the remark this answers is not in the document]
austinTalbot7241993: Yeah I personally capitalize matices with vectors lower case which is what I did here.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: The reason I initialized it like this was to counteract the zero inflation portion, which is initialized to be non-zero, rough heuristic but there's also substantially less theory in initializing simple models.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: Ah need to remove the 0.8, was a feasibility constraint to make sure no component has a zero probability inflation of more than 80%. As far as the reason for sigmoid, an easy way to do inference with gradient descent when you have variables with range not R is to make a latent variable with range R and map it to a feasible region using a transformation. In this case needs to be in the range [0,.8] which is what I have achieved.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: Yeah I'll change that. I have no clue why that happens, that's terrible.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: Yes, will do.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: OK yeah I'll do that.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: L - means latent unconstrained.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: Yeah been lazy until now.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: Make it obvious if there's a problem. Not feasible value.

python/python/bystro/tada_mixture/test/test_em.py
austinTalbot7241993: Testing the ZIP component.

python/python/bystro/tada_mixture/test/test_em.py
austinTalbot7241993: I clearly need to make model names more distinctive.

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: That name comes from pytorch.

PR #237, opened 2023-09-07
Remove apt update from install-apt.sh

akotlar:
- Remove apt update from install-apt.sh, and instead indicate in installation manual that packages should be updated at the discretion of the user.


akotlar:
Will close in favor of a single commit PR from my repo

PR #239, opened 2023-09-07
Improve INSTALL.md; remove apt update from install-apt.sh

akotlar:
* Remove apt update from install-apt.sh, and instead indicate in installation manual that packages should be updated at the discretion of the user.
* Cleaned up installation instructions

This was motivated by attempting to set up Bystro on a new Ubuntu server. I realized that users will not want Bystro to force package updates, and that the installation of the Python libraries is not spelled out in INSTALL.md

Future PRs will strip out sudo from the install scripts, this is a gentle first step.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,10 +1,8 @@
-# Bystro installation
-
 For most users, we recommend not installing the software, and using https://bystro.io, where the software is hosted
 
-The web app provides full functionality for any size experiment (up to 890GB uncompressed/129GB compressed tested), a convenient search interface, and excellent performance
+The web app provides full functionality for any size experiment, a convenient search interface, and excellent performance
 
-## Docker
+# Installing Bystro using Docker
 
 ###### The recommended way to use Bystro on the command line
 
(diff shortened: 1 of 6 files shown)


install/install-apt-deps.sh
```diff
@@ -31,4 +30,5 @@ sudo apt install -y nodejs;
 sudo apt install -y npm;
 sudo npm install -g pm2;
 
-sudo apt install -y awscli;
\ No newline at end of file
+sudo apt install -y awscli;
+sudo apt install -y pkg-config;
```
akotlar, line 34: pkg-config is required for `make build`

INSTALL.md
```diff
@@ -19,29 +17,54 @@ docker run bystro bystro-annotate.pl #Annotate
...
+ 1. The Bystro annotator: a Perl program accessed through the command line (via bin/bystro-*)
+ 2. The Bystro Python package: where the rest of Bystro's functionality lives (statistics, proteomics, etc).
 
-1.  `git clone https://github.com/akotlar/bystro.git && cd bystro && source ./install-rpm.sh`
+## Installing the Bystro annotator (Perl/cli)
+
+##### (Fedora, Redhat, Centos, openSUSE, Mandriva)
 
-## Installation on MacOS
+1.  `git clone https://github.com/akotlar/bystro.git && cd bystro && source ./install-rpm.sh`
```
poneill, line 30: here and throughout, do we want to clone from bystrogenomics rather than from your personal branch?

INSTALL.md
```diff
@@ -19,29 +17,54 @@ docker run bystro bystro-annotate.pl #Annotate
...
-###### (tested on HighSierra, interactive)
+##### MacOS (tested on HighSierra, interactive)
 
 1.  `git clone https://github.com/akotlar/bystro.git && cd bystro && source ./install-mac.sh`
 
-## Installation on Debian systems
+##### Ubuntu
+1.  Ensure that packages are up to date (`sudo apt update`), or that you are satisified with the state of package versions.
+2.  `git clone https://github.com/akotlar/bystro.git && cd bystro && source ./install-apt.sh`
+    - Please note that this installation script requires root priveleges, in order to install system dependencies
```
poneill, line 39: Maybe more direct to say something like:

`Please not that this installation script will ask you for the root password in order to install system dependencies`

if that's what will happen.

install/install-apt-deps.sh
poneill: could we just put review comments like these in directly as code comments?

poneill (dismissed):
I think the documentation is heading in a good direction here.

My one blocking concern is whether we need to update the git repos, as it seems like we could easily fall out of synch otherwise-- could we change them or explicitly document that we mean to pull from the personal repo?

INSTALL.md
akotlar: Oops good catch

akotlar:
A new build step was created to test setting up our actual dev environment and building the wheel. This is slow, but there's not much we can do about that, besides restricting what files we trigger this on. I did make an attempt at that, but did not yet have success, and it's not clear we want that (because we could easily have dependencies between perl/python, for instance). In any case, the overhead is small, 8 minutes, and is run in parallel with other tests.

poneill (approved):
lgtm!

PR #242, opened 2023-09-07
Add proteomics upload CLI

### LINKED ISSUE #238: Ingest combined protein abundance file
The batch protein submission API function will follow the existing [create_job function](https://github.com/bystrogenomics/bystro/blob/master/python/python/bystro/api/cli.py#L367)

Tasks:
   - [ ]  Taking the **create_job** function as an example, create an API (probably under proteomics, e.g. proteomics/api.py rather than /api?) with a single function that 1) authenticates using our api.py login, issues a POST request having 1+ *files* and some *payload* to /api/jobs/<id>/protein.
     - The response will be minimally a status (200, 400, 500) once files received and stored
     - Mock the response for testing purposes
  - [ ] Communicate the request shape to @akotlar for consumption in webapp

poneill:
Implement CLI for uploading Fragpipe TMT proteomics datasets to `/api/jobs/upload_protein/` (endpoint TK).

Closes #238

### DIFF
--- a/python/python/bystro/proteomics/proteomics_cli.py
+++ b/python/python/bystro/proteomics/proteomics_cli.py
@@ -1,35 +1,38 @@
-from typing import Tuple, BinaryIO
+"""Provide a CLI for proteomics analysis."""
 import argparse
-import os
-import requests
+from pathlib import Path
+from typing import Any, BinaryIO
 
+import requests
+from bystro.api.cli import DEFAULT_DIR, authenticate, login
 from msgspec import json as mjson
 
-from bystro.api.cli import DEFAULT_DIR, authenticate, login
+# ruff: noqa: T201
 
 UPLOAD_PROTEIN_ENDPOINT = "/api/jobs/upload_protein/"
+HTTP_STATUS_OK = 200
 
 
-def _package_filename(filename: str) -> Tuple[str, Tuple[str, BinaryIO, str]]:
+def _package_filename(filename: str) -> tuple[str, tuple[str, BinaryIO, str]]:
     """Wrap filename in a container suitable for upload through the requests library."""
+    filepath = Path(filename)
     return (
         "file",
         (
-            os.path.basename(filename),
-            open(filename, "rb"),  # noqa: SIM115
+            filepath.name,
+            filepath.open("rb"),
             "application/octet-stream",
         ),
     )
 
 
-def upload_proteomics_dataset(args: argparse.Namespace, print_result=True) -> dict:
-    """Upload a proteomics dataset (consisting of a protein abundance
-    file and an experiment annnotation file) through the /api/jobs/upload_protein endpoint.
+def upload_proteomics_dataset(args: argparse.Namespace, *, print_result: bool = True) -> dict[str, Any]:
+    """Upload a fragpipe-TMT dataset through the /api/jobs/upload_protein endpoint.
 
     Parameters
     ----------
     args : argparse.Namespace
-        The arguments passed to the command.
+        The arguments passed to the command
     print_result : bool, optional
         Whether to print the result of the upload operation, by default True.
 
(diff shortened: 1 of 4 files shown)


python/python/bystro/proteomics/proteomics_cli.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+    }
+
+    files_to_upload = [args.protein_abundance_file, args.experiment_annotation_file]
+    files = [_package_filename(filename) for filename in files_to_upload]
+
+    if print_result:
+        print(f"\nCreating jobs for files: {', '.join(files_to_upload)}\n")
+
+    response = requests.post(url, headers=auth_header, data=payload, files=files, timeout=30)
```
akotlar, line 64: I think we'll want a longer timeout here, assuming these files can be larger than say 30MB (1MB/s typical broadband upload speed)

python/python/bystro/proteomics/proteomics_cli.py
```diff
@@ -0,0 +1,158 @@
...
+from pathlib import Path
+from typing import Any, BinaryIO
+
+import requests
+from bystro.api.cli import DEFAULT_DIR, authenticate, login
+from msgspec import json as mjson
+
+# ruff: noqa: T201
+
+UPLOAD_PROTEIN_ENDPOINT = "/api/jobs/upload_protein/"
```
akotlar, line 12: Right now we organize connect all resources to a single "job" and job_id. A job is a collection of resources, minimally an annotation (+ ancestry, search index, etc).
  - behind the scenes, each new resource is a sub-document on the main annotation document, see https://mongoosejs.com/docs/subdocs.html

I think a straightforward next step is to add a new kind of resource, protein abundance, as another subdocument in job.

We could alternatively de-couple the two.

Which would you prefer, and why?

python/python/bystro/proteomics/proteomics_cli.py
poneill: ok, assuming GB scale data and some grace for <1MB/s connections, changing this to one hour-- sound reasonable?

python/python/bystro/proteomics/proteomics_cli.py
akotlar: >  I had assumed that jobs were more like descriptions of computations to be performed, but if they're just collections of resources that clears up some questions for me.

Our conceptions are equivalent if you consider a promise a resource

akotlar: > We allow creating a new job with only proteomics data (no annotation). This will require some updating to the frontend, to ensure that the user can add an annotation to an existing job.

That makes sense to me. Long term if we want to be a general life sciences inference platform, we probably need to be able to support resource separately.

To keep things simple, in iteration one, I'll generalize the concept of job, such that a job can be created without an annotation, and that annotation is just another promise for a resource.

python/python/bystro/proteomics/proteomics_cli.py
akotlar: Yeah that sounds fine, and practically, success rates will drop beyond this.

For large documents, maybe we should ask users to pass us s3 paths + credentials? We could add that support in Sprint 2?

python/python/bystro/proteomics/proteomics_cli.py
```diff
@@ -0,0 +1,161 @@
...
+    ----------
+    args : argparse.Namespace
+        The arguments passed to the command
+    print_result : bool, optional
+        Whether to print the result of the upload operation, by default True.
+
+    Returns
+    -------
+    dict
+        A json response from the endpoint.
```
akotlar, line 43: In a future PR, to follow our desire for validation & typed results, could we convert this to a Struct (and deserialize with that as the type)?

akotlar (approved):

python/python/bystro/proteomics/proteomics_cli.py
poneill: Sorry, missed this earlier but yes.

PR #264, opened 2023-09-11
Make sure ray starts with serve-dev

akotlar:
* Ray should be started with `make serve-dev`; without this `make serve dev` will succeed, but servers relying on Ray will fail to start


poneill:
> * Ray should be started with `make serve-dev`; without this `make serve dev` will succeed, but servers relying on Ray will fail to start

I don't have a lot of context here, but would probably just fold this comment into the Makefile as a code comment.

poneill (approved):

akotlar:
> > * Ray should be started with `make serve-dev`; without this `make serve dev` will succeed, but servers relying on Ray will fail to start
> 
> I don't have a lot of context here, but would probably just fold this comment into the Makefile as a code comment.

Done

PR #265, opened 2023-09-11
[ancestry] handle missing data in vcf parsing

poneill:
Previously we'd assumed that any VCF parsed by the ancestry model would be free of missing data.  Update the VCF parsing code to handle missing data, returning `np.nan` wherever at least one allele is missing from the genotyping data, and add a test to check that we can do this successfully.

### DIFF
--- a/python/python/bystro/ancestry/train.py
+++ b/python/python/bystro/ancestry/train.py
@@ -166,7 +166,7 @@ def _parse_vcf_line_for_dosages(
     if variant in variants_to_keep:
         fields = line.split()  # now we can parse the full line
         variant_dosages = [
-            _parse_genotype_field(g) for g in fields[NUM_VCF_METADATA_COLUMNS:]
+            _parse_genotype_field(field) for field in fields[NUM_VCF_METADATA_COLUMNS:]
         ]  #  genotype fields take the form e.g. '0|1', '0/1' or './.'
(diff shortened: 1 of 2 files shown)


python/python/bystro/ancestry/train.py
```diff
@@ -166,12 +166,20 @@ def _parse_vcf_line_for_dosages(
     if variant in variants_to_keep:
         fields = line.split()  # now we can parse the full line
         variant_dosages = [
-            int(psa[0]) + int(psa[2]) for psa in fields[NUM_VCF_METADATA_COLUMNS:]
-        ]  #  pipe-separated annotation e.g. '0|1'
+            _parse_genotype_field(field) for field in fields[NUM_VCF_METADATA_COLUMNS:]
+        ]  #  genotype fields take the form e.g. '0|1', '0/1' or './.'
```
cristinaetrv, line 170: I think one dot like this '.' is also an option for missing genotypes, do we want to be able to handle that format?

cristinaetrv (approved):
Lgtm!

python/python/bystro/ancestry/train.py
poneill: Yes, we do handle that format and test for it in the tests.  But will update the docstring to make that clearer, ty.

PR #266, opened 2023-09-11
Add noqa's to test_ancestry_types.py in order to explicitly check types in ProbabilityInterval

poneill:
In test_ancestry_types.py there are a few tests where we must explicitly assert that the type of a certain value is `float` in order to ensure msgspec compatibility, rather than using the more conventional `isinstance` check.  Somehow this made it past the linter, which is now objecting to them.  

Add noqa's to test_ancestry_types.py so we can test explicitly for floats while keeping the linter passing on master.

### DIFF
--- a/python/python/bystro/ancestry/tests/test_ancestry_types.py
+++ b/python/python/bystro/ancestry/tests/test_ancestry_types.py
@@ -14,6 +14,16 @@ from bystro.ancestry.ancestry_types import (
 from bystro.ancestry.train import POPS, SUPERPOPS
 
 
+# ruff: noqa: E721
+
+# In several tests we explicitly check that a value `is float` rather
+# than use the more pythonic `isinstance(value, float)`.  The former
+# method of checking raises E71 errors, which are exempted from the
+# linter on a file-wide basis above.  We make these explicit checks in
+# order to ensure that such values aren't np.float64's, which
+# can't easily be deserialized.
+
+
 def test_AncestrySubmission_accepts_valid_vcf_paths():
     AncestrySubmission("foo.vcf")
     AncestrySubmission("foo.vcf.gz")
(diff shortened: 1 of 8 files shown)


python/python/bystro/ancestry/tests/test_ancestry_types.py
```diff
@@ -47,8 +47,9 @@ def test_AncestrySubmission_is_frozen():
 def test_ProbabilityInterval_accepts_valid_bounds() -> None:
     """Ensure we can instantiate, validate ProbabilityInterval correctly."""
     prob_int = ProbabilityInterval(lower_bound=0.1, upper_bound=0.9)
-    assert type(prob_int.lower_bound) is float  # for msgspec serialization
-    assert type(prob_int.upper_bound) is float
+    # for msgspec serialization we specifically have to check if raw float
```
akotlar, line 50: Could you expand on this comment? This comment tells me something important, but without enough context for me to understand if this is a msgspec limitation, or how it applies.

Also, it looks like ProbabilityInterval is an attrs, so not 100% sure this comment still applies.

poneill: Yeah, this dependency in the codebase could probably be made clearer.  ProbabilityInterval used to take np.float64s, but these were changed to raw floats in https://github.com/bystrogenomics/bystro/pull/197 "to make AncestryResponse msgspec compatible".  So at some point msgspec must have been trying to consume serialized AncestryResponses and choking on the numpy floats, and the decision was made to convert them to raw floats on the backend.

Unfortunately I don't have any further record of this decision in GH or Teams.  Do you happen to recall if/where msgspec comes into contact with AncestryResponses (probably in serialized form)?  I'd like to update the comment to make all of this clearer.

akotlar: Yeah, in order to decode numpy float64 objects, we needed to specify a custom decoder; we elected to downcast them to native floats instead.

akotlar: Mostly because msgspec has no way of knowing what the byte layout of numpy is; it's not a native json type

akotlar: More info:
* https://github.com/jcrist/msgspec/issues/26
* https://github.com/jcrist/msgspec/pull/31
* https://jcristharif.com/msgspec/extending.html

akotlar (approved):

PR #267, opened 2023-09-12
Feature/multivariate tada cleanup

austinTalbot7241993:
This aligns multivariate tada Poisson models with the format we coalesced on for supervised PPCA and statistical models using SGD with pytorch.


poneill (changes requested):

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -374,6 +427,101 @@ def predict_logproba(self, data):
...
+        Returns
+        -------
+        """
+        model = GaussianMixture(self.K)
+        model.fit(X)
+        Lambda_list_l = []
+        for i in range(self.K):
+            Lambda_list_l.append(
+                torch.tensor(
+                    1.2 * model.means_[i].astype(np.float32) + 0.1,
```
poneill, line 467: here and throughout, I would recommend pulling out and naming magic numbers like these

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -374,6 +427,101 @@ def predict_logproba(self, data):
...
+        pct_0 = np.clip(pct_0, 0.01, 0.99)
+        pct_0_inv = np.log(pct_0 / (1 - pct_0))
+
+        alpha_ls = [
+            torch.tensor(pct_0_inv, requires_grad=True) for i in range(self.K)
+        ]
+        pi_ = 1 / self.K * np.ones(self.K)
+        return Lambda_list_l, alpha_ls, pi_
+
+    def _initialize_save_losses(self):
```
poneill, line 482: it might be difficult for someone walking into this code for the first time to tell what's intended here, because we have a private method whose only content is its docstring, which apparently disagrees with the method itself.  So a reader might wonder: were these methods supposed to be implemented, or are they meant to be this way?  Are they load-bearing at all?  Could I just delete them?

If we just forgot to implement them, let's implement them and put them under test to avoid future confusion.  If they're supposed to be this way, we should probably delete them.  If we can't delete them (maybe because they're required by an ABC further up the class hierarchy?), we should probably reevaluate the class design.

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -80,44 +83,114 @@ def fit(self, X, progress_bar=True, Lamb_init=None, pi_init=None):
...
-            eps = 0.001
-            pi_ = pi_ * (1 - K * eps) + np.ones(K) * eps
+        Parameters
+        ----------
+        """
 
-        self.Lambda = Lambda_
-        self.pi = pi_
-        return self
+    def _save_losses(self,i):
```
poneill, line 167: could you help me understand what's happening here?  It looks like this method isn't implemented, and we're not using the argument `i`, which ruff and / or mypy should be warning about

python/python/bystro/tada_mixture/mixture_tada_em.py
```diff
@@ -374,6 +427,101 @@ def predict_logproba(self, data):
...
+        This method initializes the arrays to track relevant variables
+        during training
+
+        TODO
+
+        Parameters
+        ----------
+        """
+
+    def _save_losses(self):
```
poneill, line 493: It seems like the name of this method (`_save_losses`) just disagrees with implementation ("do nothing and return"), which makes this potentially dangerous code.

poneill (changes requested):

poneill (changes requested):
It looks like we have a lot of methods in these classes that have empty implementations (i.e. they do nothing and return).  If the implementations were omitted by accident, let's implement them and put them under test.  If they're omitted intentionally there's probably a class design issue we should revisit-- I'm not sure off-hand but it might be the case that there's an ABC somewhere above requiring us to implement certain methods that it really shouldn't require, and the effects of that are showing up here?

akotlar (approved):

akotlar:
Tests work, wheel building is tested by CI, but `make develop` works, as does library importing and running the methods.

Only note is we should fix:

```sh
/mnt/ssd1/bystro/python/python/bystro/contrib/tada_mixture/mixture_tada_marginal.py:118: UserWarning: Implicit dimension choice for softmax has been deprecated. Change the call to include dim=X as an argument.

/mnt/ssd1/bystro/python/python/bystro/contrib/tada_mixture/mixture_tada_marginal.py:278: UserWarning: Implicit dimension choice for softmax has been deprecated. Change the call to include dim=X as an argument.
  pi_ = smax(trainable_variables[1]) * mu + (1 - mu) / self.K
```

akotlar:
Also, in future PRs, I would like to see private, unimplemented methods of BaseSGDModel, and the implemented ones to be made public.

akotlar:
2 more things:
1. `class BaseCovariance(object)` should just be `class BaseCovariance:`
2. We should move all of the various ML things to either bystro.stats, bystro.ml, or similar. Right now we have bystro.covariance, bystro.precision. Prefer bystro.ml.covariance, bystro.ml.precision. Then we have "production" ml libraries and "contrib"/more in flight libraries

python/python/bystro/tada_mixture/mixture_tada_em.py
austinTalbot7241993: I was hoping to save out the ELBO but computing that is not trivial. I was planning on adding it in later, or at least some kind of metric on convergence progress.

austinTalbot7241993:
We're moving this to contrib where the standards are lower. Please accept as is.

PR #268, opened 2023-09-14
optimize unit test suite

poneill:
Optimize unit test suite performance by setting `n_iterations` training parameter to lower values that maintain test invariants.

This results in a 6x speedup, bringing local unit testing wall time from ~70s to ~12s.


akotlar (approved):
Looks good to me.

PR #269, opened 2023-09-16
Restructuring Perl package and docker recipe

wingolab:
- Added code tidying for perl, shell, python, and markdown that use pre-commit
- Moved perl package code into perl directory
- Added dist.init (Dist::Zilla) for building perl package
- Updated Dockerfile recipe for building go and perl portions of bystro
- Added Makefile option for building docker container


perl/README.md
```diff
@@ -0,0 +1,26 @@
+# Bystro Annotator
+
+## Install
+
+```bash
+# system install
+cpanm Dist::Zilla
```
akotlar, line 7: Need 

`dzil authordeps | cpanm`

before `dzil install` or install will fail. If I just read the instructions, I'm not sure what to do with Author dependencies, and comes after the installation instructions, so I wouldn't think to run it first.

We also want `cpanm Archive::Tar::Wrapper`

to get rid of the warning `[DZ] building archive with Archive::Tar; install Archive::Tar::Wrapper 0.15 or newer for improved speed` and improve build speed

perl/README.md
```diff
@@ -0,0 +1,26 @@
...
+dzil install
```
akotlar, line 8: So I can't get this to quite work. Also tried cpanm the built tarball, same error:

```
2023-09-16 22:50:38 [info] ref: Beginning building from /home/alexkotlar/bystro2/perl/.build/u4e7cswOD8/t/tracks/gene/db/raw/ref/chrM.fa.gz
pigz: skipping: /home/alexkotlar/bystro2/perl/.build/u4e7cswOD8/t/tracks/gene/db/raw/ref/chrM.fa.gz does not exist
```

perl/t/tracks/gene/db/raw/ref/chrM.fa.gz definitely exists, but seems like it isn't getting packaged?

startup.yml
```diff
@@ -1,25 +1,21 @@
 apps:
   - name: BystroAnnotationServer
     script: bin/bystro-server.pl
```
akotlar, line 3: this should now be perl/bin/bystro-server.pl, or maybe if the installation process makes bystro-server available in PATH, that.

akotlar (changes requested):
Looks really good, thank you!

A few comments to address and then this is good to go. Besides what is commented on specific code, we also need to update mentions of bin/bystro-* in /README.md to read perl/bin/bystro-*

startup.yml
wingolab: Installing the perl package will place those scripts into the path.

wingolab: It's not clear what the startup or startup-dev files are used for since they are not mentioned in the repository as far as I could tell.

perl/README.md
wingolab: You're right. Binary files are excluded because this caused the packaging to break. I can update the README with the approach used for building the docker container, which works for installing it locally, too. I suggest making this a goal for automated testing. Right now there are two perl packages that need to be installed outside of the normal cpanm approach.

startup.yml
akotlar: Yeah, that should be clarified. startup-dev.yml is no longer used, but startup.yml is used to start the servers, use ‘make serve-dev’

perl/README.md
akotlar: Would making it just a .fa fix that?

akotlar: And yes, would love to get the updated README so that users / I can understand how to install this.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - id: end-of-file-fixer
+      - id: check-added-large-files
+  - repo: https://github.com/scop/pre-commit-shfmt
+    rev: v3.7.0-1
+    hooks:
+      - id: shfmt # (requires/installs Go to build)
+  - repo: https://github.com/psf/black
+    rev: 22.10.0
+    hooks:
+      - id: black
```
poneill, line 15: So I appreciate the spirit behind this PR and am fine with moving in this direction.  I already run black automatically on save so it doesn't make much difference to me either way.  However:

1.  We'd need to communicate the decision to use pre-commit because IIRC each user will still need to do some manual config to set it up locally.
2.  I'm not sure how this works on CI, so if a user skips that step they might end up contributing non-conformant code, and then we're back to whitespace whack-a-mole :)
3.  Doing a global reformatting will lose us a lot of `git blame` context, because this PR becomes the author of essentially the entire repo.  If we could configure this hook to work incrementally, only on changed files, we might have a better time of preserving history so people can tell who last touched a piece of code and when.
4. Another approach that might get us to the same desired end-state is to make black part of CI, that way a user can't get their pre-commit hooks misconfigured and gum up the repo.
5. We already have black rules stored in pyproject.toml (line-length, specifically).  I think the pre-commit hook automatically picks those up, but it's hard to tell here and would be worth having a comment that that was manually tested and works correctly.  If it doesn't and we have two sources of truth about black rules, it's going to be hard to sort out.

poneill:
It looks like git is getting pretty badly confused about some files in`contrib/tada_mixture`, mixing up methods from two different classes.  In light of the other concerns I'd probably recommend leaving the python code aside for now and scoping this PR to the perl code (as the PR title currently suggests).  

If we want to go ahead the easiest fix is probably something like:

- make sure the local repo is up to date`git checkout master && git pull upstream master`
- revert the python directory to master (`git checkout master path/to/bystro/python`)
- rebase on master without any formatting changes in the python directory (`git rebase master`)
- apply formatting changes to the python directory afterwards

perl/README.md
wingolab: README.md was updated with latest approach to building.

startup.yml
wingolab: I'd vote to make fixing these yaml configuration files separate issues related to integrating the Perl with JS and other components.

wingolab:
The suggestion to split the Perl code restructuring and automated building from tidying other portions or my pre-commit suggestion is reasonable. @akotlar please take the reigns to cherry-pick the pieces that would achieve that goal.

akotlar:
@wingolab I resolved the lint.yml conflict (this was due to my last PR)

akotlar:
Thomas and I talked, and he will finish this PR up

wingolab:
This work was superseded by #286.

PR #272, opened 2023-09-26
[WIP]

poneill:


python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+from bystro.search.utils.annotation import AnnotationOutputs, get_delimiters
+from bystro.search.utils.messages import SaveJobData
+from bystro.search.utils.opensearch import gather_opensearch_args
+from opensearchpy import OpenSearch
+from ruamel.yaml import YAML
+
+ray.init(ignore_reinit_error=True, address="auto")
+RAY_ERROR = -1  # the error code we use if a ray remote job fails to return
+
+ONE_DAY = "1d"  # default keep_alive time for opensearch point in time index
```
akotlar, line 22: We should just be extra careful to catch exceptions and clean up pit indices regardless of whether query succeeds

python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+        discordant_idx = field_names.index("discordant")
+    except ValueError as val_err:
+        err_msg = f"response: {resp} with field_names: {field_names} lacked field: 'discordant'"
+        raise ValueError(err_msg) from val_err
+    rows: list[Row] = []
+    docs = resp["hits"]["hits"]
+    for doc in docs:
+        row = []
+        for parent_field, child_field in zip(parent_fields, child_fields):
+            row.append(_populate_data(child_field, doc["_source"].get(parent_field)))
```
akotlar, line 136: Could you help me understand the _populate_data need?

What are the drawbacks of just doing:

```python
refSeq = doc.get('refSeq')

if not refSeq: 
   continue

gene = refSeq.get('name2')

if not gene:
  log.warn("Gene not found for refSeq record: {json.dumps(refSeq, indent=4})
```

python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+        row[discordant_idx][0][0] = cast(str, _fix_discordancy(row[discordant_idx][0][0]))  # type: ignore
+        rows.append(row)  # type: ignore
+    return rows
+
+
+def _fix_discordancy(x: bool | str) -> str:
+    return {False: "0", True: "1", "0": "0", "1": "1"}[x]
```
akotlar, line 143: Why are we trying to fix discordant fields?

akotlar: These are just rows that the user submitted with ref != assembly ref. If there are a lot of these they used the wrong genome, else they should be careful with the data. We should never fix these guys

python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+            slice_query["slice"] = {"id": slice_id, "max": num_slices}
+        remote_query = _process_query_pure.remote(  # type: ignore[call-arg]
+            query_args={"body": slice_query},
+            search_client_args=search_client_args,
+            field_names=job_data.fieldNames,
+            delimiters=delimiters,
+        )
+        remote_queries.append(remote_query)
+    results_processed = ray.get(remote_queries)
+    client.delete_point_in_time(body={"pit_id": pit_id})  # type: ignore
```
akotlar, line 262: We should wrap everything from line 245 to 261 in a try except, to ensure that we can clean up the point in time index, which will rapidly swell opensearch storage, potentially (on writes to the index only I believe, but there are also limits on the number of pit's that are allowed)

python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+    samples_and_genes_df = samples_and_genes_df.rename({"refSeq.name2": "gene_name"}, axis=1)
+    output = []
+    for _i, row in samples_and_genes_df.iterrows():
+        gene_names = deduplicate(split_gene_names(row.gene_name, delimiters))
+        homozygotes = row.homozygotes.split("|") if row.homozygotes != "!" else []
+        heterozygotes = row.heterozygotes.split("|") if row.heterozygotes != "!" else []
+        for gene_name in gene_names:
+            for heterozygote in heterozygotes:
+                output.append((heterozygote, gene_name, 1))
+            for homozygote in homozygotes:
```
akotlar, line 290: We probably also want to store missingGenos. Missing values are themselves interesting. Without knowing what is missing you can't know if the individual is reference. Researchers may care about folks who had ambiguous (missing) calls because they're non-reference

python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+    # Each sliced scroll chunk should get all records for that chunk
+    if len(resp["hits"]["hits"]) != resp["hits"]["total"]["value"]:
+        err_msg = (
+            f"response hits: {len(resp['hits']['hits'])} didn't equal "
+            f"total value: {resp['hits']['total']['value']}. "
+            "This is a bug."
+        )
+        raise ValueError(err_msg)
+    rows = _process_rows(resp, field_names)
+    output_string = _make_output_string(rows, delimiters)
```
akotlar, line 196: I wouldn't do this. Just return the object. If you're going to be parsing it anyway, no need to resplit later. Ray will happily pickle values.

python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+    return _process_payload(payload, delimiters)
+
+
+def _process_payload(payload: list[bytes], delimiters: dict[str, str]) -> pd.DataFrame:
+    payload_str = "".join([byte_str.decode("utf-8") for byte_str in payload])
+    samples_and_genes_df = pd.read_csv(io.StringIO(payload_str), sep="\t")
+    assert all(
+        samples_and_genes_df.columns == ["discordant", "homozygotes", "heterozygotes", "refSeq.name2"]
+    )
+    samples_and_genes_df = samples_and_genes_df.drop("discordant", axis=1)
```
akotlar, line 280: why do you want to return discordant just to drop it?

python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+    samples_and_genes_df = samples_and_genes_df.rename({"refSeq.name2": "gene_name"}, axis=1)
+    output = []
+    for _i, row in samples_and_genes_df.iterrows():
```
akotlar, line 283: I would do all of this deduplication on the workers

python/python/bystro/proteomics/handler.py
```diff
@@ -0,0 +1,362 @@
...
+        gene_names = deduplicate(split_gene_names(row.gene_name, delimiters))
+        homozygotes = row.homozygotes.split("|") if row.homozygotes != "!" else []
+        heterozygotes = row.heterozygotes.split("|") if row.heterozygotes != "!" else []
+        for gene_name in gene_names:
+            for heterozygote in heterozygotes:
+                output.append((heterozygote, gene_name, 1))
+            for homozygote in homozygotes:
+                output.append((homozygote, gene_name, 2))
+    return pd.DataFrame(output, columns=["sample_id", "gene_name", "dosage"])
```
akotlar, line 292: this whole function can be lifted into the ray worker, and avoid converting back and forth as strings; I think Ray should be able to handle serializing a pandas dataframe, allowing you to just concatenate them.

akotlar (changes requested):
A few comments left.

My overall feedback would be to define the fields you expect users to want to gather info from and directly access those fields from the response (e.g. doc['refSeq']['name2']), unless  you expect those fields to be variable, then use the iterative stack algorithm as you are. 

I think we also want to entirely avoid returning the results as strings, and instead send the protein abundance data in chunks to workers, do the data frame creation on those workers, and concatenate the resulting data frames. I would probably send the protein abundance data to those workers as well and do all of the combination of protein data, genetic data there; long term if protein abundance data is large, we'll store that data in separate files anyway, making it even more natural, as we'll be able to have the workers handle the file reading as well.

python/python/bystro/proteomics/handler.py
poneill: This is just a refactoring of the four lines beginning here: https://github.com/bystrogenomics/bystro/blob/f5ef13e9cb2309429d0ed79d980b754c85844c6d/python/python/bystro/search/save/handler.py#L148

Should we not be fixing them there as well?

akotlar:
I assume this is no longe relevant?

poneill:
superceded

PR #274, opened 2023-09-26
Add utility functions to get name of GNU tar executable in OS-agnostic manner

poneill:
At various points in annotation file processing we rely on `tar` to compress or decompress files.  In particular, we use GNU tar and use certain features that are specific to that implementation.  On linux, GNU tar is the default.  On MacOSX, however, the default implementation is `bsdtar`, which is not fully compatible with GNU tar.  If we end up calling bsdtar when we mean GNU tar, subtle bugs can result.

So, in order to run `tar` correctly we need to know which OS we're running it on and specify the correct executable name for GNU tar (or raise an error if it isn't installed).  This commit defines a utils module that provides a runtime constant `GNU_TAR_EXECUTABLE_NAME` valid under linux or MacOSX.


poneill:
Closing in favor of #276 due to merge conflicts

PR #275, opened 2023-09-27
Add caching to lint.yml

akotlar:
* Updates lint.yml to cache venv

Most of our lint.yml build time is spent in dependency installation. This PR caches the virtual env, so that unless requirements change, the venv will be served from cache rather than installing from scratch


poneill (approved):
thanks, LGTM!

PR #276, opened 2023-09-27
 Add utility functions to get name of GNU tar executable in OS-agnostic manner

### LINKED ISSUE #277: linux and macosx use two incompatible tar implementations, leading to bugs on macosx
### Background:
Linux uses GNU tar whereas Macs use bsdtar by default.  Previously we've written code assuming that `tar` resolves to GNU tar.  On Macs, GNU tar must be brew-installed and then invoked with `gtar`.

###### A concise description of why we're pursuing this task.
These implementations are not totally compatible, leading to bugs when we run Python scripts that invoke `tar` through the subprocess module.

###### What is the goal, and how do we know this task is accomplished? 
The ultimate goal is to replace all literal invocations of `tar` with a more modular command that will do the right thing on either linux or macosx.  The first step towards that goal is to write a function to figure out what operating system the Python process is running on and return the appropriate command for GNU tar, so that we can gradually make our programmatic interactions with the command line OS-agnostic, increasing developer velocity.

###### Give enough technical detail for a teammate to understand the proposed solution.
See above.

poneill:
At various points in annotation file processing we rely on tar to compress or decompress files. In particular, we use GNU tar and use certain features that are specific to that implementation. On linux, GNU tar is the default. On MacOSX, however, the default implementation is bsdtar, which is not fully compatible with GNU tar. If we end up calling bsdtar when we mean GNU tar, subtle bugs can result.

So, in order to run tar correctly we need to know which OS we're running it on and specify the correct executable name for GNU tar (or raise an error if it isn't installed). This commit defines a utils module that provides a runtime constant GNU_TAR_EXECUTABLE_NAME valid under linux or MacOSX.

Closes #277

### DIFF
--- a/python/python/bystro/utils/tar.py
+++ b/python/python/bystro/utils/tar.py
@@ -33,8 +33,9 @@ def _determine_os() -> OperatingSystem:
             check=True,
         )
     except FileNotFoundError as file_not_found_error:
-        err_msg = "Couldn't run `uname -a` in shell to determine OS: "
-        err_msg += "only linux and macosx are supported."
+        err_msg = (
+            "Couldn't run `uname -a` in shell to determine OS: only linux and macosx are supported."
+        )
         raise AssertionError(err_msg) from file_not_found_error
     downcased_uname_output = result.stdout.lower()
     if "darwin" in downcased_uname_output:
--- a/python/python/bystro/utils/tar.py
+++ b/python/python/bystro/utils/tar.py
@@ -54,11 +54,10 @@ def _get_gnu_tar_executable_name() -> str:
     operating_system = _determine_os()
     if operating_system is OperatingSystem.linux:
         return GNU_TAR_LINUX
-    elif operating_system is OperatingSystem.macosx:
+    if operating_system is OperatingSystem.macosx:
         _assert_gtar_installed_on_macosx()
         return GNU_TAR_MACOSX
-    else:
-        return assert_never(operating_system)
+    return assert_never(operating_system)
 
 
 def _assert_gtar_installed_on_macosx() -> None:
--- a/python/python/bystro/utils/tar.py
+++ b/python/python/bystro/utils/tar.py
@@ -19,7 +19,11 @@ class OperatingSystem(Enum):
 def _get_executable_name(name: str) -> str:
     try:
         result = subprocess.run(
-            ["which", name], capture_output=True, text=True, check=True, shell=False  # noqa: S603, S607
+            ["/usr/bin/which", name],
+            capture_output=True,
+            text=True,
+            check=True,
+            shell=False,  # noqa: S603 (input is safe)
         )
     except CalledProcessError as err:
         err_msg = f"executable `{name}` not found on system, is it installed?"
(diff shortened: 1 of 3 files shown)


python/python/bystro/utils/tar.py
```diff
@@ -0,0 +1,77 @@
...
+
+class OperatingSystem(Enum):
+    """Represent the operating system on which the current Python process is running."""
+
+    linux = "linux"
+    macosx = "macosx"
+
+
+# here and throughout, fully qualify executable names to avoid privilege escalation
+GNU_TAR_LINUX = "/usr/bin/tar"
```
akotlar, line 16: Can we use something like `which` to locate this? I'm not sure what privilege escalation is possible with that approach, please help me understand

python/python/bystro/utils/tar.py
```diff
@@ -0,0 +1,77 @@
...
+        return OperatingSystem.linux
+    err_msg = f"Could not determine OS from `uname -a` output: `{result.stdout}`"
+    raise OSError(err_msg)
+
+
+def _get_gnu_tar_executable_name() -> str:
+    """Find the name of the GNU tar executable on user's instance."""
+    # Macs use bsdtar by default.  bsdtar is not fully compatible with
+    # GNU tar, the implementation we use.  So if we're on a mac, we'll
+    # want to use 'gtar' instead of 'tar'.
```
akotlar, line 53: This makes sense; could we update the Mac install script to install gtar? It also strikes me that we'll need to do this in Perl, which we'll need an issue for.

akotlar (changes requested):
A question and request for install script update

python/python/bystro/utils/tar.py
poneill: Thanks, good call, updated to use which to locate these.  

In re privilege escalation, if command line names aren't fully qualified, a malicious user with access to $PATH could save an executable named `tar` in a location that had precedence over /usr/bin and get up to shenanigans.  Not the biggest risk in this scenario but it's a belt and suspenders kind of thing.

poneill:
@akotlar mind taking another look?  Main changes:

- updated install-mac-deps.sh
- use `which` to locate tar/gtar

python/python/bystro/utils/tar.py
```diff
@@ -0,0 +1,76 @@
...
+    linux = "linux"
+    macosx = "macosx"
+
+
+# here and throughout, fully qualify executable names to avoid privilege escalation
+
+
+def _get_executable_name(name: str) -> str:
+    try:
+        result = subprocess.run(
```
akotlar, line 21: Could we instead use shuttle.which for portability?

python/python/bystro/utils/tar.py
```diff
@@ -0,0 +1,76 @@
...
+    """Raise error for mypy if code is proven reachable."""
+    msg = f"Unhandled value: {value} ({type(value).__name__})"
+    raise AssertionError(msg)
+
+
+def _determine_os() -> OperatingSystem:
+    """Determine whether we're running on Linux or MacOSX by inspecting uname output."""
+    try:
+        result = subprocess.run(
+            ["/usr/bin/uname", "-a"],  # noqa: S603 (input is safe)
```
akotlar, line 44: Could we use sys.platform instead? https://docs.python.org/3/library/sys.html#sys.platform

More portable

akotlar (changes requested):

python/python/bystro/utils/tar.py
poneill: good call, thanks

akotlar (approved):

akotlar:
@poneill merge at your convenience

PR #278, opened 2023-09-27
Add project-level config utilities

poneill:
Provide utility functions for loading the opensearch config and finding the bystro project root directory.

### DIFF
--- a/python/python/bystro/utils/config.py
+++ b/python/python/bystro/utils/config.py
@@ -19,9 +19,9 @@ def get_bystro_project_root() -> Path:
     # find project root by walking up the tree until we get to top level bystro directory.
     # The bystro top level is assumed to be the unique directory containing a startup.yml file.
     path = Path().absolute()
-    ROOT = Path("/")
+    FILESYSTEM_ROOT = Path("/")
     found_startup_file = False
-    while path != ROOT:
+    while path != FILESYSTEM_ROOT:
         if any(path.glob("startup.yml")):
             found_startup_file = True
             break
--- a/python/python/bystro/utils/config.py
+++ b/python/python/bystro/utils/config.py
@@ -14,7 +14,7 @@ ConfigDict = Dict[str, Union["ConfigDictValue", list["ConfigDictValue"]]]
 ConfigDictValue = Union[str, int, float, "ConfigDict"]
 
 
-def get_bystro_project_root() -> Path:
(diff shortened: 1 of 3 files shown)


python/python/bystro/utils/config.py
```diff
@@ -0,0 +1,41 @@
...
+def get_bystro_project_root() -> Path:
+    """Return Path of bystro project root."""
+    # find project root by walking up the tree until we get to top level bystro directory.
+    # The bystro top level is assumed to be the unique directory containing a startup.yml file.
+    path = Path().absolute()
+    FILESYSTEM_ROOT = Path("/")
+    found_startup_file = False
+    while path != FILESYSTEM_ROOT:
+        if any(path.glob("startup.yml")):
+            found_startup_file = True
```
akotlar, line 26: Could we memoize this so that we're not hitting the filesystem a bunch if we rely on this, lifting BYSTRO_ROOT to a config-level global?

python/python/bystro/utils/tests/test_config.py
```diff
@@ -0,0 +1,22 @@
+from unittest.mock import patch
+
+import pytest
+from bystro.utils.config import get_bystro_project_root, get_opensearch_config
+
+
+def test_get_bystro_project_root():
+    bystro_project_root = get_bystro_project_root()
+    assert "bystro" in str(bystro_project_root)
```
akotlar, line 9: I don't think this works, because "bystro" will also be in python/python/bystro; so if finding the root fails, we'll still potentially find "bystro". Could we check that os.path.exists(Path.join(bystro_project_root, 'startup.yml')?

akotlar (changes requested):
Minor comments: memoize, and check startup.yml instead of "bystro" in test.

python/python/bystro/utils/config.py
poneill: good call, thanks

python/python/bystro/utils/tests/test_config.py
poneill: I think if finding the project root fails (i.e. we fail to find startup.yml) we'll raise FileNotFoundError and so won't make it to the assertion in line 9.  But happy to change the test anyway to make the success condition more explicit.

akotlar (approved):

akotlar:
Please merge when ready. Thanks for the PR!

PR #283, opened 2023-10-04
[proteomics] Add functionality to query annotation file and return variant info

poneill:
Given an opensearch query and an index name, run the query against the index and return a dataframe of matching variants / samples.


python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -0,0 +1,195 @@
...
+        logger.exception(err_msg, exc_info=e)
+        client.delete_point_in_time(body={"pit_id": pit_id})  # type: ignore[attr-defined]
+        raise
+    client.delete_point_in_time(body={"pit_id": pit_id})  # type: ignore[attr-defined]
+    return pd.concat(remote_queries)
+
+
+def _build_opensearch_query_from_query_string(query_string: str) -> dict[str, Any]:
+    return {
+        "query": {
```
poneill, line 163: Ideally we'd be selecting pre-filtering the results to only return the necessary fields, but I keep running into errors like: 

```opensearchpy.exceptions.RequestError: RequestError(400, 'parsing_exception', 'request does not support [_source]')```

see e.g. here for additional context: https://stackoverflow.com/questions/9605292/make-elasticsearch-only-return-certain-fields

File: python/python/bystro/proteomics/default_query_fields.py [hunk 2]
```diff
@@ -0,0 +1,182 @@
+"""Default fields to pass to OpenSearch query."""
+
+DEFAULT_FIELDS = [
```
akotlar, line 3: Do we need this still?

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -0,0 +1,188 @@
...
+HOMOZYGOTE_DOSAGE = 2
+MISSING_GENO_DOSAGE = np.nan
+ONE_DAY = "1d"  # default keep_alive time for opensearch point in time index
+
+# The fields to return for each variant matched by the query
+OUTPUT_FIELDS = ["refSeq.name2", "homozygotes", "heterozygotes", "missingGenos"]
+
+
+@dataclass
+class OpenSearchQueryConfig:
```
akotlar, line 25: I would keep this as a Struct. We are using 3 different slot class libraries, let's stick to 1.

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def _get_samples_genes_dosages_from_hit(hit: dict[str, Any]) -> pd.DataFrame:
+    """Given a document hit, return a dataframe of samples, genes and dosages."""
+    source = hit["_source"]
+    gene_names = _flatten(source["refSeq"]["name2"])
+    # homozygotes, heterozygotes may not be present in response, so
+    # represent them as empty lists if not.
+    heterozygotes = _flatten(source.get("heterozygotes", []))
+    homozygotes = _flatten(source.get("homozygotes", []))
```
akotlar, line 50: I would drop flatten here; this will never be a nested list, either undefined or 1D array

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -0,0 +1,188 @@
...
+    missing_genos = _flatten(source.get("missingGenos", []))
+    rows = []
+    for gene_name in gene_names:
+        for heterozygote in heterozygotes:
```
akotlar, line 54: We should check if the gene_name has been seen before. 99% of cases for multiple name2 is alternate transcripts of a single gene.

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -0,0 +1,188 @@
...
+
+    n_docs: int = response["count"]
+    if n_docs < 1:
+        err_msg = (
+            f"Expected at least one document in `response['count']`, got response: {response} instead."
+        )
+        raise RuntimeError(err_msg)
+
+    num_slices_necessary = math.ceil(n_docs / OPENSEARCH_QUERY_CONFIG.max_query_size)
+    num_slices_planned = min(num_slices_necessary, OPENSEARCH_QUERY_CONFIG.max_slices)
```
akotlar, line 113: Why do we want `num_slices_planned`? If we have more than 10,000 * 1024 records, we want to return all of them

python/python/bystro/proteomics/default_query_fields.py
akotlar: Let's drop Default Fields. We'll need a build step that generates this automatically from the yaml configs.

akotlar (changes requested):
A few small things

python/python/bystro/proteomics/annotation_interface.py
poneill: per discussion off-line, holding off on this comment until we determine why this implementation was chosen for search/save/handler.py

akotlar: Remind myself of the API: https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#scroll-search-results

The sliced scroll is distinct from regular scroll in that there are multiple scroll contexts, which evenly divide (if possible) the total hits. Each scroll context is paginated independently, from the first document accessible to that scroll context, to the last, and pagination happens in "size" records.

So this use, and the one in the save handler, are fine.

akotlar:
Needs lint fix

python/python/bystro/utils/config.py
```diff
@@ -41,3 +42,21 @@ def get_opensearch_config() -> ConfigDict:
     with opensearch_config_filepath.open() as search_config_file:
         config_dict: ConfigDict = YAML(typ="safe").load(search_config_file)
     return config_dict
+
+
+# define the set of permissible reference genomes so that they're
+# available both statically and at runtime
+ReferenceGenome = Literal["hg19", "hg38"]
```
akotlar, line 49: This is fine as a starting point. We do support non-human genomes, and have mapping configs for those, but I am ok with focusing on humans for the initial re-release of bystro, so non-blocking, and I appreciate the mapping config checks

python/python/bystro/proteomics/tests/integrations/test_annotation_interface_integration.py
```diff
@@ -0,0 +1,59 @@
...
+        BYSTRO_PROJECT_ROOT / "python/python/bystro/proteomics/tests/integrations/trio_trim_vep_vcf.tar"
+    )
+    assert tar_path.exists()
+    asyncio.run(
+        upload_and_index_vcf(annotation_filename, mapping_config, opensearch_config, str(tar_path))
+    )
+    # after uploading, some additional time is required on the annotator's end before the results
+    # are available to query.  Pause here for a few seconds in order to avoid prematurely querying
+    # the annotation file.
+    time.sleep(3)
```
akotlar, line 41: no longer needed

akotlar (approved):
Generally good to go. 2 non-blocking comments

PR #292, opened 2023-10-17
Update Perl Docker recipe

wingolab:
- change cpanm to cpm for faster building
- minor tweaks to dist and cpanfile


akotlar (approved):
Tested:
`cpanm --installdeps .`
`docker build . && docker run bystro-cli:latest bystro-annotate.pl`
`dzil install`

All work as expected

PR #293, opened 2023-10-17
Update Perl CI

wingolab:
Updates github workflows for Perl portions of the Bystro project.

### DIFF
--- a/.github/workflows/perl-package-build.yml
+++ b/.github/workflows/perl-package-build.yml
@@ -5,30 +5,23 @@ on: [push, pull_request]
 jobs:
   test-perl-package:
     runs-on: ubuntu-latest
+    container:
+      image: perldocker/perl-tester:latest
 
     steps:
-      - name: Setup perl
-        uses: shogo82148/actions-setup-perl@v1
-        with:
-          perl-version: ${{ inputs.perl-version }}
-
-      - name: Install Dist::Zilla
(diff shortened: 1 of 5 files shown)


.github/workflows/perl-package-build.yml
```diff
@@ -5,40 +5,38 @@ on: [push, pull_request]
...
         run: |
-          cpanm --quiet https://github.com/bystrogenomics/msgpack-perl.git
-          cpanm --quiet --notest MouseX::Getopt
-          cpanm --quiet DBD::mysql@4.051
+          cpm install -g https://github.com/bystrogenomics/msgpack-perl.git
+          cpm install -g --no-test MouseX::Getopt
+          cpm install -g DBD::mysql@4.051
           git clone --depth 1 --recurse-submodules https://github.com/salortiz/LMDB_File.git \
             && cd LMDB_File \
             && cpanm --quiet .
```
akotlar, line 27: non-blocking minor note, but do we need to continue to use cpanm here?

akotlar (approved):
One minor note, a remaining cpanm use; will merge as is

.github/workflows/perl-package-build.yml
wingolab: The only cpanm we need to use is the one with LMDB_File. The one you point out could be changed.

PR #298, opened 2023-10-17
Perl test cleanup

wingolab:
An approach to tidying up tests by standardizing imports, using temporary directories, and using a testing library for common test components.

The intention is to illustrate an approach to tidying tests so that they are easier to run - i.e., do not produce artifacts because they add or update files under version control - and maintain.


akotlar:
I think this is a good change. Ready to review when out of Draft

wingolab:
This was submitted as PR #303.

PR #299, opened 2023-10-17
Add shellcheck ci

wingolab:
Shell scripts should be checked, linted, or both. This is a simple github workflow for adding shell script checking. Adding linting with shfmt could be done, but, that will not address potentially problematic (usually non-portable) scripts in the same way so I think shellcheck is sufficient.


akotlar:
I think that this is staged on #298 , pulls in additional changes. Once we merge #298, this will need a rebase, because we use the squash merge strategy which rewrites git history, at the advantage of a cleaner commit history for the repo.

wingolab:
You're right about how it was added on top of other changes. It's just a single file so it could be added anytime.

PR #303, opened 2023-10-18
Perl test cleanup

wingolab:
1. Use a common approach to preparing test directory, config file and data that is placed into a temporary directory to avoid leaving behind test artifacts.
2. Update Perl package CI to include bystro-vcf binary to ensure all tests run (since they are skipped if bystro-vcf binary is not present)


akotlar (approved):
Looks good! LGTM

PR #305, opened 2023-10-19
Improve indexing to avoid creating unnecessary inner overlap delimiter

akotlar:
* Simplifies indexing
* Fixes saving parsing regression
* Adds index string -> search _source parsing test

The aim of this PR is to simplify the source structure in indexed documents, and fix a saving regression. We do this by ensuring that if there are no overlap delimited values (the innermost dimension) the innermost value is a scalar.

Briefly, this will ensure that:

het1;het2 will shown as **[ [het1, het2] ]** and not [ [ [het1], [het2] ] ]
and
het1 as **[ [ het1 ] ]**

which will make parsing more straightforward.

Still a bit more work on this one, especially in the need save handler test demonstrating parsing from source document

cc @poneill

### DIFF
--- a/python/python/bystro/search/index/bystro_file.pyx
+++ b/python/python/bystro/search/index/bystro_file.pyx
@@ -95,10 +95,7 @@ cdef class ReadAnnotationTarball:
 
                             overlap_values.append(overlap_value)
 
-                        if len(overlap_values) == 1:
-                            values.append(overlap_values[0])
-                        else:
-                            values.append(overlap_values)
+                        values.append(overlap_values)
 
                     position_values.append(values)
 
--- a/python/python/bystro/search/index/bystro_file.pyx
+++ b/python/python/bystro/search/index/bystro_file.pyx
@@ -95,12 +95,12 @@ cdef class ReadAnnotationTarball:
 
                             overlap_values.append(overlap_value)
 
-                        values.append(overlap_values)
+                        if len(overlap_values) == 1:
+                            values.append(overlap_values[0])
+                        else:
+                            values.append(overlap_values)
 
-                    if len(values_raw) > 1:
-                        position_values.append(values)
-                    else:
-                        position_values.append(values[0])
+                    position_values.append(values)
 
                 populate_hash_path(_source, self.paths[i], position_values)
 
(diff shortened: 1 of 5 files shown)


python/python/bystro/search/index/tests/test_bystro_file.py
```diff
@@ -0,0 +1,88 @@
...
+    df = delims["field"]  # e.g. \t
+    do = delims["overlap"]  # e.g. chr(31)
+    dp = delims["position"]  # e.g. |
+
+    header = f"field1{df}field2{df}field3\n"
+    field1val = f"value1a{dv}value1b{dp}value2aa{do}value2ab{dv}value2b{df}"
+    field2val = f"value3a{dv}value3b{df}"
+    field3val = f"value4a{dp}value4b\n"
+    annotation_content = header + field1val + field2val + field3val
+    print(annotation_content)
```
poneill, line 41: here and throughout, I think we want to either remove print statements or convert them to logger.debug

python/python/bystro/search/index/bystro_file.pyx
```diff
@@ -95,12 +95,12 @@ cdef class ReadAnnotationTarball:
 
                             overlap_values.append(overlap_value)
 
-                        values.append(overlap_values)
+                        if len(overlap_values) == 1:
```
poneill, line 98: honest question: I'm not sure I have enough context here but I think in general I'd opt for collapsing this ambiguity in favor of keeping the data type consistent whether the list has one or many elements, so that values will always be a list of lists of overlap_values (some of which may be singletons).  Otherwise the consuming code will have to make an isinstance check and branch appropriately.

(But if we've considered that alternative and it has other drawbacks that make this the best option, it's a non-blocking concern)

poneill (approved):
LGTM, would just remove / convert print statements.  Other comment is just a non-blocking suggestion I'm not sure is correct

python/python/bystro/search/index/bystro_file.pyx
akotlar: I don't feel strongly either way; let's go that route.

PR #310, opened 2023-10-20
[python/search/index] wait for search index refresh

akotlar:
* Forces wait for index refresh


poneill (approved):
lgtm, would just add a little more context to the commit message :)

PR #311, opened 2023-10-21
update installation and packaging documentation

wingolab:


perl/README.md
```diff
@@ -38,59 +45,62 @@ The remaining dependencies are installed like this:
...
-```
-
 ## Install Bystro using `cpm`
 
 Install [cpm](https://metacpan.org/pod/App::cpm) with `curl -fsSL https://raw.githubusercontent.com/skaji/cpm/main/cpm | perl - install -g App::cpm`.
 
 ```bash
 # install msgpack fork
-cpm https://github.com/bystrogenomics/msgpack-perl.git
+cpm install -g https://github.com/bystrogenomics/msgpack-perl.git
```
akotlar, line 54: we use cpm here and cpanm above; could we go to cpm throughout? I have a change I can make the corresponding change to the perlbrew script

akotlar (changes requested):
Could we remove cpanm in favor of cpm throughout README.md?

perl/README.md
wingolab: The problem with using `cpm` throughout is that `cpanm` is needed for installing LMDB_File - at least I couldn't find a way of using it. If your concern is that it is confusing to have two sections - one for `cpamn` and one for `cpm` - then I suggest consolidation of the two approaches with the acknowledgement that both tools are needed.

Issue #314, opened 2023-10-23
[search/save] Add statistics calculation and statistical filters support 


akotlar:
###### A concise description of why we're pursuing this task.
The Perl implementation of the search saving module (https://github.com/bystrogenomics/bystro/blob/b10/lib/SeqFromQuery.pm) ran the saved annotation through bystro-stats, and also supported pipeline filters (Hardy-Weinberg Equilibrium and Binomial MAF). The current Python implementation does not support this.

###### What is the goal, and how do we know this task is accomplished? 
search/save/handler.py calls into bystro-stats before and packs the statistics outputs into the tarball, and adds filter support.

###### Give enough technical detail for a teammate to understand the proposed solution.
Refer to SeqFromQuery.pm for the implementation to follow. The Python version will follow that directly.

akotlar:
Partially addressed in https://github.com/bystrogenomics/bystro/pull/321

PR #321, opened 2023-10-24
[search/save] Run bystro-stats on saved annotations

akotlar:
* Add running bystro-stats on annotation output
* Add tests for search/utils/annotation
* Fix startup.yml and ensure proteomics server is defined in beanstalkd.yml

Partially addresses #314

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["maturin>=0.14,<0.15", "setuptools", "wheel", "Cython"]
+requires = ["maturin>=1.3.0,<1.4.0", "setuptools", "wheel", "Cython"]
 build-backend = "maturin"
 
 [project]
--- a/python/python/bystro/search/utils/annotation.py
+++ b/python/python/bystro/search/utils/annotation.py
@@ -202,7 +202,6 @@ class Statistics:
 
     @property
     def stdin_cli_stats_command(self) -> str:
-        # Placeholder logic for delimiters
         value_delim = self._delimiters.value
         field_delim = self._delimiters.field
         empty_field = self._delimiters.empty_field
--- a/python/python/bystro/search/utils/annotation.py
+++ b/python/python/bystro/search/utils/annotation.py
@@ -79,19 +79,23 @@ class AnnotationOutputs(Struct, frozen=True, forbid_unknown_fields=True):
         archived: str
             Basename of the archive
         annotation: str
-            Basename of the annotation TSV file, inside the archive
+            Basename of the annotation TSV file, found inside the archive only
         sampleList: Optional[str]
-            Basename of the sample list file, inside the archive
-        log: Basename of the log file, inside the archive
-        statistics: Optional[StatisticsOutputs]
-            Basenames of the statistics files, inside the archive
+            Basename of the sample list file, in the archive and output directory
+        log: str
+            Basename of the log file, in the archive and output directory
+        statistics: StatisticsOutputs
+            Basenames of the statistics files, in the archive and output directory
+        header: Optional[str]
+            Basename of the header file, in the archive and output directory
     """
 
     archived: str
     annotation: str
     sampleList: str
     log: str
-    statistics: StatisticsOutputs | None = None
+    statistics: StatisticsOutputs
+    header: str | None = None
 
     @staticmethod
     def from_path(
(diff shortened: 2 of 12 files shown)


python/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["maturin>=0.14,<0.15", "setuptools", "wheel", "Cython"]
+requires = ["maturin>=1.3.0,<1.4.0", "setuptools", "wheel", "Cython"]
```
akotlar, line 2: Without this we get warnings on building the python project.

poneill: could we just fold that into a code comment?

python/python/bystro/search/utils/tests/test_annotation.py
```diff
@@ -0,0 +1,240 @@
...
+    # Mocking which to return a valid program path
+    mocker.patch("shutil.which", return_value="/path/to/bystro-stats")
+
+    statistics = Statistics("/dummy/output_base_path", None)
+
+    # Note: Your actual values might differ based on the defaults
+    # in StatisticsConfig and DelimitersConfig
+    expected_args = (
+        "/path/to/bystro-stats -outJsonPath /dummy/output_base_path.statistics.json "
+        "-outTabPath /dummy/output_base_path.statistics.tsv "
```
poneill, line 231: should these be long flags here?  (ps disregard if it's correct as written)

poneill (approved):
nice work on the tests-- would recommend revising `from_dict` methods to require valid dicts or raise ValueError in order to avoid silent failures on typos, pls consider everything else non-blocking

python/python/bystro/search/utils/annotation.py
```diff
@@ -1,44 +1,89 @@
...
+    @staticmethod
+    def from_dict(annotation_config: dict[str, Any] | None = None):
+        """Get statistics config from a dictionary"""
+        if annotation_config is None:
+            return StatisticsConfig()
+
+        stats_config: Optional[dict[str, Any]] = annotation_config.get("statistics")
+
+        if stats_config is None:
+            return StatisticsConfig()
```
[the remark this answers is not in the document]
akotlar: It may be, or they may have wanted the defaults (presumably in this scenario this is a user-supplied file). I'll raise a warning.

python/python/bystro/search/utils/annotation.py
```diff
@@ -71,46 +115,115 @@ def from_path(
...
 
-_default_delimiters = {
-    "field": "\t",
-    "position": "|",
-    "overlap": chr(31),
-    "value": ";",
-    "empty_field": "!",
-}
 
+def get_delimiters(annotation_config: dict[str, Any] | None = None):
```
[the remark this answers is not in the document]
akotlar: This was necessary to avoid touching other code that relies on the dict form of DelimitersConfig, but I can fix here.

python/pyproject.toml
akotlar: It's caused by a mismatch between the version in requirements.txt and pyproject.toml. Do you think that's a connection that needs to be drawn (that the packages in [build-system] are required to be in your environment?)

python/python/bystro/search/utils/annotation.py
```diff
@@ -1,44 +1,89 @@
...
+    Paths to all possible Bystro annotation outputs
+
+    Attributes:
+        output_dir: str
+            Output directory
+        archived: str
+            Basename of the archive
+        annotation: str
+            Basename of the annotation TSV file, inside the archive
+        sampleList: Optional[str]
```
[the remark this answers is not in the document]
akotlar: Agreed. The elegance of this system is that our YAML configuration maps 1:1 with source code. I would like to avoid the bookkeeping of needing to rewrite these values on a case by case basis, so instead made a ticket to update the YAML configuration to use snake_case directly: https://github.com/bystrogenomics/bystro/issues/328

python/python/bystro/search/utils/tests/test_annotation.py
akotlar: Correct as written: https://pkg.go.dev/flag

python/pyproject.toml
poneill: I don't have full context here but in general whenever we're pinning on a minor release I've never regretted documenting the reason :). If there's a version mismatch between requirements.txt and pyproject.toml, optimally we'd resolve the mismatch, but if we can't it's probably worth jotting down somewhere why they have to differ.

python/python/bystro/search/utils/annotation.py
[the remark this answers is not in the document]
poneill: it's what the youth crave, true

PR #324, opened 2023-10-24
[WIP] join annotation query results to protein abundance matrix

poneill:
Implement the joining of annotation query results to tandem mass tags datasets on (sample_id, gene_name) pairs.


python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -227,5 +229,15 @@ def get_samples_and_genes_from_query(
 ) -> pd.DataFrame:
     """Given a query and index, return a dataframe of (sample_id, gene, dosage) rows matching query."""
     query = _build_opensearch_query_from_query_string(user_query_string)
-    samples_and_genes_df = _run_annotation_query(query, index_name, client)
-    return samples_and_genes_df
+    return _run_annotation_query(query, index_name, client)
+
+
+def join_annotation_results_to_proteomics(
+    query_result_df: pd.DataFrame, tmt_dataset: TandemMassTagDataset
```
cristinaetrv, line 236: Just to make sure I'm following correctly, the query_result_df going in here including the columns sample_id and gene names is the final result generated by _get_samples_genes_dosages_from_hit that gets kicked off by get_samples_and_genes_from_query?

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
```diff
@@ -15,6 +15,25 @@ class TandemMassTagDataset:
...
+        try:
+            check_df_starts_with_cols(self.abundance_df, ABUNDANCE_COLS[1:])
+        except ValueError as e:
+            err_msg = "Received abundance_df with unexpected columns"
+            raise ValueError(err_msg) from e
+
+    def get_melted_abundance_df(self) -> pd.DataFrame:
+        """Return a melted abundance df with columns [gene_name, sample_id, value]"""
+        abundance_df = self.abundance_df
+        columns_to_drop = ["NumberPSM", "ProteinID", "MaxPepProb", "ReferenceIntensity"]
```
cristinaetrv, line 28: Are these columns always guaranteed? Also, isn't there a chance there are more extra columns than this? I think having extra columns should be allowed without throwing an error as long as we have the expected starting columns. In that case, it might be more worthwhile to focus on columns you want to keep and then subtract out everything you want to drop without specifying the extra column names?

python/python/bystro/proteomics/tests/test_annotation_interface.py
```diff
@@ -55,3 +60,38 @@ def tests__process_response():
...
+
+
+def test_join_annotation_results_to_proteomics_dataset():
+    # Step 1: Get an annotation query result
+    user_query_string = "exonic (gnomad.genomes.af:<0.1 || gnomad.exomes.af:<0.1)"
+    index_name = None
+    mock_client = MockOpenSearch()
+    query_result_df = get_samples_and_genes_from_query(user_query_string, index_name, mock_client)
+
+    # Step 2: Construct a proteomics dataset with some shared sampled_ids and gene_names
```
cristinaetrv, line 72: We may want to make a note somewhere that in future we'll have to take in a third kind of 'tracking ID' that will connect proteomic and genomic sample_ids

cristinaetrv (approved):
Overall lgtm! I would like a little clarification on a couple of points but no blocking comments.

python/python/bystro/proteomics/annotation_interface.py
poneill: yes, maybe `get_samples_and_genes_from_query` should be renamed in order to make that clearer

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
poneill: Yes, these columns are guaranteed by the `__post_init__` validator of this class.  In this method we're converting the df from wide to long format, i.e. converting a matrix with labeled rows and columns to a table of the form `(row_label, col_label, value)`, and that conversion might silently fail if there are additional unexpected columns present.  So I would opt for the strict validator that will fail on unexpected input data, and if it chokes on unexpected data later, the class can be revised in order to accommodate both input types.  Wdyt?

python/python/bystro/proteomics/tests/test_annotation_interface.py
poneill: Good point, do we have a sense now of how we want that tracking ID to work?  Does the relationship look like:

sample -> tracking_id -> (genomic_id, proteomic_id)?

i.e. a sample is associated with a unique tracking_id, and each tracking_id has a unique genomic_id and a unique_proteomic_id that are not necessarily equal to each other?

cristinaetrv: Yes, I believe that's correct.

python/python/bystro/proteomics/fragpipe_tandem_mass_tag.py
cristinaetrv: I think that's a good strategy for now. In future, I'm worried there's a pretty high chance there will be additional columns in the input files, and I can imagine a point at which we might want to be able to keep the extra columns for some kind pre-processing purposes before analysis as we expand functionality.  I think we should be able to take in everything as long as the required columns are present and just drop the extra columns in terms of requirements for analysis. Though, for now, I can just keep this as a note for future work.

python/python/bystro/proteomics/annotation_interface.py
cristinaetrv: maybe just a docstring to indicate where query_result_df comes from to start following the thread? just a little extra clarification would be appreciated

python/python/bystro/proteomics/tests/test_annotation_interface.py
poneill: ok, added in the plumbing to map genomic_ids and proteomic_ids to tracking_ids in the future

PR #327, opened 2023-10-24
Feature/python api 

dlin30:
Still a WIP, it does not include PIT snapshot and missing user_id

Query function will connect to Postgres:
- For experiment name it will return the top_n (default is 10) of the Sample_ID->Subject_ID mappings 
- For query without experiment name it will match/replace on the query string, looking for subject_id, and replacing it with sample_id, and will print out the modify/updated query string 

**Example cli (With Expriment Name):**
python cli.py query --query "Dennis/Alex" --experiment_name Experiment_6  --postgres_config ../../../../config/postgres.yml 

**Expected output:**
Sample ID => Subject ID mappings for Experiment: Experiment_6
{
    "123": "Dennis",
    "24123": "Alex",
   
**Example cli (Without Expriment Name):**
python cli.py query --query "heterozygotes:(Alex && Cristina)" --postgres_config ../../../../config/postgres.yml 

**Expected output:**
Original Query: heterozygotes:(Alex && Cristina)
Modified Query: heterozygotes:(24123 && 34441)


python/python/bystro/api/cli.py
```diff
@@ -464,6 +468,76 @@ def get_user(args: argparse.Namespace, print_result=True) -> UserProfile:
...
+        conn = psycopg2.connect(
+            host=db_host,
+            port=db_port,
+            user=db_user,
+            password=db_password,
+            database=db_database
+        )
+        cursor = conn.cursor()
+
+        if args.experiment_name:
```
akotlar, line 504: Instead of doing this, let's just use the existing webapp endpoint we have

python/python/bystro/api/cli.py
```diff
@@ -464,6 +468,76 @@ def get_user(args: argparse.Namespace, print_result=True) -> UserProfile:
...
+            if not rows:
+                print(f"\nExperiment name '{args.experiment_name}' does not exist.")
+                return
+            
+            sample_subject_mappings = {row[0]: row[1] for row in rows}
+            print("\nSample ID => Subject ID mappings for Experiment:", args.experiment_name)
+            print(json.dumps(sample_subject_mappings, indent=4))
+            return
+
+        sql_query = "SELECT sample_id, subject_id FROM user_experiments"
```
akotlar, line 523: When the user doesn't supply an experiment_name we don't want to query user_experiments. The user hasn't given us a mapping, and subject_id's are not guaranteed to be unique across experiments.

akotlar (changes requested):
Left comments

dlin30:
PR has been updated to follow the new structure in corresponding with bystro web PR [#336](https://github.com/bystrogenomics/bystro-web/pull/336) and will be using the already established endpoint.

INSTALL.md
```diff
@@ -127,3 +127,19 @@ The outputs will be:
...
+   - sudo yum install openssl-devel
+- rustup
+   - curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+   - source ~/.bashrc
+   - rustc --version
+
+Install Anaconda (current version as of Oct 18, 2023)
+- wget https://repo.anaconda.com/archive/Anaconda3-2023.09-0-Linux-x86_64.sh
+- bash Anaconda3-2023.09-0-Linux-x86_64.sh 
+- source ~/.bashrc
```
akotlar, line 145: This is a duplicate of https://github.com/bystrogenomics/bystro/blob/master/INSTALL.md#installing-the-bystro-python-libraries-and-cli-tools. Please take a look at that section and make it more discoverable, so that our next dev has an easier time finding it, it's a bit buried.

openssl-devel is also a dependency in our existing install script https://github.com/bystrogenomics/bystro/blob/master/install/install-rpm-deps.sh

python/python/bystro/api/cli.py
```diff
@@ -1,8 +1,11 @@
 import argparse
 import datetime
 import os
+import sys
 
 import requests
+import psycopg2
```
akotlar, line 7: not needed

python/python/bystro/api/cli.py
```diff
@@ -464,6 +467,57 @@ def get_user(args: argparse.Namespace, print_result=True) -> UserProfile:
 
     return user_profile
 
+def query(args: argparse.Namespace) -> None:
+    """
+    Perform a query with the given arguments.
```
akotlar, line 472: let's expand on what arguments are expected

python/python/bystro/api/cli.py
```diff
@@ -572,6 +626,14 @@ def main():
     )
     jobs_parser.set_defaults(func=get_jobs)
 
+    query_parser = subparsers.add_parser("query", help="Perform a query")
+    query_parser.add_argument("--dir", default=DEFAULT_DIR, help="Where Bystro API login state is saved")
+    query_parser.add_argument("--query", required=True, help="Query properties")
+    query_parser.add_argument("--size", default=10, type=int, help="How many records, max should be the maximum the server supports (default: 10)")
+    query_parser.add_argument("--from_value", default=0, type=int, help="From value from which record (default: 0)")
```
akotlar, line 633: Let's change this to --from and then add the `dest='from_'` value (you'll need to replace `from_value` with `from_`). `from_` is a bit more idiomatic, see https://peps.python.org/pep-0008/

"""
If a function argument’s name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus class_ is better than clss. (Perhaps better is to avoid such clashes by using a synonym.)
"""

akotlar (changes requested):
See comments

python/python/bystro/api/cli.py
```diff
@@ -572,6 +626,14 @@ def main():
     )
     jobs_parser.set_defaults(func=get_jobs)
 
+    query_parser = subparsers.add_parser("query", help="Perform a query")
+    query_parser.add_argument("--dir", default=DEFAULT_DIR, help="Where Bystro API login state is saved")
+    query_parser.add_argument("--query", required=True, help="Query properties")
```
akotlar, line 631: Let's change the help for query to "The OpenSearch query string query, e.g. (cadd: >= 20)"

python/python/bystro/api/cli.py
```diff
@@ -572,6 +626,14 @@ def main():
...
+    query_parser.add_argument("--size", default=10, type=int, help="How many records, max should be the maximum the server supports (default: 10)")
```
akotlar, line 632: "max should be the maximum the server supports " - let's remove this, that is not a concern for users, that was mostly a note for us to consider.

python/python/bystro/api/cli.py
akotlar: The help string should be more explanatory: "The first record to return from the matching results. Used for pagination."

python/python/bystro/api/cli.py
```diff
@@ -572,6 +626,14 @@ def main():
...
+    query_parser.add_argument("--from_value", default=0, type=int, help="From value from which record (default: 0)")
+    query_parser.add_argument("--job_id", required=True, type=str, help="The job the user wants to query")
```
akotlar, line 634: "The job id to query"

INSTALL.md
```diff
@@ -1,3 +1,17 @@
+# Table of Contents
```
akotlar, line 1: awesome!

python/python/bystro/api/cli.py
```diff
@@ -572,6 +640,14 @@ def main():
     )
     jobs_parser.set_defaults(func=get_jobs)
 
+    query_parser = subparsers.add_parser("query", help="The OpenSearch query string query, e.g. (cadd: >= 20)")
+    query_parser.add_argument("--dir", default=DEFAULT_DIR, help="Where Bystro API login state is saved")
+    query_parser.add_argument("--query", required=True, help="The OpenSearch query string query, e.g. (cadd: >= 20)")
+    query_parser.add_argument("--size", default=10, type=int, help="How many records (default: 10)")
+    query_parser.add_argument("--from_", default=0, type=int, help="The first record to return from the matching results. Used for pagination.")
```
akotlar, line 647: minor nit: let's remove the period after "Used for pagination" to match the periodless other help strings

akotlar (approved):

PR #329, opened 2023-10-25
[annotation] uniquefy Bystro outputs

akotlar:
* If all delimited outputs are duplicates, output only the single unique value
* Join ref on '', converting A|C|G|T => ACGT in the del case, or A|C => AC in the ins case. SNV are unaffected

Needs tests, but worth a look anyway. Live on bystro-dev.emory.edu

This is motivated by Pat's feedback and agreed upon with Dave (going to have him double check the output of a test case just in case)

Attached before & after annotations:
[before_trio_trim_vep_vcf.annotation.tsv.zip](https://github.com/bystrogenomics/bystro/files/13161175/before_trio_trim_vep_vcf.annotation.tsv.zip)
[after_trio_trim_vep_vcf.annotation.tsv.gz](https://github.com/bystrogenomics/bystro/files/13172838/after_trio_trim_vep_vcf.annotation.tsv.gz)

### DIFF
--- a/perl/lib/Seq/Output.pm
+++ b/perl/lib/Seq/Output.pm
@@ -9,7 +9,6 @@ use List::Util qw/min max/;
 
 use Seq::Output::Delimiters;
 use Seq::Headers;
-use DDP;
 
 with 'Seq::Role::Message';
 
@@ -22,6 +21,11 @@ has delimiters => (
   default => sub {
     return Seq::Output::Delimiters->new();
   }
+);
+
+has refTrackName => (
+  is      => 'ro',
+  isa     => 'Str',
 );
 
 sub BUILD {
@@ -52,6 +56,11 @@ sub BUILD {
 
     push @{ $self->{_trackOutIndices} }, $outIdx;
 
+    if($trackName eq $self->refTrackName) {
+      $self->{_refTrackIdx} = $outIdx;
+      next;
+    }
+
     if ( ref $trackName ) {
       $self->{_trackFeatCounts}[$outIdx] = $#$trackName;
     }
@@ -61,8 +70,23 @@ sub BUILD {
   }
 }
 
-# TODO: will be singleton, configured once for all consumers
-sub initialize {
+sub uniqueify {
+    my %count;
+    my $undefCount = 0;
+
+    foreach my $value (@{$_[0]}) {
+      if(!defined $value) {
+        $undefCount++;
+      } else {
+        $count{$value} = 1;
+      }
+    }
+
+    if($undefCount == @{$_[0]} || scalar keys %count == 1) {
+        return [$_[0]->[0]];
+    }
+
+    return $_[0];
 }
 
 # ABSTRACT: Knows how to make an output string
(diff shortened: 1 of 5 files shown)


perl/lib/Seq/Output.pm
```diff
@@ -61,8 +70,24 @@ sub BUILD {
   }
 }
 
-# TODO: will be singleton, configured once for all consumers
-sub initialize {
+sub uniqueify {
```
poneill, line 73: seems worth a test or two just to verify that the tricky bits like checking for undefined values, converting `keys %count` to scalar are working as intended.

(Technically this sub could return `$_[0]` as soon as we detect two distinct values, but unless it's in a tight inner loop and the input array is typically large the perf probably isn't worth the additional complexity)

perl/lib/Seq/Output.pm
```diff
@@ -131,13 +161,13 @@ sub makeOutputString {
               ref $_
               ?
                 # at this position this feature has multiple values
-                join( $valDelim, map { defined $_ ? $_ : $missChar } @$_ )
+                join( $valDelim, map { defined $_ ? $_ : $missChar } @{ uniqueify($_) } )
```
poneill, line 164: Maybe I just don't know perl, but I got tripped up on this line.  It looks like $_ is being used in two different scopes here? (the scope of $_ passed to uniqueify and an inner scope for the map).  Pls disregard if this is an idiomatic construct in perl but I would consider recasting this line to avoid the two-scopes problem if possible.

perl/lib/Seq/Output.pm
```diff
@@ -170,19 +200,25 @@ sub makeOutputString {
 
           $row->[$oIdx][$featIdx] = join(
```
poneill, line 201: is there any way to let this expression breathe a bit?  It seems like the intent of this line is pretty straightforward, but I'm finding it a little tough to parse the three nested ternary operators.  It looks like we use the construct `map {defined $_ ? $_ : $missChar}` in a few other places as well, suggesting that we could pull it out into a helper named `fillUndefinedWithMissChar` (or something) and cut back on the nesting a bit.

perl/lib/Seq/Output.pm
```diff
@@ -210,23 +246,30 @@ sub makeOutputString {
           # t1;t2 \t t1_name1\\t1_name2;t2_onlyName
           $posData = join(
```
poneill, line 247: in fact it looks like this whole expression could be pulled out into a helper?

I notice there's a slight difference between these two expressions: the first assigns to `$row->[$oIdx][$featIdx]` a function of `$row->[$oIdx][$featIdx][0]`, and the second assigns to `$posData` the same function of `$posData`.    I am probably missing context here but that almost-but-not-exactly-identical expressions might benefit from a comment explaining the difference.

poneill (approved):

perl/lib/Seq/Output.pm
akotlar: Yeah I didn't micro bench, branching in loops can be cache unfriendly and most of the arrays it will checking are very small (<3 length I would guess)

perl/lib/Seq/Output.pm
akotlar: So 99% of sites won't have positional data, because only roughly 1% of all mutations are indels. It turns out that not entering the loop over position indices saves something like 20% runtime (tested on a small run, will benchmark more in depth on larger ones later).

akotlar: Also, we save a large number of allocations by directly mutating the array, which in perl you can do as seen because $posData in `for my $posData (@data)` is a reference to the array entry.

cristinaetrv (approved):
I've reviewed the before and after outputs and the changes look good to me. 

I would consider changing around nearest.name and nearest.name2 so they're consistent with the order of the refseq name and name 2 fields, but this could be a separate PR. We also discussed potentially adding a test for the case of the refSeq.name field having a missing value among non-missing values such as NM_1;!;NM_2 or !;NM_2, which should never happen in that field. These cases are frequently seen in refSeq.mRNA but they are expected in refSeq.mRNA.

PR #331, opened 2023-10-26
[proteomics] Provide Uniprot ID / Hugo symbol crosswalk

poneill:
This PR provides:

- a module for bi-directional lookup of gene name from uniprot id and vice versa
- the data necessary for that module
- a script for downloading that data


python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+
+"""Download a list of uniprot_id / gene_name pairs from Uniprot and store the result on disk as csv."""
+
+import pandas as pd
+import tqdm
+from unipressed import UniprotkbClient  # type: ignore[import]
```
cristinaetrv, line 7: Will this allow us to support non-human proteomes as well? Since we're supporting non-human genomes, there's a pretty good chance we might want to support non-human proteome uniprot IDs.

python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
```diff
@@ -0,0 +1,42 @@
...
+
+from bystro.utils.config import BYSTRO_PROJECT_ROOT
+
+
+def get_uniprot_id_gene_name_mapping() -> pd.DataFrame:
+    """Get a dataframe associating each uniprot id with its cognate gene names."""
+    APPROXIMATE_TOTAL_RECORDS = 240_000
```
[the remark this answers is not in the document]
poneill: I think the 10x inflation factor mostly come from isoforms, see e.g. https://www.uniprot.org/uniprotkb/A0A8V8TRG9/entry, an isoform of the canonical https://www.uniprot.org/uniprotkb/P42345/entry for MTOR.

In any case the approximate total is just for a progress bar to display to the user while downloading.  We can keep the progress bar without assuming a total, though.

python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
poneill: The client will allow us to support non-human proteomes, yes.  We could just download everything, or at least everything for the reference genomes we support.  Given that this script is speculative development (we don't actually know that we'll need it even for human datasets), would you feel comfortable noting it as a todo for now?

python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    uniprot_id_gene_name_mapping = []
+    for record in tqdm.trange(query_results.each_record(), total=APPROXIMATE_TOTAL_RECORDS):
+        genes = record.get("genes", [])  # type: ignore[attr-defined]
+        gene_names = []
+        for gene in genes:
+            if (gene_name := gene.get("geneName")) and (value := gene_name.get("value")):
+                gene_names.append(value)
+        if not gene_names:
+            gene_names = [None]
```
cristinaetrv, line 28: I think since our missing convention is '!' in the annotation, it makes more sense to go with that. I'm not sure if we longer-term want to necessarily not bring in regions of the genome that don't have a gene name in Swiss-Prot but do have a TrEMBL uniprot ID to the joined data in case some of the top fold changes for the proteomics are in that category and users still want to see the variants specified to that region. For the current version it makes the most sense to drop them. I'm making a note here that we may want to include whether something is in Swiss-Prot or TrEMBL as a separate column in future for those kinds of cases (Example UniprotID: D3WYY1).

python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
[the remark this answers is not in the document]
cristinaetrv: Ah ok sounds good. I'm just guessing the number will change over time as well as IDs are reviewed and added into Swiss-Prot.

python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    return uniprot_id_gene_name_mapping_df
+
+
+if __name__ == "__main__":
+    uniprot_id_gene_name_mapping_df = get_uniprot_id_gene_name_mapping()
+    output_filename = (
+        BYSTRO_PROJECT_ROOT / "python/python/bystro/proteomics" / "uniprot_id_gene_name_mapping.csv"
+    )
+    uniprot_id_gene_name_mapping_df.to_csv(output_filename, index=False)
```
cristinaetrv, line 46: So the goal of this PR is to create this mapping file of uniprot IDs to gene names, so then it's on the user to replace them? Am I understanding that correctly? The reason I ask is because I think that users that will find it straightforward to replace uniprot IDs with Hugo gene names using a mapping file will likely find it straightforward to replace uniprot IDs with Hugo gene names on their own and users that may find that difficult could have trouble with the next step before their proteomics file is ready for proteomic/genomic combined analyses. But if the intention is just to have this available to make that conversion easier on our side if we need to then I'm happy with that.

python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
cristinaetrv: Yep, I'm happy with that being a todo for now.

python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
poneill: No, I don't think it should ever be incumbent upon the user to perform that mapping themselves.

Earlier we were worried that proteomics datasets might use Uniprot IDs for to reference peptides, whereas the annotation file uses HUGO symbols for gene names.  But it looks like Fragpipe is converting Uniprot identifiers to HUGO symbols for us already, so our concern was misplaced, which makes this PR superfluous.  @akotlar recommended PRing it anyway for now, so that if we later encounter a use case where such a mapping is necessary, the building blocks will already be in place.

python/python/bystro/proteomics/scripts/get_uniprot_id_gene_name_mapping.py
poneill: The mapping csv isn't intended to be user-facing, but only accessed through the pair of public functions `get_uniprot_ids_from_gene_name` and `get_gene_names_from_uniprot_id`.  Pandas natively understands conventional null values like `None` or `np.nan` to represent missing data, but doesn't understand `"!"` that way.  My recommendation would be to continue to use pythonic representations of missing data except where necessary to preserve backwards compatibility in user-facing data, wdyt?

Happy to drop the uniprot_ids without corresponding gene names entirely, though.

cristinaetrv: Ok got it, as long as they don't interact with any potential querying (i.e. clinvar gets connected to this at some point and a record with Ubiquinone gets pulled if you were trying to query where the gene name is 'None') and it's not user-facing, I'm ok with that. May be safer to go with np.nan just in case.

I do think for now we should just drop uniprot_ids without corresponding gene names in Swiss-Prot entirely with a todo to add that information to a future prot tutorial for users unless we decide to add TrEMBL at some point and mark those accordingly.

cristinaetrv (approved):
Lgtm! Approved pending a couple todo's marking what was discussed in the comments.

PR #332, opened 2023-10-26
[save/search] Add pipeline tagged union type to SaveJobData

akotlar:
* Add pipeline tagged union type to SaveJobData

This is the first step in supporting filter pipelines, which are procedures that run in a streaming manner to filter values from the query we're saving.

No tests written because it is trivial to test (it is live now on bystro-dev), and we will write tests in the followup PR that actually implements filters.


python/python/bystro/search/utils/messages.py
```diff
@@ -30,6 +61,7 @@ class SaveJobData(BaseMessage, frozen=True):
     indexName: str
     outputBasePath: str
     fieldNames: list[str]
+    pipeline: list[BinomialMafFilter | HWEFilter] | None = None
```
poneill, line 64: if the semantics of `None` here are equivalent to an empty list of filters, we could reduce this type to `list[BinomialMafFilter | HWEFilter]` and avoid explicit checks for `None`

python/python/bystro/search/utils/messages.py
```diff
@@ -22,6 +27,32 @@ class IndexJobCompleteMessage(CompletedJobMessage, frozen=True, kw_only=True):
...
 
+class BinomialMafFilter(
+    Struct, frozen=True, tag="binomMaf", tag_field="key", forbid_unknown_fields=True
+):
+    """
+    Example: {'critValue': '.05', 'estimates': ['gnomad.genomes.af', 'gnomad.exomes.af'],
+             'privateMaf': 0.0001, 'snpOnly': True, 'numSamples': 3, 'key': 'binomialMaf'}
+    """
+
+    privateMaf: float
```
poneill, line 38: nit: would recommend converting these to snake_case

python/python/bystro/search/utils/messages.py
```diff
@@ -22,6 +27,32 @@ class IndexJobCompleteMessage(CompletedJobMessage, frozen=True, kw_only=True):
     results: IndexJobResults
 
 
+class BinomialMafFilter(
+    Struct, frozen=True, tag="binomMaf", tag_field="key", forbid_unknown_fields=True
```
poneill, line 31: What are the tradeoffs of this approach to tagging vs `tag=True`?  The latter approach would add a key-value pair `"type":"BinomialMafFilter"`to the json, which (at least imo) seems easier to parse as a piece of metadata indicating what kind of Struct the json represents.  This approach would add `"key":"binomMaf"` which seems a little more open to interpretation, wdyt?

poneill (approved):

python/python/bystro/search/utils/messages.py
```diff
@@ -22,6 +27,32 @@ class IndexJobCompleteMessage(CompletedJobMessage, frozen=True, kw_only=True):
...
+):
+    """
+    Example: {'critValue': '.05', 'estimates': ['gnomad.genomes.af', 'gnomad.exomes.af'],
```
cristinaetrv, line 34: Can you add a brief description of what the binomial maf filter is useful for here? Especially since it's not as common as the HWE step in genomics workflows.

python/python/bystro/search/utils/messages.py
akotlar: `tag=True` would require our webapp frontend to be dependent on what we name this class. `key: "binomMaf"` is what the frontend sends (and has for years), and what SeqFromQuery.pm expected.

python/python/bystro/search/utils/messages.py
akotlar: Done!

PR #333, opened 2023-10-27
[proteomics] add README

poneill:
Add README for proteomics module


akotlar (approved):
Outdated, but a start

PR #334, opened 2023-10-29
use zstd for all mapping codecs

akotlar:
* Uses the same zest compression algorithm for all mapping configs

ZSTD brings throughput benefits vs "best_compression" https://opensearch.org/docs/latest/im-plugin/index-codecs/


akotlar:
I'm going to close this for now, doesn't work with KNN search, which we may want

PR #337, opened 2023-11-05
[feature/ancestry-ppca] Ignore fewer import types

akotlar:


python/python/bystro/covariance/_base_precision.py
```diff
@@ -54,11 +54,13 @@
...
 _get_covariance(covariance)
 """
-from numpy import linalg as la
 from datetime import datetime as dt
-from ._base_covariance import BaseCovariance
+
+from numpy import linalg as la
 import pytz
 
+from bystro.covariance._base_covariance import BaseCovariance
```
akotlar, line 62: The traditional way to structure imports is to make them alphabetical, and ordered as follows:

1. Python stdlib
2. 3rd party libraries
3. Our libraries

PR #339, opened 2023-11-08
Fix crashing save handler

akotlar:
Don't requires total hits found in slice to match the number of records returned, this is not guaranteed to be the case


austinTalbot7241993 (approved):
Why did we alter max_slices default? Otherwise seems fine

akotlar:
> Why did we alter max_slices default? Otherwise seems fine



Good question. The number of slices determines the max number of threads. 128 for a single save job is more than enough.

Issue #261, opened 2023-09-11
Documentation: Get consensus on how to handle the overlap delimiter. Control character seems problematic


akotlar:
Right now we have a "UNIT SEPARATOR" overlap delimiter, which is not rendered, typically. This is used because we have cases where we have multiple hits from a database, for a given annotation field, at a given position/allele.

We want to move instead on to something that is rendered

akotlar:
The result is that we will use "/"

PR #343, opened 2023-11-10
[annotator] Add script to extract AN, AF, nhomalt

akotlar:
* Switch to "/" overlap delimiter, see https://github.com/bystrogenomics/bystro/issues/261
* Add script to extract the hundreds of populations from gnomad so that we don't need to type so much to build an annotation config

### DIFF
--- a/perl/t/tracks/build/coerceFeatureType.t
+++ b/perl/t/tracks/build/coerceFeatureType.t
@@ -43,7 +43,8 @@ $expected =
   '2: Homo sapiens BRCA1,BRCA2-containing complex subunit 3 (BRCC3), transcript variant 2, mRNA. (from RefSeq NM_001018055)';
 $res = $seq->coerceFeatureType( 'someString', $str );
 
-ok( $res eq $str && $str eq $expected, 'Can clean string containing / followed by a-zA-Z' );
+ok( $res eq $str && $str eq $expected,
+  'Can clean string containing / followed by a-zA-Z' );
 
 $str =
     '3: Homo sapiens BRCA1'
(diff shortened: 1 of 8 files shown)


perl/t/tracks/build/coerceFeatureType.t
```diff
@@ -24,12 +24,12 @@ my $seq = Seq::Tracks::Build->new(
...
 
 # unit separator
-my $overlap = "\x1f";
+my $overlap = "/";
 
 my $str =
   '1: Homo sapiens BRCA1/BRCA2-containing complex subunit 3 (BRCC3), transcript variant 2, mRNA. (from RefSeq NM_001018055)';
 my $expected =
-  '1: Homo sapiens BRCA1/BRCA2-containing complex subunit 3 (BRCC3), transcript variant 2, mRNA. (from RefSeq NM_001018055)';
+  '1: Homo sapiens BRCA1,BRCA2-containing complex subunit 3 (BRCC3), transcript variant 2, mRNA. (from RefSeq NM_001018055)';
```
cristinaetrv, line 32: Does this example still make grammatical sense? Would it make sense to add a '-' instead of a comma (e.g. This gene encodes a subunit of the BRCA1-BRCA2-containing complex (BRCC)) or is it possible that it wouldn't always make sense to replace with a hyphen?

perl/t/tracks/build/coerceFeatureType.t
```diff
@@ -63,7 +65,8 @@ $expected =
   '4: Homo sapiens BRCA1,(BRCA2)-containing complex subunit 3 (BRCC3), transcript variant 2, mRNA. (from RefSeq NM_001018055)';
 $res = $seq->coerceFeatureType( 'someString', $str );
 
-ok( $res eq $str && $str eq $expected, 'Can clean string containing \x1f(' );
+ok( $res eq $str && $str eq $expected,
+  'Can clean string containing / followed by a special character, a parenthsis' );
```
cristinaetrv, line 69:
```suggestion
  'Can clean string containing / followed by a special character, a parenthesis' );
```

cristinaetrv (approved):
Overall looks good, no blocking comments but found small typo and have question about grammar after delimiter changes.

perl/t/tracks/build/coerceFeatureType.t
akotlar: Ok, I can make a ticket for this; we can do this in the YAML config, with a build_field_transformation

Issue #256, opened 2023-09-11
Documentation: Gather feedback on annotations to add for V1.0.0 release (do we add all gnomad annotations, etc)


akotlar:
* Identify and document the annotations we'll launch with for December.

akotlar:
Feedback from Dave on Gnomad:
- All of it, every pop. We'll keep storing the an and af annotations for each population

poneill:
Just verifying here that the feedback's been collected.

PR #344, opened 2023-11-11
[annotator] Move cadd to exact match

akotlar:
* Following issue #256, switch CADD to exact match, and add tests


perl/lib/Seq/Tracks/Cadd.pm
```diff
@@ -45,12 +45,33 @@ sub get {
...
+    return $_[6];
+  }
+
   if ( !defined $_[1]->[ $_[0]->{_dbName} ] ) {
     $_[6][ $_[5] ] = undef;
 
     return $_[6];
   }
 
+  # For indels, we return undef
```
cristinaetrv, line 64: Is 'undef' a standard within CADD or other databases that might be expected by users or would NA/nan make sense in these cases too?

perl/t/tracks/cadd/integration.t
```diff
@@ -0,0 +1,157 @@
...
+    my $bystro_style_ins_skipped_tiling =
+      $caddGetter->get( $data, $chr, $ref, "+ACTG", 0, \@out );
+
+    ok( !defined $bystro_style_ins_skipped_tiling->[0] );
+    ok( @{$bystro_style_ins_skipped_tiling} == 1 );
+
+    $pos += 1;
+  }
+
+  # We don't currently support VCF style deletions, but we should still return undef
```
cristinaetrv, line 142: Can you describe a little more what you mean by VCF style deletions versus the type that we do support? I would also say that they seem more 'unavailable' or 'unsupported' versus 'undefined' if they do have a score but we're just unable to support it in the current version, but I think these could also be replaced by the missing NA/nan and we specify what we're able to support in our CADD section of the field descriptions.

perl/t/tracks/cadd/integration.t
```diff
@@ -0,0 +1,157 @@
...
+  my $vcf_style_del_skipped =
+    $caddGetter->get( $firstData, $firstChr, $firstRef . "CTG", $firstRef, 0, [] );
+
+  ok( !defined $vcf_style_del_skipped->[0] );
+  ok( @{$vcf_style_del_skipped} == 1 );
+
+  # We don't currently support VCF style insertions, but we should still return undef
```
cristinaetrv, line 149: Same as above - could you specify the difference between VCF/non-VCF style indels and what we do support?

cristinaetrv (dismissed):
Requested clarification in a couple areas and suggested potentially changing 'undef' to the 'missing' character.

perl/lib/Seq/Tracks/Cadd.pm
akotlar: Undef is the way that bystro internally handles missing values; I'll make sure it's described better

akotlar: You don't/can't use the missing character, because then we would mix functionality (we have a dedicated Output creator, and we want multiple possible Output backends)

cristinaetrv (approved):
Comments are clear now, lgtm!

PR #346, opened 2023-11-12
[annotator] Add no_build option which allows tracks to be skipped

akotlar:
* Add no_build option which allows tracks to be skipped
* Cleans up unnecessary DDP imports

This is necessary, because in some cases, like join tracks, we may not want to build the track referenced in the join definition. This comes up with clinvar, which defines large variants and small in the tsv output, but not in the vcf output. So we define a joint track "refSeq.clinvar" to show those large structural variants that intersect with refSeq, but do not want to build the small variants in the actual clinvar track, because those are inexact in position (we instead now have an exact match clinvar2).

### DIFF
--- a/perl/lib/Seq/Tracks/Build.pm
+++ b/perl/lib/Seq/Tracks/Build.pm
@@ -41,9 +41,6 @@ has db => (
   }
 );
 
-# Don't actually build, used if we want a track to serve as a join track
-has no_build => ( is => 'ro', isa => 'Bool', default => 0 );
-
 # Allows consumers to record track completion, skipping chromosomes that have
 # already been built
 has completionMeta => (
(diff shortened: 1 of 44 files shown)


perl/lib/Seq/Tracks/Build.pm
```diff
@@ -156,7 +158,6 @@ sub BUILD {
   my $d = Seq::Output::Delimiters->new();
   $self->{_cleanDelims} = $d->cleanDelims;
   $self->{_missChar}    = $d->emptyFieldChar;
-  $self->{_replChar}    = $d->globalReplaceChar;
```
akotlar, line 159: this is unused

perl/t/tracks/gene/join_no_build.yml
```diff
@@ -0,0 +1,78 @@
+---
```
cristinaetrv, line 1: Could you add a comment here pointing to what the no build will affect?

cristinaetrv (approved):
Overall lgtm! I reviewed the outputs during our discussion on this. I recommended potentially adding a description in the join_no_build.yml to point to make it clear what isn't getting built.

PR #350, opened 2023-11-14
[annotator] Cast binomial maf filter values to float

akotlar:
* Cast the values used by the binomial MAF filter to floats, since we cannot guarantee that the inputs are floats.

The apparent extra commits are an artifact of squash merges.


cristinaetrv (approved):
Pretty straightforward, lgtm!

PR #351, opened 2023-11-14
[annotator] Move the discordant field, and have `inputRef` next to `alt`

akotlar:
* The 3rd field is now inputRef, the input reference coming from the file preprocessor
* The discordant field is now the first non-preprocessor field

This PR addresses Dave's suggestion that we move the ref next to the alt. Because our 'ref' is the reference track, in this implementation we name the ref coming from the input file 'inputRef'.

Example output attached, as well as before the change

After:
[after_ALL_chr1_annotation.annotation.100lines.tsv.zip](https://github.com/bystrogenomics/bystro/files/13344390/after_ALL_chr1_annotation.annotation.100lines.tsv.zip)

Before:
[before_ALL_chr1_annotation.annotation.100lines_v1.tsv.zip](https://github.com/bystrogenomics/bystro/files/13344410/before_ALL_chr1_annotation.annotation.100lines_v1.tsv.zip)

Dave / Thomas like the placement of discordant and inputRef

### DIFF
--- a/perl/lib/Seq.pm
+++ b/perl/lib/Seq.pm
@@ -5,7 +5,6 @@ use warnings;
 package Seq;
 
 our $VERSION = '0.001';
-
 # ABSTRACT: Annotate a snp file
 
 # TODO: make temp_dir handling more transparent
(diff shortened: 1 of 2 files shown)


cristinaetrv (approved):
I reviewed the attached before/after output files and I also like the position of discordant and ref/inputref.

Issue #302, opened 2023-10-18
Annotation Sprint 3 Task List


cristinaetrv:
- [x] Switch from using ClinVar tab separated to VCF file for exact match (tab separated gives no guarantee that allelic representation matches)

- [x] Replace unit separator with forward slash, rebuild database with that in place

- [x] Add CADD indel scores

- [x] Remove position delimiter 

- [ ] #349
heterozygotes:(4805 && 1805) cadd > 20
and
heterozygotes:(4805 1805) cadd > 20 (no &&) work the same
Previously, in Elasticsearch 5.6 (b10), these were equivalent

- [x] Check for number of result deviation between && and no &&, and other issues
- [x] Update search/save handler to generate statistics and support statistical filters (https://github.com/bystrogenomics/bystro/issues/314)

akotlar:
#343 , #344 , #345 , #346 , #347  close this ticket

The "Fix opensearch regression making" will need to wait for future sprint. Needs more investigation, we appear to be using the appropriate setting for setting spaces as "ands", and updating our query parser to fix any problems in Opensearch are outside this scope. There remains a possibility that there is a new setting that we have missed, and in sprint 4 I will investigate that.

PR #356, opened 2023-11-23
Improve indexing performance

akotlar:
* Switches us back to a lightly modified version of the Perl indexer, which performs better than the Python version, and is more battle tested
* Fixes the hg19 index to work with the latest field definitions in the v6 database (all of the new gnomad, caddIndel, clinvervcf, dbSNP annotations introduced as part of issue #302)
* Also updates several python dependencies, and removes unused dependencies

Both bystro-index-server.pl and SeqElastic.pm can be lightly reviewed. 

These came from the b10 branch with only the modifications needed to 1) support the updated "addresses" field in beanstalk.yml (and the removal of the events field), and the storage of consistently 3D array values (whereas in b10 we would store 1D arrays where possible).

The movitation here: The new annotation outputs are so large that they seem to be causing memory pressure for Ray, which is really optimized for numpy arrays not raw strings. This seems to result in many copies and inefficiency

I have made a Go version of the codebase, and a Rust version, but we have enough on our plate to avoid introducing completely new codebases, and both libraries are severely limited by network performance anyhow. So instead, I figured out how to save some processing time (avoiding making the nested JSON, which Opensearch 2 supports, it will automatically convert field.child to field: {child: value}), and moved back to the stable codebase.

I will leave the python code for now, because we may return to it in 2024 (at least the handler.py may be reused). This may go away in the future.

### DIFF
--- a/config/hg19.mapping.yml
+++ b/config/hg19.mapping.yml
@@ -27,6 +27,9 @@ index_settings:
     refresh_interval: -1
     number_of_replicas: 0
     codec: best_compression
+    mapping:
+      total_fields:
+        limit: 5000
   analysis:
     normalizer:
       lowercase_normalizer:
@@ -378,7 +381,6 @@ mappings:
               type: integer
             chromEnd:
               type: integer
-
     nearest:
       properties:
         refSeq:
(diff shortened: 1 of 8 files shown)


akotlar:
Closing in favor of https://github.com/bystrogenomics/bystro/pull/357

PR #357, opened 2023-11-25
Multithreaded Go OpenSearch indexer

akotlar:
* A multithreaded search index creator (from annotation file) that is up to 10x faster on the same hardware than previous python/perl versions. 1000 genomes Phase 3 chromosome 1 indexes in roughly 33 minutes, down from 3 - 6 hours . trio.trim.vcf.gz (13341 variants) indexes in 6-10 seconds, down from 40-50s.
  - Compared to b10 running on Dave/Emory Integrated Genomics indexing takes around 2.5x more time per thread, but we're indexing 7x the data, an acceptable tradeoff and as indexing scales almost linearly with the number of threads, in practice we achieve similar performance as b10 using marginally more expensive hardware.
* Removes no-longer needed python indexer
* Updates python dependencies
* Fixes perl/README.md `cpm` references to include `cpm install`
* Creates helpful Bystro-server.pl usage string if required arguments are not provided, and removes unused variables

This also sets us up the go project structure needed to pull bystro-vcf, bystro-stats into the main Bystro repo to simplify future development.
Screenshot 2023-11-27 at 12 37 55 AM


go/cmd/opensearch/main.go
```diff
@@ -0,0 +1,656 @@
...
+	var lines []string
+
+	for {
+		rawLines, err := reader.readLines()
+
+		if len(rawLines) > 0 {
+			lines = strings.Split(string(rawLines), parser.LINE_DELIMITER)
+		}
+
+		if err != nil {
```
wingolab, line 574: 1. Typically, you'd check for an error before doing anything with returned data. 
2. Ultimately, you read a chunk of lines and then split those lines. It would seem more straightforward to use a bufio.Scanner that is reading from either the gzip or bgzip reader interface. If you want to enlarge the buffer to improve performance, you can call scanner.Buffer() to set the byte slice and max size. The default scanner breaks on lines so you could either accumulate lines or hand off individual lines to the parser. Did you try this approach and find it was problematic?

wingolab (commented):
Did you try using bufio.Scanner for reading the annotation file? I'd suggest giving that a try with or without adjusting the buffer size to suit your desired size to see if serves your needs.

go/decompress/decompress.go
```diff
@@ -0,0 +1,205 @@
...
+	Reader *bgzf.Reader
+}
+
+type BufioBystroReader struct {
+	Reader *bufio.Reader
+}
+
+// Read a line up to the next newline character, and return the line excluding the newline character
+// Implementation follows the example in the bgzf package documentation: https://github.com/biogo/hts/blob/bb1e21d1bfc7f2b1e124ca0a1ed98493d191db78/bgzf/line_example_test.go#L70
+func _readLineBgzip(r *bgzf.Reader) ([]byte, error) {
```
wingolab, line 34: If you're aiming for idomatic go I'd 1) drop the underscore from the function name and 2) add a comment that begins // readLineBgzip takes a reader and returns a byte slice that ends in a new line or an error.

go/decompress/decompress.go
```diff
@@ -0,0 +1,205 @@
...
+
+		bufioReader := bufio.NewReader(b)
+
+		return &BufioBystroReader{Reader: bufioReader}, fileStats, nil
+	}
+
+	return &BzfBystroReader{Reader: b}, fileStats, err
+}
+
+func _getAnnotationFhFromTarArchive(archive *os.File) (*tar.Reader, fs.FileInfo, error) {
```
wingolab, line 181: Same issue with underscore as the first comment. Also, I'd suggest slightly refactoring to filehandleFromTarArchive so the comment would be // filehandleFromTarArchive takes a pointer to a tar archive and a filename and returns a desired filehandle from the tar archive.

go/decompress/decompress.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		if b == '\n' {
+			break
+		}
+		data = append(data, b)
+	}
+	return data, err
+}
+
+func ReadLines(r *bufio.Reader) ([]byte, error) {
```
wingolab, line 99: Can you please add a test for this function? To be safe, testing with multiple threads seems like a good idea.

wingolab (dismissed):
The parse portion looks fine. Would add a test for your ReadLines func given its central role.

akotlar:
> Did you try using bufio.Scanner for reading the annotation file? I'd suggest giving that a try with or without adjusting the buffer size to suit your desired size to see if serves your needs.

I haven't, I followed the bgzf library's approach directly. I've created a ticket to evaluate bufio.Scanner use: https://github.com/bystrogenomics/bystro/issues/366

go/cmd/opensearch/main.go
akotlar: I've created a ticket to evaluate this approach. bgzf's use of transactions would I believe require me to wrap the scanner call in one long transaction, but I don't know how problematic that is.

wingolab (commented):
Hi Alex, my earlier concern was that readLines seemed like it could fail in some subtle ways. Here are some tests that illustrate the issues.

```go


// NOTE: test data comes from `../test/opensearch/testdata/input.txt`
func Test_readLinesWithBuffer(t *testing.T) {
	type args struct {
		bufferSize int
		input      []byte
	}
	type want struct {
		file   string
		golden []byte
	}
	tests := []struct {
		name    string
		args    args
		want    []byte
		wantErr bool
	}{
		{
			name: "empty input",
			args: args{
				bufferSize: 100,
				input:      []byte(""),
			},
			want:    []byte(""),
			wantErr: false,
		},
		{
			name: "negative buffer",
			args: args{
				bufferSize: -1,
				input:      []byte(""),
			},
			want:    []byte(""),
			wantErr: true,
		},
		{
			name: "ends with \\n",
			args: args{
				bufferSize: 100,
				input:      []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5\n"),
			},
			want:    []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
			wantErr: false,
		},
		{
			name: "ends without \\n",
			args: args{
				bufferSize: 100,
				input:      []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
			},
			want:    []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
			wantErr: false,
		},
		{
			name: "add blank lines after first line",
			args: args{
				bufferSize: 100,
				input:      []byte("a\tb.c\td.e.f\n\n\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5\n"),
			},
			want:    []byte("a\tb.c\td.e.f\n\n\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
			wantErr: false,
		},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			b := bytes.NewReader(tt.args.input)
			r := bufio.NewReader(b)
			got, err := readLinesWithBuffer(r, tt.args.bufferSize)
			if (err != nil) != tt.wantErr {
				t.Errorf("readLinesWithBuffer() error = %v, wantErr %v", err, tt.wantErr)
				return
			}
			if !reflect.DeepEqual(got, tt.want) {
				t.Errorf("readLinesWithBuffer() = %v, want %v", got, tt.want)
			}
		})
	}
}
```

akotlar:
> Hi Alex, my earlier concern was that readLines seemed like it could fail in some subtle ways. Here are some tests that illustrate the issues.
> 
> ```go
> // NOTE: test data comes from `../test/opensearch/testdata/input.txt`
> func Test_readLinesWithBuffer(t *testing.T) {
> 	type args struct {
> 		bufferSize int
> 		input      []byte
> 	}
> 	type want struct {
> 		file   string
> 		golden []byte
> 	}
> 	tests := []struct {
> 		name    string
> 		args    args
> 		want    []byte
> 		wantErr bool
> 	}{
> 		{
> 			name: "empty input",
> 			args: args{
> 				bufferSize: 100,
> 				input:      []byte(""),
> 			},
> 			want:    []byte(""),
> 			wantErr: false,
> 		},
> 		{
> 			name: "negative buffer",
> 			args: args{
> 				bufferSize: -1,
> 				input:      []byte(""),
> 			},
> 			want:    []byte(""),
> 			wantErr: true,
> 		},
> 		{
> 			name: "ends with \\n",
> 			args: args{
> 				bufferSize: 100,
> 				input:      []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5\n"),
> 			},
> 			want:    []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
> 			wantErr: false,
> 		},
> 		{
> 			name: "ends without \\n",
> 			args: args{
> 				bufferSize: 100,
> 				input:      []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
> 			},
> 			want:    []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
> 			wantErr: false,
> 		},
> 		{
> 			name: "add blank lines after first line",
> 			args: args{
> 				bufferSize: 100,
> 				input:      []byte("a\tb.c\td.e.f\n\n\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5\n"),
> 			},
> 			want:    []byte("a\tb.c\td.e.f\n\n\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
> 			wantErr: false,
> 		},
> 	}
> 	for _, tt := range tests {
> 		t.Run(tt.name, func(t *testing.T) {
> 			b := bytes.NewReader(tt.args.input)
> 			r := bufio.NewReader(b)
> 			got, err := readLinesWithBuffer(r, tt.args.bufferSize)
> 			if (err != nil) != tt.wantErr {
> 				t.Errorf("readLinesWithBuffer() error = %v, wantErr %v", err, tt.wantErr)
> 				return
> 			}
> 			if !reflect.DeepEqual(got, tt.want) {
> 				t.Errorf("readLinesWithBuffer() = %v, want %v", got, tt.want)
> 			}
> 		})
> 	}
> }
> ```

I added your test case. After fixing disallowing negative buffers (will default to DEFAULT_BUFFER_SIZE for buffer sizes of <=0, the test cases work.

go/decompress/decompress.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+		data = append(data, b)
+	}
+	return data, err
+}
+
+func readLinesWithBuffer(r *bufio.Reader, bufferSize int) ([]byte, error) {
+	if bufferSize <= 0 {
+		bufferSize = DEFAULT_BUFFER_SIZE
+	}
```
wingolab, line 109: I would vote for returning an error if someone passed a buffer that is at or below a certain threshold (or zero at a minimum).

go/decompress/decompress_test.go
```diff
@@ -0,0 +1,349 @@
...
+			want:    []byte(""),
+			wantErr: io.EOF,
+		},
+		{
+			name: "ends with \\n",
+			args: args{
+				bufferSize: 100,
+				input:      []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5\n"),
+			},
+			want:    []byte("a\tb.c\td.e.f\n1\tA\t1;2|3;4/5\n1|2\tA\t1;2|3;4/5\na/2\tA;B\t1/2|3;4/5"),
```
wingolab, line 305: I forgot to mention that the input and expected output don't match in this case (because I was playing around with it and I wasn't sure if this was your goal).  I would have expected the output has a final \n but it doesn't. Can you check on this?

go/decompress/decompress_test.go
```diff
@@ -0,0 +1,349 @@
...
+			},
+			want:    []byte(""),
+			wantErr: io.EOF,
+		},
+		{
+			name: "negative buffer",
+			args: args{
+				bufferSize: -1,
+				input:      []byte(""),
+			},
```
wingolab, line 295: the error hear is a bit odd since it really results from the input having nothing to read rather than the error with the buffer size. if the data were a positive length byte slice then it will error by not equaling the expected result.

wingolab (commented):
sorry, two more comments.

go/decompress/decompress_test.go
akotlar: The read functions always trim the trailing new line. I’ll add a comment above the interface explaining that.

go/decompress/decompress_test.go
akotlar: So, I am now using io.ReadFull in the bufio/bgzip path, which will guarantee reading N bytes (bufferSize), or, if the buffer is longer than the remainder of the file, io.ReadFull will raise io.ErrUnexpectedEOF, and bgzf.Read will raise just io.EOF

wingolab (approved):

PR #360, opened 2023-12-11
Add PCs to ancestry response for biplots

cristinaetrv:
Take existing PCs generated during ancestry inference and make them available in the response


akotlar (approved):
lgtm

PR #362, opened 2023-12-17
PPCA Missing Data

austinTalbot7241993:
**Description:**

This pull request introduces a version of probabilistic Principal Component Analysis (PCA) designed to handle covariates that are missing completely at random, a common occurrence in certain datasets. The implementation is found in the gf_generative_missing_pt file. In addition to the main implementation, comprehensive unit tests have been added to validate the functionality and robustness of the code.

**Changes Made:**

**Probabilistic PCA Enhancement**: Implemented modifications to probabilistic PCA to accommodate covariates missing completely at random.

**New File**: Added the gf_generative_missing_pt file to house the updated probabilistic PCA implementation.

**Unit Tests**: Included unit tests associated with the new implementation to ensure correctness and reliability.


austinTalbot7241993:
Don't know what to do with that ruff error, it doesn't recognize that I want an element-wise comparison with numpy (requiring ==) rather than testing if the object is False.

python/python/bystro/supervised_ppca/_misc_np.py
```diff
@@ -59,3 +59,25 @@ def subset_square_matrix_np(Sigma, idxs):
     """
     Sigma_sub = Sigma[np.ix_(idxs == 1, idxs == 1)]
     return Sigma_sub
+
+
+def classify_missingness(matrix):
```
akotlar, line 64: This incurs O(N) lookups in unique patterns, and performs 2 iterations over pattern_indcies, one for matrices_list, one for vectors_list. Let's simplify and improve performance:

```python
import numpy as np

def classify_missingness(matrix):
    pattern_dict = {}
    matrices_list = []
    vectors_list = []

    for i, row in enumerate(matrix):
        # Using a hashable representation of the NaN pattern
        pattern = tuple(np.isnan(row))

        if pattern not in pattern_dict:
            pattern_dict[pattern] = [i]
        else:
            pattern_dict[pattern].append(i)

    for indices in pattern_dict.values():
        sub_matrix = matrix[indices]
        matrices_list.append(sub_matrix)
        vectors_list.append(~np.isnan(sub_matrix[0]))

    return matrices_list, vectors_list
```

akotlar: https://replit.com/join/thsiluytwj-akotlar demonstrating equivalence

python/python/bystro/supervised_ppca/gf_generative_missing_pt.py
```diff
@@ -0,0 +1,326 @@
...
+            trainable_variables,
+            lr=training_options["learning_rate"],
+            momentum=training_options["momentum"],
+        )
+        eye = torch.tensor(np.eye(p).astype(np.float32))
+        softplus = nn.Softplus()
+
+        _prior = self._create_prior()
+
+        for i in trange(
```
akotlar, line 124: Let's optimize this loop a bit

```python
for i in trange(
    training_options["n_iterations"], disable=not progress_bar
):
    sigma = softplus(sigmal_)
    WWT = torch.matmul(torch.transpose(W_, 0, 1), W_)
    Sigma = WWT + sigma * eye

   like_marginal = []
   for y in range(n_groups):
      sigma = Sigma[miss_pat[y]][:, miss_pat[y]]
      mvn = MultivariateNormal(torch.zeros(p_list[y]), sigma)
      like_marginal.append( torch.sum(mvn.log_prob(Xt_list[y])) )
   
    like_tot = torch.sum(torch.stack(like_marg)) / N
    like_prior = _prior(trainable_variables)
    posterior = like_tot + like_prior / N
    loss = -1 * posterior

    optimizer.zero_grad()
    loss.backward()
    optimizer.step()

    self._save_losses(i, like_tot, like_prior, posterior)
```

akotlar (dismissed):
Requested small optimizations, provided implementations

python/python/bystro/supervised_ppca/_misc_np.py
austinTalbot7241993: Compared to the training loop the cost of this is O(0). But if you want it optimized I will.

python/python/bystro/supervised_ppca/gf_generative_missing_pt.py
akotlar: Instead of having 3 loops over n_groups, combine them into 1

python/python/bystro/supervised_ppca/_misc_np.py
akotlar: Yeah, thanks! I gave the implementation, if you use that directly should work fine.

python/python/bystro/supervised_ppca/gf_generative_missing_pt.py
akotlar: There is probably a vectorized approach as well:

‘’’py

# Assuming Sigma, Xt_list, p_list, and N are already defined

# Step 1: Create batched mean and covariance matrices
batch_means = torch.stack([torch.zeros(p_list[y]) for y in range(n_groups)])
batch_covariances = torch.stack([Sigma[miss_pat[y]][:, miss_pat[y]] for y in range(n_groups)])

# Step 2: Create a batch of multivariate normal distributions
batch_mvn = MultivariateNormal(batch_means, batch_covariances)

# Step 3: Calculate log probabilities for each group's data in a vectorized way
log_probs = batch_mvn.log_prob(torch.stack(Xt_list))

# Step 4: Sum up these log probabilities and divide by N
like_tot = torch.sum(log_probs) / N

‘’’

akotlar (approved):
LGTM

PR #363, opened 2023-12-18
Feature/ancestry mixture ppca

austinTalbot7241993:
For some reason kept pushing to local branch


python/python/bystro/ancestry/gmm_ancestry.py
```diff
@@ -0,0 +1,436 @@
...
+            "momentum": 0.9,
+        }
+        tops = {**default_options, **training_options}
+
+        default_keys = set(default_options.keys())
+        final_keys = set(tops.keys())
+
+        expected_but_missing_keys = default_keys - final_keys
+        unexpected_but_present_keys = final_keys - default_keys
+        if expected_but_missing_keys:
```
akotlar, line 116: I don't believe this condition is possible: https://replit.com/@akotlar/CooperativeIckyWireframe#main.py

python/python/bystro/ancestry/gmm_ancestry.py
```diff
@@ -0,0 +1,436 @@
...
+        """
+        Fills in options for prior parameters on latent space
+
+        Paramters
+        ---------
+        new_dict : dictionary
+            The prior parameters used to specify the prior
+        """
+        default_dict = {"mu_l2": 1.0}
+        new_dict = {**default_dict, **prior_options}
```
akotlar, line 132: Do we need to raise a value error for expected priors?

austinTalbot7241993: I'll come up with better priors later. Immediate task will be creating a more meaningful unit test and making sure it fits well.

akotlar (approved):
LGTM, an important step forward

PR #367, opened 2024-01-10
Adversarial Principal Component Analysis

austinTalbot7241993:
We can use this model to make the principal components orthogonal to sensitive variables such as race or sex.


python/python/bystro/supervised_ppca/gf_adversarial_pt.py
```diff
@@ -0,0 +1,318 @@
...
+
+        Parameters
+        ----------
+        X : NDArray,(n_samples,p)
+            The data
+
+        Returns
+        -------
+        W_ : torch.tensor,shape=(n_components,p)
+            The loadings of our latent factor model
```
akotlar, line 290: Do we want to keep documentation specific to factor models?

python/python/bystro/supervised_ppca/gf_adversarial_pt.py
```diff
@@ -0,0 +1,318 @@
...
+    n_components : int
+        The latent dimensionality
+
+    prior_options : dict
+        The hyperparameters for the prior on the parameters
+
+    mu : float>0,default=1.0
+        The adversarial penalty
+
+    gamma : float,default=10.0
```
akotlar, line 91: missing the training options parameter description

akotlar (commented):
very minor comments

python/python/bystro/supervised_ppca/gf_adversarial_pt.py
austinTalbot7241993: What do you mean?

Issue #376, opened 2024-01-29
Don't archive bystro annotation outputs by default


akotlar:
###### A concise description of why we're pursuing this task.
Currently the bystro annotator places all of its outputs in a tar archive. It was once the case that the only module that outputted data was the annotator. That is no longer the case, and reading from a tar archive is less ergonomic than reading the files of interest directly

###### What is the goal, and how do we know this task is accomplished? 
1. The Bystro annotator server no longer archives job outputs. 
2. Update the Python and Go libraries to read their dependencies (annotation output) directly, rather than through tar.

###### Give enough technical detail for a teammate to understand the proposed solution.
perl/bin/bystro-server.pl needs to not set archive => 1
The go indexer needs to read the annotation file directly (and will need therefore to take the annotation file name as an input).
The saving function should not tar the archive

akotlar:
Done, merged. Ok to close

PR #377, opened 2024-01-29
Issue #376: Remove dependence on annotation tarball

akotlar:
* Removes dependence on annotation tarball, in favor of directly accessing all files

This change is necessary to simplify how we extract data from bystro outputs, which are currently split between the annotation tarball, which Bystro Annotator outputs, and all other outputs, adding complexity. Additionally, it is unnecessarily complicated to read Bystro Annotator data, like the dosage matrix output, as a result of the tarring process.

This corresponds to a PR in Bystro-web that switches from downloading a pre-computed annotation tarball, to dynamically zipping outputs at download time.


akotlar:
No squash

python/python/bystro/supervised_ppca/tests/test_utils_pt.py
```diff
@@ -1,4 +1,7 @@
+import random
 import torch
+import numpy as np
```
akotlar, line 3: The changes in this file (setting seeds) are necessary to make the test consistently pass in GitHub actions.

perl/bin/bystro-server.pl
```diff
@@ -279,7 +279,8 @@ sub coerceInputs {
     compress       => 1,
     archive        => 1,
```
cristinaetrv, line 280: I think you forgot to get rid of this archive here since you specify archive => 0 in your new code on line 283. 

Also, on line 267 commonArgs has an extra m but if that's intentional since it seems to be like that in other spots too just me know.

python/python/bystro/search/index/listener.py
```diff
@@ -80,8 +80,8 @@ def run_handler_with_config(
         mapping_config,
         "--opensearch-config",
         opensearch_config,
-        "--tarball-path",
-        tar_path,
+        "--input",
```
cristinaetrv, line 83: Just a clarification question - if the tarball-path is the output, shouldn't this be --output instead of --input? Or is this the vcf path that's the input and it's also the same path for the annotation output?

akotlar: Ah, good question, it is confusing; here we're calling the bystro opensearch program, which takes an annotation as an input. Previously it took the tarball as input (since the tarball contained the annotation)

perl/bin/bystro-server.pl
akotlar: oh good catch!

python/python/bystro/search/index/listener.py
cristinaetrv: Ok got it!

cristinaetrv (approved):
Lgtm!

PR #379, opened 2024-01-30
Add better squash instructions

akotlar:


CONTRIBUTING.md
```diff
@@ -23,7 +23,7 @@ git checkout -b feature/some_feature
 git commit -a -m "Addresses issue #NNN".
 
 # Push to the remote repository, explicitly telling git that the remote repository you want to push to is "origin" (the foobar fork).
-git push --set-upstream origin feature/some_feature
+git push --set-upstream origin feature/999
```
cristinaetrv, line 26: This no longer matches the example on line 19 - do you want it to say 999 up there as well?

CONTRIBUTING.md
```diff
@@ -74,35 +74,63 @@ git pull upstream refs/pull/999/head
 # Pull Request Etiquette
```
cristinaetrv, line 74: I think this section should go before 'Reviewing pull requests.' It's a little bit confusing to follow making another branch called pr_999 and then going back to the feature/999 one. Having everything grouped for what to do for one's own branch and then for someone else's work I think would be easier to follow.

CONTRIBUTING.md
```diff
@@ -74,35 +74,63 @@ git pull upstream refs/pull/999/head
...
-  git remote --verbose
-  
   # Overwrite your fork's remote history to mirror the local history (with the past 7 commits condensed into 1).
   # With --force-with-lease, the push will be rejected if there are new commits on the remote branch that you have not pulled.
-  git push --force-with-lease origin feature/999
+  git push --force-with-lease
   ```
 
- - This will ensure that git history remains clean, while still allowing you to commit frequently during development.
+  If you have already created a branch from feature/999, which was just squashed, you can bring a child branch (feature/1000) up to date by:
```
cristinaetrv, line 107: I think I would also add another subheading here, to indicate this section is just about child branches. I think the only part that really relates to etiquette is really lines 75 and 76, so maybe a subheading to indicate the different procedures would be useful. Like if line 79 had a subheading with something like 'How to maintain 1 commit per PR' for instance.

CONTRIBUTING.md
```diff
@@ -74,35 +74,63 @@ git pull upstream refs/pull/999/head
...
+  git reset --soft feature/999
+
+  # Commit all of the changes, thereby squashing them
+  git commit -a -m "some commit message for the commits exclusive to feature/1000 that have been squashed"
+
+  # Push to the feature branch
+  git push --force-with-lease
+  ```
 
+This will ensure that git history remains clean, while still allowing you to commit frequently during development.
```
cristinaetrv, line 136: I think I would add some kind of wrap up statement at the end. Maybe add some kind of contact information if they have questions, that sort of thing.

CONTRIBUTING.md
```diff
@@ -74,35 +74,63 @@ git pull upstream refs/pull/999/head
...
+  # bring feature/1000 up to date with feature/999
+  git rebase --fork-point feature/999
 
+  # push the changes to feature/1000 to remote
+  git push --force-with-lease
+  ```
 
+  If you now want to squash the commits that are exclusive to feature/1000:
+  ```sh
+  # Checkout the child branch (feature/1000), and bring it up to date with the changes in the parent branch (feature/999)
```
cristinaetrv, line 121: You might want to mention or clarify that squashing of the parent is not required before this part, worded something like: Whenever you want to bring the child branch up to date

And that you'll have to resolve conflicts and commit it before continuing the procedure.

PR #287, opened 2023-10-12
Set up for standing up PRS worker

cristinaetrv:
Provides basic functionality for PRS worker and describes basic framework for PRS module.

PR #381, opened 2024-01-30
Issue 378: PRS Listener Scaffold

akotlar:
* Creates PRS module, with beanstalkd listener and handler functions. Includes test demonstrating 2 ways to read the input dosage matrix. The preferred way is through Arrow, as that will be more memory efficient

Stacked on https://github.com/bystrogenomics/bystro/pull/377. Review that first.


python/python/bystro/prs/tests/test_prs_dosage.py
```diff
@@ -0,0 +1,277 @@
...
+            weighted_column = pc.multiply(table.column(column), pa.array(weights))
+            prs_scores[column] = pc.sum(weighted_column).as_py()
+
+    assert len(prs_scores) == len(table.column_names) - 1  # excluding 'locus' column
+    print(prs_scores)
+
+
+def calculate_prs_in_pandas(file_path, weights):
+    """
+    Calculate PRS scores efficiently, in batches
```
cristinaetrv, line 113: Since we decided we're doing PRS scores by sample rather than in batches, do we still need this test or do you want to leave it as an example?

cristinaetrv:
Should we go ahead and close #287 with this one, and just add the handler function on to this as a base? Or should we just removed the prs_listener.py from that PR and keep everything else in there? Basically, how should we handle the overlap?

python/python/bystro/prs/tests/test_prs_dosage.py
```diff
@@ -0,0 +1,277 @@
...
+
+            weights_subset = weights.loc[sample_genotypes.index]
+            prs_scores_batch = sample_genotypes.multiply(weights_subset, axis="index").sum()
+
+            if sample in prs_scores:
+                prs_scores[sample] += prs_scores_batch
+            else:
+                prs_scores[sample] = prs_scores_batch
+
+    print(prs_scores)
```
cristinaetrv, line 277: Will all of these tests that are performance tests and examples generating dummy prs_scores be running every time? Shouldn't we just do these kinds of tests one time and then decide the best method? Just want some clarification on the purpose of the tests here/whether they're temporary while we're deciding on the design details of this module since they might kind of visually clog up the test suite outputs versus just having unit tests or integration tests in the codebase.

python/python/bystro/prs/tests/test_prs_dosage.py
```diff
@@ -0,0 +1,277 @@
...
+            else:
+                prs_scores[sample] = prs_scores_batch[sample]
+
+    return prs_scores
+
+
+def test_prs_one_batch(create_feather_file):
+    """
+    Test PRS calculations with Pandas
+    This will be less memory efficient than operating on the Feather file directly
```
cristinaetrv, line 198: If you want to include some kind of performance testing here, I think it would be useful to have a performance metric as the output vs printing the scores if the scores are going to be the same between methods that you're testing.

python/python/bystro/prs/tests/test_prs_dosage.py
cristinaetrv: Follow up question, can you batch by samples rather than by loci?

python/python/bystro/prs/tests/test_prs_dosage.py
akotlar: >  if the scores are going to be the same between methods that you're testing.

We need to test that is the case, which is why we do `assert prs_scores == prs_scores_pandas`, because it is not guaranteed that we handle batching correctly. 

I can drop all print statements. There is a better way of handling benchmarking, and I can add that in.

python/python/bystro/prs/tests/test_prs_dosage.py
akotlar: These tests cumulatively take 

```py
================================================================== 7 passed in 0.73s ==================================================================

real    0m1.140s
user    0m1.172s
sys     0m0.280s
```

to run

Which ones do you want me to remove?

Also pytest never outputs the print messages. You have to explicitly ask for them.

akotlar: I've cleaned up the tests, removing print statements, and doing a better job demonstrating the equivalence of different approaches.

akotlar: I'll also mention that a large benefit of a testing suite is as documentation. Here we are demonstrating different ways that someone can use our output format to generate PRS scores

cristinaetrv: Ok gotcha, didn't realize those were optional outputs. Mostly just wanted a clarification on general purpose of tests wrt which tests we want to have in the codebase vs running in advance of PRing to figure out a design choice outside of the codebase.

akotlar:
> Should we go ahead and close #287 with this one, and just add the handler function on to this as a base? Or should we just removed the prs_listener.py from that PR and keep everything else in there? Basically, how should we handle the overlap?

Let's close it, and port anything over that's needed?

python/python/bystro/prs/tests/test_prs_dosage.py
akotlar: Added the benchmark and clarified the tests

python/python/bystro/prs/tests/test_prs_dosage.py
akotlar: Yep! (you can grab [sample1, sample2, sample3] instead of [sample1])

cristinaetrv (approved):
Lgtm, with a reminder to @cristinaetrv to add a subsequent PR that ports over types and changes to config and startup files from #287 (Set up for standing up PRS worker), which we are closing to avoid stale design choices creating overlap issues with this PR.

PR #385, opened 2024-01-31
Extracted and Updated proteomics

dlin30:
- Extracted and moved Proteomics cli library to the api folder
- Update proteomic api endpoint
- Added validation for annotation job
- Added Linkage between a proteomic job and a pre-existing annotation job

WIP - need to ask aboutt parentjob object that should be sent with the proteomics job


python/python/bystro/api/cli.py
```diff
@@ -280,7 +280,7 @@ def login(args: argparse.Namespace, print_result=True) -> CachedAuth:
     return state
 
 
-def authenticate(args) -> tuple[CachedAuth, dict]:
+def authenticate(dir_path: str) -> tuple[CachedAuth, dict]:
```
akotlar, line 283: Let's name dir_path something a bit more idiomatic (bystro_token_path or something like that), and set a default (~/.bystro).

THen, let's make sure the bystro_token_path exists or make it. the simplest way is to os.path.mkdir(bystro_token_path , exists_ok=True)

akotlar:
> WIP - need to ask aboutt parentjob object that should be sent with the proteomics job

We discussed this, no parent job needed

python/python/bystro/api/proteomics_cli.py
```diff
@@ -0,0 +1,243 @@
...
+class DatasetTypes(Enum):
+    fragpipe_TMT = "fragpipe-TMT"
+    somascan = "somascan"
+
+
+def upload_proteomics_dataset(
+    protein_abundance_file: str,
+    experiment_annotation_file: str = None,
+    annotation_job_id: str = None,
+    proteomics_dataset_type: str = "fragpipe_TMT",
```
akotlar, line 41: don't use the bare "fragpipe_TMT" string, instead DatasetTypes.fragpipe_TMT

python/python/bystro/api/cli.py
```diff
@@ -155,15 +155,15 @@ def load_state(state_dir: str) -> CachedAuth | None:
...
+def save_state(data: CachedAuth, bystro_token_path: str, print_result=True) -> None:
     """
     Saves the authentication state to a file.
 
     Parameters
     ----------
     data : CachedAuth
         The data to save.
-    state_dir : str
+    bystro_token_path : str
```
akotlar, line 166: let's rename bystro_token_dir, to better communicate that this isn't a path to the token, but the directory where we want to store bystro credentials

akotlar: please also make the default for this DEFAULT_DIR

python/python/bystro/api/cli.py
akotlar: Please make the default for this DEFAULT_DIR

python/python/bystro/api/proteomics_cli.py
```diff
@@ -0,0 +1,243 @@
...
+    Returns
+    -------
+    dict
+        A json response with annotationID and proteomicsID.
+    """
+
+    abundance_required_headers = ["Index", "NumberPSM", "ProteinID", "MaxPepProb", "ReferenceIntensity"]
+    with open(protein_abundance_file, "r") as file:
+        first_line = file.readline().strip()
+        if not all(header in first_line for header in abundance_required_headers):
```
akotlar, line 73: Let's make the check a little more flexible by making it case insensitive

python/python/bystro/api/proteomics_cli.py
```diff
@@ -0,0 +1,243 @@
...
+    state, auth_header = authenticate(user_dir)
+    url = state.url + UPLOAD_PROTEIN_ENDPOINT
+
+    if annotation_job_id:
+        annotation_url = state.url + GET_ANNOTATION.replace(":id", annotation_job_id)
+        annotation_response = requests.get(annotation_url, headers=auth_header)
+
+        if annotation_response.status_code != HTTP_STATUS_OK:
+            print(
+                f"The annotation with ID {annotation_job_id} does not exist"
```
akotlar, line 86: does not exist or you do not have permissions to access this annotation.

python/python/bystro/api/proteomics_cli.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class DatasetTypes(Enum):
+    fragpipe_TMT = "fragpipe-TMT"
+    somascan = "somascan"
+
+
+def upload_proteomics_dataset(
+    protein_abundance_file: str,
+    experiment_annotation_file: str = None,
```
akotlar, line 39: the type here needs to be : str | None

python/python/bystro/api/proteomics_cli.py
```diff
@@ -0,0 +1,243 @@
...
+    annotation_job_id: str = None,
```
akotlar, line 40: type is : str | None

python/python/bystro/api/proteomics_cli.py
```diff
@@ -0,0 +1,243 @@
...
+) -> dict[str, Any]:
+    """
+    Upload a fragpipe-TMT dataset through the /api/jobs/proteomics/ endpoint and 
+    update the annotation job.
+
+    Parameters
+    ----------
+    protein_abundance_file : str
+        Path to the protein abundance file.
+    experiment_annotation_file : str
```
akotlar, line 53: Update the types for the optional parameters (str | None)

akotlar:
Tests are still failing. Can we correct this tomorrow?

akotlar:
Instead of the current folder structure, lets restructure as follows:

|── api/                          # CLI-specific code
│   ├── __init__.py
│   ├── auth.py             (state management, login, logout)
│   └── proteomics.py   (the code you're breaking out)
├── cli/                          # CLI-specific code
│   ├── __init__.py
│   ├── main.py             # Entry point for the CLI, sets up the argument parser
│   └── auth.py

dlin30:
Let me know if this the correct structure that you had in mind or if it needs to be corrected.

python/python/bystro/cli/auth.py
```diff
@@ -0,0 +1,242 @@
...
+
+import requests
+
+from msgspec import Struct, json as mjson
+
+DEFAULT_DIR = os.path.expanduser("~/.bystro")
+STATE_FILE = "bystro_authentication_token.json"
+
+
+class SignupResponse(Struct, rename="camel"):
```
akotlar, line 12: This is duplicated from api/auth.py, let’s remove here

akotlar: Same for the other structs below. They should only exist in api/auth.py

python/python/bystro/cli/auth.py
```diff
@@ -0,0 +1,242 @@
...
+    path = os.path.join(bystro_credentials_dir, STATE_FILE)
+
+    if os.path.exists(path):
+        with open(path, "r", encoding="utf-8") as f:
+            return mjson.decode(f.read(), type=CachedAuth)
+
+    return None
+
+
+def save_state(data: CachedAuth, bystro_credentials_dir: str = DEFAULT_DIR, print_result=True) -> None:
```
akotlar, line 97: This is again duplicated. All duplicated code should be removed. The only functions here should be those that takes argsparse.Namespace, and call the api/auth.py function

python/python/bystro/cli/auth.py
```diff
@@ -0,0 +1,242 @@
...
+    with open(save_path, "w", encoding="utf-8") as f:
+        f.write(encoded_data)
+
+    if print_result:
+        print(
+            f"\nSaved auth credentials to {save_path}:\n{mjson.format(encoded_data, indent=4)}"
+        )
+
+
+def signup(args: argparse.Namespace, print_result=True) -> CachedAuth:
```
akotlar, line 129: The body of this and login need to call the api/auth.py code.

dlin30:
I've tried to figure out the auth pathing but it does not seem to generate bystro_authenciate_token when it is being imported from a different directory.

dlin30:
Restructuring has been cleaned/tested, and proteomics has been added into the cli commands. However, im stuck on a blocker with the test not creating the access token, which is causing the test_cli to fail.

python/python/bystro/api/auth.py
```diff
@@ -79,7 +79,7 @@ class UserProfile(Struct, rename="camel"):
         The accounts of the user.
     role : str
         The role of the user.
-    last_login : datetime.datetime
+    lastLogin : datetime.datetime
```
akotlar, line 82: This is not needed. rename=“camel” takes a lastLogin and stores the value on last_login, so that we don’t need camel case in Python.

python/python/bystro/api/auth.py
```diff
@@ -89,7 +89,7 @@ class UserProfile(Struct, rename="camel"):
     email: str
     accounts: list[str]
     role: str
-    last_login: datetime.datetime
+    lastLogin: datetime.datetime
```
akotlar, line 92: Again not needed

python/python/bystro/api/proteomics.py
```diff
@@ -34,13 +34,36 @@ class DatasetTypes(Enum):
...
+    ----------
+    protein_abundance_file : str
+        Path to the protein abundance file.
+    experiment_annotation_file : str | None
+        Path to the experiment annotation file.
+    experiment_annotation_file : str | None
+        ID of the job associated with the annotation dataset.
+    proteomics_dataset_type : DatasetTypes
+        Type of the proteomics dataset (we only support fragpipe-TMT currently).
+    bystro_credentials_dir : str
```
akotlar, line 57: Credentials dir no longer needed

python/python/bystro/api/auth.py
akotlar: One thing you can do is make last_login optional. It doesn’t exist immediately after signup; a better solution would be fix that in bystro-web however.

python/python/bystro/cli/cli.py
```diff
@@ -0,0 +1,314 @@
...
+    query_parser.add_argument("--size", default=10, type=int, help="How many records (default: 10)")
+    query_parser.add_argument(
+        "--from_",
+        default=0,
+        type=int,
+        help="The first record to return from the matching results. Used for pagination",
+    )
+    query_parser.add_argument("--job_id", required=True, type=str, help="The job id to query")
+    query_parser.set_defaults(func=query_cli)
+    proteomics_parser = subparsers.add_parser("proteomics", help="Upload a proteomics dataset")
```
akotlar, line 295: I think we talked about putting the proteomics subparser into a separate file, and importing that subparser here. Long term that will be more sustainable that placing everything in this file. See this answer in chatgpt

https://chat.openai.com/share/9ba20cbd-2b6b-4aea-ad60-78e3fa03e861

akotlar: “””
To achieve a CLI (Command Line Interface) setup where you have a main CLI tool (`mycli`) and a subparser for a tool (`tool1`) that lives in a physically different Python file, you can use Python's `argparse` module. The `argparse` module supports the creation of subparsers, which allows you to define commands and subcommands, making it suitable for your use case.

Here's a step-by-step approach to achieve this:

### Step 1: Define the Subparser in a Separate File

First, let's create the subparser for `tool1` in a separate file. Let's call this file `tool1.py`.

```python
# tool1.py
import argparse

def setup_tool1_subparser(subparsers):
    tool1_parser = subparsers.add_parser('tool1', help='Tool1 help')
    tool1_parser.add_argument('tool1cmd', help='Command for tool1')
    tool1_parser.add_argument('--tool1cmdarg1', help='Argument for tool1cmd')
    
    # You can add command function here
    tool1_parser.set_defaults(func=tool1_command)

def tool1_command(args):
    print(f"Executing tool1 with command {args.tool1cmd} and argument {args.tool1cmdarg1}")
```

### Step 2: Main CLI Tool Setup

Next, let's set up the main CLI tool that uses `argparse` to integrate the subparser defined above. We'll call this main file `mycli.py`.

```python
# mycli.py
import argparse
from tool1 import setup_tool1_subparser

def main():
    parser = argparse.ArgumentParser(description='MyCLI tool')
    subparsers = parser.add_subparsers(help='sub-command help')
    
    # Setup subparsers for different tools
    setup_tool1_subparser(subparsers)
    
    # Parse the arguments
    args = parser.parse_args()
    
    if hasattr(args, 'func'):
        args.func(args)
    else:
        # If no subcommands were used, show help
        parser.print_help()

if __name__ == '__main__':
    main()
```

### How It Works

- In `tool1.py`, we define a function `setup_tool1_subparser` that takes `subparsers` as an argument. This function adds a subparser for `tool1` and sets up its arguments.
- In `mycli.py`, we import the `setup_tool1_subparser` function from `tool1.py` and call it, passing the `subparsers` object to it. This way, `mycli` knows about the commands and arguments defined for `tool1`.
- The `tool1_command` function in `tool1.py` is designed to handle the execution logic when `tool1` and its associated commands and arguments are invoked from the CLI.

### Usage

With this setup, you can call `tool1` from your CLI as follows:

```bash
python mycli.py tool1 tool1cmd --tool1cmdarg1 value
```

This command will invoke the `tool1_command` function and print a message indicating that the command and argument have been received and processed.

This approach allows you to maintain a clean separation of concerns between your main CLI tool and its sub-tools, with each sub-tool potentially living in its own file.

akotlar (changes requested):
See comments

python/python/bystro/cli/tests/test_cli.py
```diff
@@ -5,28 +5,36 @@
...
 
+from bystro.cli.cli import (
+    login_cli,
+    signup_cli,
+    get_jobs_cli,
+    create_jobs_cli,
+    get_user_cli,
+) 
+
+DEFAULT_DIR = os.path.expanduser("~/.bystro")
```
akotlar, line 29: You don’t want this. Use the default fit and credentials path from api/auth.py

python/python/bystro/cli/tests/test_cli.py
```diff
@@ -104,23 +111,24 @@ def test_load_state_existing_file(mocker):
...
+    result = load_state()
     assert result is None
 
 
 def test_save_state(mocker):
     mock_open = mocker.patch("builtins.open", mocker.mock_open())
 
-    save_state(EXAMPLE_CACHED_AUTH, "./", print_result=False)
+    save_state(EXAMPLE_CACHED_AUTH, print_result=True)
+    expected_file_path = os.path.join(DEFAULT_DIR, "bystro_authentication_token.json")
```
akotlar, line 129: The expected path should be imported from api/auth.py

python/python/bystro/cli/tests/test_cli.py
```diff
@@ -166,77 +174,88 @@ def test_create_job(mocker):
...
-    response = get_jobs(args, print_result=False)
+    args = Namespace(type="incomplete", id=None, print_result=False)
+    response = get_jobs_cli(args)
     assert response == []
 
 
 def test_get_job(mocker):
     mocker.patch(
-        "bystro.api.cli.authenticate",
+        "bystro.api.auth.authenticate",
```
akotlar, line 258: You need to patch the cli import not the auth import

python/python/bystro/api/auth.py
akotlar: @dlin30 please change the docstring to match the python property (last_login)

python/python/bystro/api/proteomics.py
akotlar: @dlin30 the bystro_credentials_dir mentions need to be ripped out from docstrings

python/python/bystro/api/annotation.py
```diff
@@ -0,0 +1,232 @@
...
+    createdAt: datetime.datetime
+
+
+def get_jobs(job_type=None, job_id=None, print_result=True) -> list[JobBasicResponse] | dict:
+    """
+    Fetches the jobs for the given job type, or a single job if a job id is specified.
+
+    Parameters
+    ----------
+    bystro_credentials_dir : str
```
akotlar, line 45: bystro_credentials_dir needs to be removed

akotlar:
This is close. A few changes:

1. Update the way the CLI is invoked. "proteomics" should be a sub parser, with a single command "upload". The help message for the sub parser should be something like "Bystro proteomics data management". 
   - `bystro-api proteomics upload --arg1 --arg2` is how this should be invoked; later we will move auth and annotation behind dedicated sub parsers
2. There is a bug, the `annotation_response.status_code` block is incorrectly indented, causing an error when you run:

It should be (notice `if annotation_response.status_code != HTTP_STATUS_OK:` only runs if annotation_job_id and isintance(annotation_job_id, str)
```py
if annotation_job_id and isinstance(annotation_job_id, str):
        annotation_url = state.url + GET_ANNOTATION.replace(":id", annotation_job_id)
        annotation_response = requests.get(annotation_url, headers=auth_header)

        if annotation_response.status_code != HTTP_STATUS_OK:
            print(
                f"The annotation with ID {annotation_job_id} does not exist or you do not have "
                "permissions to access this annotation."
            )
            return {}
```
- Also, instead of `and isinstance(annotation_job_id, str)`, if the argument passed in is not correct, you should return a value error, though we should allow any number or string (we don't care what the ID is, we just know it isn't a collection, like dict, array, tuple, etc)

3. The current message upon success, when no annotation_id is provided is confusing:

```sh
Link established successfully.

{
    "annotationID": null,
    "proteomicsID": "65cee736a93a2e13bd5c3a13"
}
```

What link? Here annotationID is null, so no link has been established. I would instead omit this when annotation_id is not provided, and if provided make it clear what job (by name + id) was linked:

`Annotation {name} (id: {annotation_id}) was linked to this proteomics submission (_id: "65cee736a93a2e13bd5c3a13")`

python/python/bystro/cli/proteomics_cli.py
```diff
@@ -0,0 +1,70 @@
...
+
+    return upload_proteomics_dataset(
+        protein_abundance_file=args.protein_abundance_file,
+        experiment_annotation_file=args.experiment_annotation_file,
+        annotation_job_id=args.annotation_job_id,
+        proteomics_dataset_type=proteomics_dataset_type,
+        print_result=True,
+    )
+
+def add_proteomics_subparser(subparsers) -> None:
```
akotlar, line 48:
```suggestion

def add_proteomics_subparser(subparsers) -> None:
    proteomics_parser = subparsers.add_parser("proteomics", help="Bystro proteomics dataset operations")

    proteomics_subparser = proteomics_parser.add_subparsers(
        title="commands", dest="command", help="Commands for proteomics datasets"
    )

    # Define a default function for proteomics that shows help if no command is provided
    def proteomics_default_help(_args):
        # Capture the help message
        buffer = io.StringIO()
        proteomics_parser.print_help(file=buffer)
        help_message = buffer.getvalue()

        # Custom message
        custom_message = "Bystro proteomics dataset operations.\n\n"

        # Find the insertion point for the custom message
        insertion_point = help_message.find('options:')
        if insertion_point != -1:
            # Insert the custom message just before "options:"
            help_message = help_message[:insertion_point] + custom_message + help_message[insertion_point:]
        else:
            # If "options:" is not found, prepend the custom message
            help_message = custom_message + help_message

        print(help_message)
        proteomics_parser.exit()

    # Set the default function for the subparser
    proteomics_parser.set_defaults(func=proteomics_default_help)

    # Command 1: upload
    upload_parser = proteomics_subparser.add_parser("upload", help="Upload a proteomics dataset")

    upload_parser.add_argument(
        "--protein_abundance_file", required=True, help="Path to the protein abundance file"
    )
    upload_parser.add_argument(
        "--experiment_annotation_file", help="Path to the experiment annotation file"
    )
    upload_parser.add_argument(
        "--annotation_job_id", help="ID of the annotation job to link with the proteomics dataset"
    )
    upload_parser.add_argument(
        "--proteomics_dataset_type",
        choices=[e.name for e in DatasetTypes],
        default="fragpipe_TMT",
        help="Type of the proteomics dataset",
    )
    upload_parser.set_defaults(func=upload_proteomics_cli)

```

This give the following behavior:

```sh
(bystro) (base) ubuntu@ip-10-98-135-15:~/bystro/python$ bystro-api
Using default credentials path: /home/ubuntu/.bystro/bystro_authentication_token.json
usage: bystro-api [-h] {login,signup,get-user,create-annotation,get-jobs,query,proteomics} ...

Bystro CLI tool for making API calls.

options:
  -h, --help            show this help message and exit

commands:
  {login,signup,get-user,create-annotation,get-jobs,query,proteomics}
    login               Authenticate with the Bystro API
    signup              Sign up to Bystro
    get-user            Handle user operations
    create-annotation   Create an annotation
    get-jobs            Fetch one job or a list of jobs
    query               The OpenSearch query string query, e.g. (cadd: >= 20)
    proteomics          Bystro proteomics dataset operations
```

Calling just the proteomics parser:
```sh
(bystro) (base) ubuntu@ip-10-98-135-15:~/bystro/python$ bystro-api proteomics
usage: bystro-api proteomics [-h] {upload} ...

Bystro proteomics dataset operations.

options:
  -h, --help  show this help message and exit

commands:
  {upload}    Commands for proteomics datasets
    upload    Upload a proteomics dataset
```

as an aside, please remove `print(f"Using default credentials path: {CREDENTIALS_PATH}")` from api/auth.py

python/python/bystro/api/auth.py
```diff
@@ -0,0 +1,370 @@
...
+DEFAULT_DIR = os.path.expanduser("~/.bystro")
+STATE_FILE = "bystro_authentication_token.json"
+
+CREDENTIALS_PATH = os.path.join(DEFAULT_DIR, STATE_FILE)
+
+if not os.path.exists(DEFAULT_DIR):
+    os.makedirs(DEFAULT_DIR, exist_ok=True)
+    print(f"Created default credentials path: {CREDENTIALS_PATH}")
+
+print(f"Using default credentials path: {CREDENTIALS_PATH}")
```
akotlar, line 17: please remove

python/python/bystro/api/auth.py
```diff
@@ -0,0 +1,370 @@
...
+    )
+
+    if print_result:
+        print("\nSignup & authentication successful. You may now use the Bystro API!\n")
+
+    return state
+
+
+def login(
+        email: str,
```
akotlar, line 240: These arguments seems overly indented ( 8 spaces). If you run through VS code's Format Document (command + shift + P), using black, on my machine I get 4 spaces.

Same for `def signup`.

Not a huge deal, but we should be consistent. In api/annotate.py we have 4 lines indent for `def create_jobs`

python/python/bystro/api/auth.py
```diff
@@ -0,0 +1,370 @@
...
+    body = {"email": email, "password": password}
+
+    response = requests.post(url, data=body, timeout=30)
+
+    if response.status_code != 200:
+        raise RuntimeError(
+            f"Login failed with response status: {response.status_code}. Error: \n{response.text}\n"
+        )
+
+    print("response.text", response.text)
```
akotlar, line 287: this should only happen if print_results is True?

akotlar (changes requested):
See comments

akotlar (changes requested):
Needs the changes we discussed:

1. Change "Proteomics Upload response" -> "Proteomics Upload Response"
2. When you provide an experiment_annotation file, the outputBaseFileName should be the name of the protein abundance file, not the experiment file. Also when this happens you don't upload the abundance file anymore, just the experiment annotation.tsv. I've attached the screenshot.
3. The docstring for upload_proteomics_data has experiment_annotation_file instead of annotation_job_id (in both cli and api)
4. When you upload an experiment file, it doesn't seem like the experiment actually gets put into postgres
5. annotationID and proteomicsID are not recorded on the proteomics job and annotation job when linking annotation to proteomics

python/python/bystro/api/proteomics.py
```diff
@@ -0,0 +1,185 @@
...
+        experiment_file = _package_filename(experiment_annotation_file)
+
+        experiment_response = requests.post(
+            state.url + EXPERIMENT_ENDPOINT,
+            headers=auth_header,
+            files={"file": experiment_file[1]},
+            data={"experimentName": experiment_name},
+        )
+        if experiment_response.status_code != HTTP_STATUS_OK:
+            raise Exception(
```
akotlar, line 118: Please convert this to a RuntimeError, squash this PR, and then we can get it in.

python/python/bystro/api/proteomics.py
```diff
@@ -0,0 +1,185 @@
...
+
+    proteomics_response_data = response.json()
+    if print_result:
+        print("\nProteomics Upload Response:", json.dumps(proteomics_response_data, indent=4))
+
+    proteomics_job_id = proteomics_response_data.get("_id")
+    final_response = {"proteomicsID": proteomics_job_id}
+
+    if annotation_job_id:
+        update_annotation_payload = {"proteomicsID": proteomics_job_id}
```
akotlar, line 166: This doesn't seem to work. I would actually prefer to remove it for now, because I don't have a strong use case for the link in the reverse direction

python/python/bystro/api/proteomics.py
```diff
@@ -0,0 +1,185 @@
...
+        update_annotation_response = requests.patch(
+            annotation_url, headers=auth_header, json=update_annotation_payload
+        )
+
+        if update_annotation_response.status_code == HTTP_STATUS_OK:
```
akotlar, line 171: You forgot to check the case when the status was not OK, but that is ok, superseded by my comment above

python/python/bystro/api/proteomics.py
akotlar: Let's do this:

```py
    if annotation_job_id:
        final_response["annotationID"] = annotation_job_id
        if print_result:
            annotation_name = annotation_response.json().get("name", "Unknown Annotation")
            print(
                f'\nAnnotation "{annotation_name}" (id: {annotation_job_id}) '
                f'was linked to this proteomics submission (id: "{proteomics_job_id}").\n'
            )
```

This will skip patching the annotation job, and just print out what was linked after fetching the name

akotlar (approved):

akotlar:
LGTM!

akotlar (approved):
LGTM!

Issue #383, opened 2024-01-30
Switch Ancestry Module to msgspec


akotlar:
###### A concise description of why we're pursuing this task.
Currently we use msgspec, data loader, and attrs in Bystro. We need to have a single approach, and msgspec is the most performant of the three, and offers a practically identical feature set. Notably it is a bit more cumbersome to define runtime validation that happens at object instantiation time outside of instantiation the object during deserialization of a JSON message, but this is not a big limitation because 1) we don't really want to validate objects during instantiation from code, as the static type checker (mypy) we use will catch type errors during development, 2) not only does msgspec support custom __post__init__ checks, but also directly supports attrs validators should we ever need those.

As an advantage, we will simplify the codebase and remove a dependency.

###### What is the goal, and how do we know this task is accomplished? 
Remove attrs entirely from the Ancestry

###### Give enough technical detail for a teammate to understand the proposed solution.
Remove attrs from every use in the Ancestry module, replacing it with Struct. For cases where Attrs is used to define the expected JSON input or output, extend from BaseMessage, [SubmittedJobMessage, or CompletedJobMessage as appropriate](https://github.com/bystrogenomics/bystro/blob/11bec306d591eb6d026916d310b1485b749b5020/python/python/bystro/beanstalkd/messages.py#L17-L28)

Issue #382, opened 2024-01-30
Switch Ancestry module to using dosage matrix outputted by Bystro annotator


akotlar:
###### A concise description of why we're pursuing this task.
Currently, one of the biggest bottlenecks in the ancestry module is that the dosage matrix is re-generated from the VCF the user inputted originally to the annotator, using a naively written python module. This is slow and memory intensive. Since we now generate dosage matrices directly during annotation, we should instead use those: https://github.com/bystrogenomics/bystro/pull/375

###### What is the goal, and how do we know this task is accomplished? 
The Arrow IPC dosage matrix is taken in AncestryJobData and used for calculating ancestry scores, rather than re-processing the input VCF to generate the dosage matrix first.

###### Give enough technical detail for a teammate to understand the proposed solution.
1. [AncestrySubmission](https://github.com/bystrogenomics/bystro/blob/11bec306d591eb6d026916d310b1485b749b5020/python/python/bystro/ancestry/ancestry_types.py#L49) is updated to expect a dosage matrix path instead of a vcf path. Currently this will be a disk path, but in the future could be expanded to an S3 object path
2. The API endpoint that sends the AncestryJobData, which wraps AncestrySubmission is updated to send the dosage matrix path
3. Update the handler function to use the genotype dosage matrix directly: https://github.com/bystrogenomics/bystro/blob/11bec306d591eb6d026916d310b1485b749b5020/python/python/bystro/ancestry/listener.py#L83
4. Update [_load_vcf](https://github.com/bystrogenomics/bystro/blob/11bec306d591eb6d026916d310b1485b749b5020/python/python/bystro/ancestry/listener.py#L70) to use the arrow genotype dosage matrix. 
5. Update the [training routine](https://github.com/bystrogenomics/bystro/blob/11bec306d591eb6d026916d310b1485b749b5020/python/python/bystro/ancestry/train.py) to generate dosage matrix with bystro-vcf with --noOut --dosageMatrix <path to output>, and use that for training.

akotlar:
The PR that addresses this is linked and has been merged. Ok to close

PR #386, opened 2024-02-01
Issue #382 #383: Remove Attrs from Ancestry and Convert to Genotype Dosage Matrix

akotlar:
* Remove Attrs from Ancestry
* Switch to genotype dosage matrix

In testing, this roughly drops ancestry runtime from multiple minutes for a small dataset to a few seconds, by avoiding VCF conversion and using an efficient data format that supports random row and column seeking.

We also delete a large amount of code that stemmed from a misunderstanding of the guarantees that the msgspec/attrs provide. We do not need to duplicate those libraries' test coverage.

Stacked on https://github.com/bystrogenomics/bystro/pull/377.


python/python/bystro/ancestry/tests/test_ancestry_types.py
```diff
@@ -1,16 +1,13 @@
 """Test ancestry_types.py."""
+import re
```
akotlar, line 2: Most of these tests are not needed. The entire point using something like msgspec is to rely on the static type system that python offers, and which is enforced by mypy, at compile time. Previously, this was obvious: we had to add # type: ignore assertions everywhere to satisfy mypy, because it caught those errors.

So I've elected to keep test_SuperpopVector_rejects_missing_key and test_SuperpopVector_rejects_extra_key to prove that even if mypy didn't catch the errors, msgspec strict mode prevents unknown parameters from being passed in, like expected, and then only tested the unit validator, which cannot be enforced by mypy currently (there is a solution I'm exploring, but its outside the scope of this PR).

python/python/bystro/ancestry/inference.py
```diff
@@ -22,14 +25,13 @@
...
-@dataclass(frozen=True)
-class AncestryModel:
+class AncestryModel(Struct, frozen=True, forbid_unknown_fields=True):
     """Bundle together PCA and RFC models for bookkeeping purposes."""
 
     pca_loadings_df: pd.DataFrame
     rfc: RandomForestClassifier
 
-    def __post_init__(self) -> "AncestryModel":
+    def __post_init__(self) -> None:
```
akotlar, line 34: By dropping attrs, we dramatically simplify our codebase. It is trivially easy to add runtime checks through __post_init__, as you would in data class.

I left the runtime validations checks even though they are entirely pointless, because I was not far enough along to understand whether there was some way that Pat could have been generating or accepting invalid probabilities. I've learned that isn't the case.

The only place where you need runtime validation, is when you are taking user data. So I will in a followup PR remove all __post__init validation, in favor of using msgspec's "Constraints" (https://jcristharif.com/msgspec/constraints.html), which will perform the requisite checks when deserializing messages. Notably, since all of the ancestry types are send only (they are sent to the backend, but never currently serialized and sent to the ancestry module), these constraints will never run, but maybe one day we'll want another module to take ancestry data in JSON form and check that it is correct.

python/python/bystro/ancestry/ancestry_types.py
```diff
@@ -153,73 +82,60 @@ class SuperpopVector:
...
-    # needs to be literal float for msgspec
-    missingness: float = field(converter=float, validator=unit_float_validator)
+    sample_id: str
+    top_hit: AncestryTopHit
+    populations: PopulationVector
+    superpops: SuperpopVector
+    missingness: float
 
+    def __post_init__(self):
+        # Due to PEP 484, typescheckers do not distinguish between float and int
```
cristinaetrv, line 124: I think this is a small typo?
```suggestion
        # Due to PEP 484, type checkers do not distinguish between float and int
```

python/python/bystro/ancestry/inference.py
```diff
@@ -107,35 +104,47 @@ def _package_ancestry_response_from_pop_probs(
...
+        sample_missingnesses = genotypes.isna().mean(axis=0) + missingness
+
+        genotypes[genotypes.isna()] = 1
+
+        pcs_for_plotting, pop_probs_df = ancestry_model.predict_proba(genotypes)
+    logger.debug("Completed ancestry inference in %f seconds", timer.elapsed_time)
+
     return _package_ancestry_response_from_pop_probs(
-        vcf_path, pcs_for_plotting, pop_probs_df, missingnesses
+        pcs_for_plotting, pop_probs_df, sample_missingnesses
```
cristinaetrv, line 147: While I do prefer sample_missingness over missingness as the term used here, I think it would be helpful to include somewhere that this sample_missingness is relative to the trained model and not the measure of sample quality in any way, maybe not anywhere in the code necessarily but just wanted to mark that we should at least state this in the tutorial/info about the ancestry model somewhere.

python/python/bystro/ancestry/tests/test_listener.py
```diff
@@ -1,100 +1,38 @@
...
-    ancestry_submission = AncestrySubmission("foo.vcf")
-    ancestry_job_data = AncestryJobData(
-        submissionID="my_submission_id2", ancestry_submission=ancestry_submission
-    )
-    with patch("bystro.ancestry.listener._load_vcf", return_value=FAKE_GENOTYPES) as _mock:
-        ancestry_response = handler_fn(publisher, ancestry_job_data)
-    assert ancestry_submission.vcf_path == ancestry_response.vcf_path
-
-
-def test_submit_msg_fn_happy_path():
```
cristinaetrv, line 39: Consider adding this back

akotlar: Done, added tests for submit_msg_fn and completed_msg_fn

cristinaetrv (approved):
Lgtm!

Issue #388, opened 2024-02-02
[ancestry] Don't output results to beanstalkd


akotlar:
###### A concise description of why we're pursuing this task.
Currently we output the result of the ancestry module in the AncestryJobCompleteMessage as bytes. This is problematic, because the results of ancestry will be easily larger than 15MB, which is the Mongo document limit

###### What is the goal, and how do we know this task is accomplished? 
Stop writing the output as bytes to beanstalkd. Instead, write the output to the user's output directory, and return the path to the output.

###### Give enough technical detail for a teammate to understand the proposed solution.
1. Update AncestryJobCompleteMessage to have a result_path: str in place of the AncestryResult
2. Update AncestryJobData to take an outDir.
3. Write the AncesryResult to disk as json (e.g. serialize it). 
4. In bystro-web deserialize it as an AncestryResult using Zod.

akotlar:
This is complete in the linked issue. Ok to close

PR #389, opened 2024-02-02
Issue #388: Don't output ancestry to beanstalkd

akotlar:
* Write ancestry results to disk instead of sending the entire AncestryResult object through beanstalkd.

As currently architected, there is no way for Ancestry to scale beyond about 3000 samples, because 1) beanstalkd is not designed to send objects of 20+ MB (it can, but communication slows, clogs up queue), 2) mongodb can only store 16MB of data for the entire job, of which ancestry is just a part, 3) users almost certainly want to be able to download results.

In this PR we write to the user specified output directory, and return the name of the file written. We write a json file, uncompressed.


python/python/bystro/ancestry/dummy_ancestry_response.py
```diff
@@ -1,150 +0,0 @@
-"""Provide dummy ancestry response for integration testing."""
```
akotlar, line 1: This file was unused and redundant with the test_ancestry_listener.py examples on how to build an AncestryJobCompleteMessage

cristinaetrv:
General question - is this how we'll want to handle outputs for all analysis modules going forward since they'll all be similar sizes or larger?

cristinaetrv (approved):
Lgtm!

dlin30 (commented):
LGTM

Issue #392, opened 2024-02-03
Write annotation conf file to output dir


akotlar:
###### A concise description of why we're pursuing this task.
Currently we output the annotation configuration file directly to beanstalkd as a serialized value. These annotations can become very large, which both adds risk that the beanstlalkd message will be rejected, and that the mongodb (or postgres in the future) data store will exceed document size limits.

Additionally, by not writing the annotation configuration to file, the user cannot download it, only view it in the browser.

###### What is the goal, and how do we know this task is accomplished? 
Much like with ancestry, write the configuration output to disk.

###### Give enough technical detail for a teammate to understand the proposed solution.
Add the configuration as an output carried on outputFilesInfo in Seq::Definition, and write that output to disk after annotation succeeds in Seq.pm

akotlar:
This is completed. Ok to close.

PR #395, opened 2024-02-04
Issue #392: Write conf to output dir

akotlar:
* Don't output configuration objects directly to beanstalkd. Instead, write the configuration files to the user's output directory for reproducibility
* Continue threading through snake_case in python. 

This matches a corresponding change in Bystro-web that dynamically loads configurations, rather than relies on the data being passed around on the mongo object.

In a future PR I will completely remove camelCase from Python.


dlin30 (approved):
LGTM

PR #396, opened 2024-02-05
Local Ancestry Initial Implementation

austinTalbot7241993:
This implements a deep learning version of RF mix with substantially better scaling properties. In this case, most of the computation time scales independently of the number of segments considered. The model functions by training a neural network for each region of the genome, but all the layers after the first are completely shared. This means that we learn stochastically both by sample and by number of regions dramatically increasing computation speed.


akotlar:
Awesome, LGTM!

akotlar (approved):
Good to go; I reviewed a version of this code carefully already

Issue #393, opened 2024-02-03
Consistent bystro config naming scheme


wingolab:
###### A concise description of why we're pursuing this task.

The bystro configuration file uses different styles to name keys. It would make it easier if it was more consistent. I suggest lower case with underlines  (e.g., [snake_case](https://en.wikipedia.org/wiki/Snake_case)) but recognize this is arbitrary.

###### What is the goal, and how do we know this task is accomplished? 

- Update the bystro perl code to use lower case with underlines.
- Convert the existing example genome builds to the lower case with underlines convention.

###### Give enough technical detail for a teammate to understand the proposed solution.

Examine the code base - probably at the build steps - to determine the scope of change that this might entail.

akotlar:
For the perl codebase, the intention was to only use snake_case for CLI arguments, so converting entirely to camelCase will be far easier. Would this be workable?

akotlar:
Thinking about this some more I don't think that will be satisfactory. I'll convert the perl codebase to snake_case in a series of PRs; we'll do it gradually to avoid breaking too much.

wingolab:
The intention behind my suggestion was around the configuration file and any areas of the code that require adapting to a change from the existing name to one that uses lower case with underscores (e.g., snake_case). I didn't mean that we should change the entire Perl code base to use snake_case.

The bigger issue with the configuration files are their complexity, and what seems like some inconsistency between how track keys are used between different tracks (it was only one track's keys that behaved differently - not recalling it at the moment). This led me to look at configuration languages that 1) could enforce a schema; 2) allow the configuration to be split across several files for readability and maintenance ease; and, 3) would not make us change anything about the YAML and Perl code. I came across [pkl](https://github.com/apple/pkl) that might be something worth considering. It seems like it was just released but the tutorial and docs made me think it would be a few days of work to adapt for our project.

akotlar:
I like the suggestion and will look into pkl. I agree the configuration files are complex. Better documentation will help some as well.

PR #398, opened 2024-02-06
Issue #393: Consistent snake_case in python library

akotlar:
* Enforces snake_case in python. This auto-converts from camelcase, and will convert back to camelcase upon serialization. This way our JSON responses and inputs can follow JSON conventions , but also Python conventions.
* As part of this submissionID -> submissionId and queueID -> queueId, because this makes auto-camelization/un-camelization easier
* Fork liftover and fix liftover failure due to https://github.com/jeremymcrae/liftover/issues/19

### DIFF
--- a/perl/bin/bystro-server.pl
+++ b/perl/bin/bystro-server.pl
@@ -53,11 +53,11 @@ my $conf = LoadFile($queueConfigPath);
 
 # The properties that we accept from the worker caller
 my %requiredForAll = (
-  output_file_base => 'output_base_path',
+  output_file_base => 'outputBasePath',
   assembly         => 'assembly',
 );
 
-my $requiredForType = { input_file => 'input_file_path' };
+my $requiredForType = { input_file => 'inputFilePath' };
 
 say "Running Annotation queue server";
 
@@ -125,8 +125,8 @@ while ( my $job = $beanstalk->reserve ) {
         data     => encode_json(
           {
             event        => $STARTED,
-            submissionID => $jobDataHref->{submissionID},
-            queueID      => $job->id,
+            submissionId => $jobDataHref->{submissionId},
+            queueId      => $job->id,
           }
         )
       }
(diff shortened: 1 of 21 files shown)


perl/bin/bystro-server.pl
```diff
@@ -53,11 +53,11 @@
 
 # The properties that we accept from the worker caller
 my %requiredForAll = (
-  output_file_base => 'output_base_path',
```
akotlar, line 56: this is necessary because currently we don't use un-camelization in perl; I will introduce it later as we begin to convert perl to snake_case

python/python/bystro/search/utils/messages.py
```diff
@@ -8,7 +8,7 @@
 from bystro.search.save.binomial_maf import BinomialMafFilter
 
 
-class IndexJobData(BaseMessage, frozen=True, forbid_unknown_fields=True):
+class IndexJobData(BaseMessage, frozen=True, forbid_unknown_fields=True, kw_only=True, rename="camel"):
```
akotlar, line 11: kw_only because this one has many arguments

wingolab (approved):
This seems fine. I feel compelled to point out that there's no testing to show that this is performing as intended as far as I can tell. Since I've not used the python library I'm not in a position to know if this is an important step.

akotlar:
There is 1 failure on this branch. I will force merge because the liftover library appears to not be able to connect to UCSC servers (ERROR python/bystro/ancestry/tests/test_define_callset.py - ValueError: problem accessing https://hgdownload.cse.ucsc.edu/goldenPath/hg19/liftOver/hg19ToHg38.over.chain.gz). Nothing we can do about it at this time. If it turns out to not be a transient error, we will need to solve by bringing the chain files into the repo

akotlar:
The root cause of the liftover library issue is an address change: https://github.com/jeremymcrae/liftover/issues/19

I will make the test skipped until that is resolved or we pull in local liftover files

Issue #399, opened 2024-02-08
Filter dosage matrix during saving


akotlar:
###### A concise description of why we're pursuing this task.
We output an Arrow dosage matrix, which is used for PRS, ancestry. The search engine allows easy filtering of loci, and we have the ability to save the annotations corresponding to those queries. However currently, that functionality does not filter the dosage matrix, so we cannot calculate ancestry (or PRS) easily on the subset.

###### What is the goal, and how do we know this task is accomplished? 
Subset the dosage matrix during saving

###### Give enough technical detail for a teammate to understand the proposed solution.
Following the approach described in https://github.com/bystrogenomics/bystro/blob/master/python/python/bystro/prs/tests/test_prs_dosage.py#L177, within each Ray process, accumulate a list of loci (chr:pos:ref:alt), filter the dosage matrix, and write it to disk. The location bering written to should be a folder that is the name of the job.dosage. Within that folder, write each partition with the id of the worker process as the name name of the file. 

For instance, if the save job is named "foo" the filtered output will look like:
foo.dosage/
  0.feather
  1.feather
  2.feather

Arrow can read these partitioned datasets efficiently, and the solution as described will allow us to write this dataset in a multi-process way without blocking.

akotlar:
Done in the linked PR

PR #400, opened 2024-02-10
Issue #399 filter dosage matrix during save

akotlar:
* Writes dosage matrix during saving from query

Notice that this writes in chunks; an Arrow table written in multiple parts (partitions) is a valid Arrow dataset; e.g. you can read either a folder, or a file, with the equivalent code.

Ready to review

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,6 @@
     "editor.defaultFormatter": "ms-python.black-formatter"
   },
   "python.formatting.provider": "none",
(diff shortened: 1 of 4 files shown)


cristinaetrv (approved):
Tested in bystro-dev, seems to be working as expected. To keep in mind for future - pandas might not take in directory of feather files but 
`import pyarrow.dataset as ds; 
 ds.dataset('rarevarshg19.dosage.feather',format='arrow').to_table().to_pandas() `
works well for taking in directory of feather files as one file.

Lgtm!

Issue #402, opened 2024-02-11
Make saving much faster


akotlar:
###### A concise description of why we're pursuing this task.
Currently, saving queries is suboptimal: we download the opensearch JSON object, parse that back into a tsv format, and write to disk. Instead, much more efficient would be to simply get the document ids, which always correspond to the row indices of the annotation file (offset by 1, since the annotation file has a header), and copy the relevant lines from the parent job's annotation file. 

###### What is the goal, and how do we know this task is accomplished? 
A version of the save handler that 1) only requests during search the the chrom, pos, ref, alt fields (for locus filtering of the dosage matrix), 2) accumulates the document ids, sorts them, and uses them to filter the parent annotation file, copying those lines to the save job's output directory with the appropriate file name.

###### Give enough technical detail for a teammate to understand the proposed solution.
1) Use the Opensearch "fields" property, specifying chrom, pos, inputRef, alt, and set _source to False. This will ensure that we fetch only 4 fields, rather than the full document source, which could be massive (and pure waste)
2) Continue to perform queries and filter the dosage matrix in ray processes, but now also accumulate the document ids into a list, convert to an np array of ints (since Ray efficiently handles sharing data between processes when the data is np array)
3) Sort the document ids in each ray process before returning; this will make the final sort easier. 
4) In the master process, concatenate the arrays and sort them
5) Decompress the parent annotation file using bgzip with multiple threads, do this using a subprocess pipe. Write the wanted lines to another pipe, which is the bgzip compress command with multiple threads.For every line that is written to the bgzip compress command, also write to a 3rd pipe for bystro-stats, to ensure that we interleave writing the annotation file and calculating statistics on it

akotlar:
The work as described is done in the linked PR. Saving time drops to around 2 min for 167,000 variants (down from 10 min) and 3000 samples, and 30-40 min for 3 million variants and 2504 samples. Further improvements will come from randomly seeking lines either in bgzipped annotations or storing annotations as arrow IPC files. 

Memory usage is O(1) due to use of batching and streaming.

PR #403, opened 2024-02-12
Issue #402 fast saving

akotlar:
* Triples performance of saving queries on the same hardware. Saving a query of all records now takes as little time as annotation. This will improve further once we are storing the annotation as an arrow feather file
* Fixes issue with how the number of slices for scrolled search is generated; previously we assumed that Opensearch would return exactly the number of records requested per slice, except when n_slices was over provisioned, so that n_slices * query_size would yield >= the expected record count. It turns out that even if we round n_slices up ceil(n_documents / query_size), Opensearch may fail to fetch the expected number of records per slice, and the total number of fetched documents will be less than expected. To fix this, we assume that Opensearch may fetch as few as 85% of the requested query size, driving up the number of slices requested; empirically the true number appears to be about 3%, so this should give us sufficient buffer.

Stacked on https://github.com/bystrogenomics/bystro/pull/403/commits/bd8b27cb74a742515caa06bebd9b4f8d3922d4a1

First commit in this PR is: https://github.com/bystrogenomics/bystro/pull/403/commits/bd0c11ae2bc38116c7649cfc6fcf13f71c203841


python/python/bystro/ancestry/listener.py
```diff
@@ -90,7 +90,7 @@ def handler_fn(publisher: ProgressPublisher, job_data: AncestryJobData) -> Ances
         # simply threading it through for later.
         _reporter = get_progress_reporter(publisher)
 
-        dataset = ds.dataset([job_data.dosage_matrix_path], format="arrow")
+        dataset = ds.dataset(job_data.dosage_matrix_path, format="arrow")
```
akotlar, line 93: this is needed to support partitioned feather files (which are folders with arrow feather files in them that share the same schema)

cristinaetrv (approved):
Tested on bystro-dev with ~4.5 million variant ~2k sample file and everything seemed to work as expected afaict. Also requested '0 records written.' removed from logging step and that is working as expected as well.

Here are the times for saving a 41 variant subset of that base file:
Annotation: 00:04:24.430
Search index: 00:00:03.694

Here is the time recorded for saving a slightly smaller (~4.4 million variants) subset of that same base file (didn't finish indexing):
Annotation: 02:45:25.365

Ran into a memory problem when trying to save another subset (~1 million variants) of that same base file while the 4.4 million subset was still indexing (Error message: Task was killed due to the node running low on memory.), so this may be worth investigating in future iterations of the save module and we might want to figure out what our exact current limitations are in terms of saving until then.

akotlar:
> Tested on bystro-dev with ~4.5 million variant ~2k sample file and everything seemed to work as expected afaict. Also requested '0 records written.' removed from logging step and that is working as expected as well.
> 
> Here are the times for saving a 41 variant subset of that base file: Annotation: 00:04:24.430 Search index: 00:00:03.694
> 
> Here is the time recorded for saving a slightly smaller (~4.4 million variants) subset of that same base file (didn't finish indexing): Annotation: 02:45:25.365
> 
> Ran into a memory problem when trying to save another subset (~1 million variants) of that same base file while the 4.4 million subset was still indexing (Error message: Task was killed due to the node running low on memory.), so this may be worth investigating in future iterations of the save module and we might want to figure out what our exact current limitations are in terms of saving until then.

Yeah, a current limitation of this approach is that if you have a large file and are saving just a few records from it, this version will be slower than the old one. Once we can randomly seek to lines in the file, that will resolve. However, for a large query, it should be much faster.

To reduce memory pressure I will reduce the chunk size to 2 * 10,000 records (half of what it is now)

wingolab (approved):

PR #406, opened 2024-02-15
Parent of Origin Implementation

austinTalbot7241993:
In this PR I implement an initial parent of origin caller. This exploits the low-rank structure of such an effect to efficiently estimate the effect.

### DIFF
--- a/python/python/bystro/rare_variant/parent_of_origin.py
+++ b/python/python/bystro/rare_variant/parent_of_origin.py
@@ -68,6 +68,8 @@ class BasePOE:
     implements methods to test inputs for proper dimensionality and a
     method to classify heterozygotes based on their phenotypes
     """
+    def __init__(self):
+        self.parent_effect_ = None
 
     def _test_inputs(self, X, y):
         if not isinstance(X, np.ndarray):
--- a/python/python/bystro/rare_variant/parent_of_origin.py
+++ b/python/python/bystro/rare_variant/parent_of_origin.py
@@ -1,8 +1,59 @@
+"""
+This module implements a parent of origin effect caller. The parent of 
+origin effect refers to the difference in phenotypes 
+depending on whether the allele is inherited paternally or maternally. The 
+caller employs a method that draws inspiration from the community 
+detection problem studied in computer science.
+
+The distribution of phenotypes in the heterozygotes is modeled as a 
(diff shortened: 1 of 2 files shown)


python/python/bystro/rare_variant/parent_of_origin.py
```diff
@@ -0,0 +1,178 @@
...
+
+POESingleSNP(BasePOE)
+    A caller for the POE of a single SNP. No sparsity assumptions are 
+    implemented in this model.
+"""
+import numpy as np
+import numpy.linalg as la
+
+
+def get_low_rank(A):
```
cristinaetrv, line 42: Could you include a test for this function? It's not clear from the code/tests alone the relationship between get_low_rank and POESingleSNP and how they're meant to interact when using this method.

python/python/bystro/rare_variant/tests/test_parent_of_origin.py
```diff
@@ -0,0 +1,115 @@
...
+
+def cosine_similarity(vec1, vec2):
+    v1 = np.squeeze(vec1)
+    v2 = np.squeeze(vec2)
+    num = np.dot(v1, v2)
+    denom = la.norm(v1) * la.norm(v2)
+    return num / denom
+
+
+def generate_data(beta_m, beta_p, rng, n_individuals=10000, cov=None, maf=0.25):
```
cristinaetrv, line 21: You might want to include a quick test for generate_data as well to make sure you're getting what you expect here because another test is dependent on it.

python/python/bystro/rare_variant/parent_of_origin.py
austinTalbot7241993: Ah residual no longer needed. Will remove.

python/python/bystro/rare_variant/tests/test_parent_of_origin.py
cristinaetrv: For some context, I have a vcf simulator that I use for multiple tests so I have it under test just to make sure if it ever incurs an accidental typo or something that it's still working as expected in any of the tests that rely on it, so just a small test with a small n that tests that you're getting the expected shape/type of data I think would be sufficient.

cristinaetrv (approved):
Lgtm!

PR #409, opened 2024-02-19
WIP Feature/annotation interface

dlin30:
Added Listener for beanstalkd to annotation_interface
Added Test for annotation_interface

last piece is understanding the opensearch better (where are we getting the client info from beanstalkd or config dir)

Stacked on #385


akotlar (dismissed):
nvm: wrong pr!

PR #410, opened 2024-02-20
Don't attempt to load conf_dir, ensure camel case for incoming messages and serialized values

akotlar:
* Don't attempt to load conf_dir, ensure camel case for incoming messages and serialized values


cristinaetrv (approved):
Lgtm!

PR #411, opened 2024-02-20
[WIP] [annotation] Cleanup IO.pm

akotlar:
* Cleanup IO.pm, removing unused functions
* Fix support for zip files; while gzip and pgzip can decompress zip files, bgzip cannot.


wingolab (approved):
Looks good to me.

PR #413, opened 2024-02-22
Add initial loading and preprocess for PRS.

cristinaetrv:
This PR introduces the initial section of the preprocessing pipeline for the Polygenic Risk Score (PRS) module which implements the clumping and thresholding (C+T) method. 

Upcoming PRs will include the rest of the preprocess steps and core PRS calculation logic and ingestion of GWAS summary statistic file.


python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -0,0 +1,113 @@
...
+import re
+import pandas as pd
+from pyarrow import feather  # type: ignore
+
+logger = logging.getLogger(__name__)
+
+
+def _load_preprocessed_dosage_matrix(dosage_matrix_path: str) -> pd.DataFrame:
+    """Load in dosage matrix generated from bystro-vcf."""
+    return feather.read_feather(dosage_matrix_path)
```
akotlar, line 14: This will not work for any meaningfully large dataset. This has to use the approach seen in the prs tests.

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -0,0 +1,113 @@
...
+    A dictionary where keys are 'GeneticMap{chromosome_number}' and values are the corresponding
+    DataFrames loaded from Feather files.
+    """
+    file_pattern = f"{map_directory_path}/*.feather"
+    map_file_list = glob.glob(file_pattern)
+    genetic_maps = {}
+    for file in map_file_list:
+        try:
+            genetic_map = pd.read_feather(file)
+            match = re.search(r"chromosome_(\d+)_genetic_map", file)
```
akotlar, line 40: If we can't determine the chromosome, is the file useful?

If it isn't, which I don't think it can be, 1) you have to raise an exception if you don't have a match, rather than logging a warning, 2) you shouldn't load the file until you have successfully parsed the match.

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -0,0 +1,113 @@
...
+                genetic_maps[key] = genetic_map
+                logging.info("Successfully read %s from %s", key, file)
+            else:
+                logging.warning("Could not determine chromosome number from %s", file)
+        except Exception:
+            logging.exception("Failed to read from %s", file)
+    return genetic_maps
+
+
+def _preprocess_scores(gwas_scores_filepath: str) -> pd.DataFrame:
```
akotlar, line 53: If the goal of this is to generate a locus column in the GWAS covariates tables, this isn't the most optimal approach.  Instead, following the approaches described in https://stackoverflow.com/questions/39291499/how-to-concatenate-multiple-column-values-into-a-single-column-in-pandas-datafra, you should use Pandas vectorized functions:

df['CHR'] + ':' + df['POS'] + x["OTHER_ALLELE"] + x["EFFECT_ALLELE"]

after checking that those columns exist:

```py
columns = set(df.columns)
if "CHR" not in columns:
   raise ValueError('CHR does not exist')
etc

File: python/python/bystro/prs/preprocess_for_prs.py [hunk 4]
```diff
@@ -0,0 +1,113 @@
...
+            genetic_map = pd.read_feather(file)
+            match = re.search(r"chromosome_(\d+)_genetic_map", file)
+            if match:
+                chrom_num = match.group(1)
+                key = f"GeneticMap{chrom_num}"
+                genetic_maps[key] = genetic_map
+                logging.info("Successfully read %s from %s", key, file)
+            else:
+                logging.warning("Could not determine chromosome number from %s", file)
+        except Exception:
```
akotlar, line 48: prefer RuntimeError, rather than generic Exception

python/python/bystro/prs/preprocess_for_prs.py
akotlar:
```py
df['CHR'] = df['CHR'].as_type('str')
df['POS'] = df['POS'].as_type('str')
df['OTHER_ALLELE'] = df['OTHER_ALLELE'].as_type('str')
df['EFFECT_ALLELE'] = df['EFFECT_ALLELE'].as_type('str')

scores["ID_effect_as_alt"] = df['CHR'] + ':' + df['POS'] + x["OTHER_ALLELE"] + x["EFFECT_ALLELE"]
```

python/python/bystro/prs/preprocess_for_prs.py
akotlar: //2024-03-12 We talked about this, and agreed that in v1 we go with this approach and next PR we'll switch to the memory efficient one.

PR #417, opened 2024-03-06
Bugfix/ppca transform subset

austinTalbot7241993:
Fixed transform_subset method.


akotlar (changes requested):
Please add a test, since it looks like our previous test coverage didn't catch this.

akotlar (approved):
LGTM!

PR #420, opened 2024-03-07
Add support for Opensearch 2.12; Bring field analyzers in line with b10

akotlar:
* Copies the mapping settings from bystro b10, which was used in the publication (see attached) and which we know works well
  * [hg19.mapping.b10.yml.zip](https://github.com/bystrogenomics/bystro/files/14531044/hg19.mapping.b10.yml.zip) 
* Adds support for our Opensearch 2.12 deployment, which defaults to https and having a username/password

### DIFF
--- a/config/hg19.mapping.yml
+++ b/config/hg19.mapping.yml
@@ -29,6 +29,31 @@ index_settings:
           - uppercase
           - asciifolding
     filter:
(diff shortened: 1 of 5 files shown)


dlin30 (approved):
LGTM

PR #425, opened 2024-03-10
Add script to split vcf file by chromosome

akotlar:
* Adds script to split vcf file by chromosome. Used for dbSNP track creation.


wingolab (dismissed):
looks good to me; could you run perl linting and re-pull?

wingolab (approved):

PR #426, opened 2024-03-10
Rotation Domain Adaptation

austinTalbot7241993:
This provides an initial implementation of domain adaptation for the proteomics module. Given a dataset composed of data obtained from multiple experiments, it finds the centroid mean and covariance matrices then maps each dataset to match this covariance and mean using an affine transformation. Future work will be used to replace use of the empirical covariance matrix with estimators that scale better and work with small sample sizes.


akotlar (approved):
Has test and documentation. Code follows our usual structure. LGTM!

PR #430, opened 2024-03-12
[proteomics] listener

dlin30:
added listener for annotation interface

### DIFF
--- a/python/python/bystro/proteomics/listener_annotation_interface.py
+++ b/python/python/bystro/proteomics/listener_annotation_interface.py
@@ -3,8 +3,10 @@ import argparse
 import logging
 from ruamel.yaml import YAML
 from pathlib import Path
-import pyarrow.dataset as ds
+import pyarrow.dataset as ds    # type: ignore
 from opensearchpy import OpenSearch
+import pandas as pd
+from datetime import datetime, timezone
 
 from bystro.beanstalkd.worker import listen, QueueConf, ProgressPublisher
 from bystro.beanstalkd.messages import BaseMessage, CompletedJobMessage, SubmittedJobMessage
@@ -15,7 +17,6 @@ from bystro.proteomics.annotation_interface import (
 
 logger = logging.getLogger(__file__)
 
-EFS_BASE_PATH = "/seqant/user-data/"
 PROTEOMICS_TUBE = "proteomics"
 
 
@@ -25,8 +26,6 @@ class ProteomicsJobData(BaseMessage, frozen=True, rename="camel"):
 
     Parameters
     ----------
-    submission_id: str
-        The unique identifier for the job.
     data_path: str
         The path to the proteomics data file.
     out_dir: str
@@ -37,7 +36,6 @@ class ProteomicsJobData(BaseMessage, frozen=True, rename="camel"):
         The name of the OpenSearch index from which annotation data will be queried.
     """
 
-    submission_id: str
     data_path: str
     out_dir: str
     annotation_query: str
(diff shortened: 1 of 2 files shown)


python/python/bystro/proteomics/listener_annotation_interface.py
```diff
@@ -0,0 +1,122 @@
...
+        The path to the proteomics data file.
+    out_dir: str
+        The directory to write the results to.
+    annotation_query: str
+        The query string used to fetch annotation data from OpenSearch index.
+    index_name: str
+        The name of the OpenSearch index from which annotation data will be queried.
+    """
+
+    submission_id: str
```
akotlar, line 40: You don't need to provide submission_id here, as BaseMessage already provides that

python/python/bystro/proteomics/listener_annotation_interface.py
```diff
@@ -0,0 +1,122 @@
...
+    with Path(queue_conf_path).open(encoding="utf-8") as queue_config_file:
+        raw_queue_conf = YAML(typ="safe").load(queue_config_file)
+    beanstalk_conf = raw_queue_conf["beanstalkd"]
+    return QueueConf(addresses=beanstalk_conf["addresses"], tubes=beanstalk_conf["tubes"])
+
+
+def handler_fn(_publisher: ProgressPublisher, job_data: ProteomicsJobData) -> str:
+    logger.info("Processing Proteomics job: %s", job_data)
+
+    efs_out_dir = Path(EFS_BASE_PATH) / job_data.out_dir
```
akotlar, line 61: You don't need this. job_data.out_dir is a fully qualified path to the output.

We also can't assume job_data.out_dir refers to an EFS path; it could be an S3 or google cloud bucket path in the future.

python/python/bystro/proteomics/listener_annotation_interface.py
```diff
@@ -0,0 +1,122 @@
...
+from bystro.beanstalkd.worker import listen, QueueConf, ProgressPublisher
+from bystro.beanstalkd.messages import BaseMessage, CompletedJobMessage, SubmittedJobMessage
+from bystro.proteomics.annotation_interface import (
+    get_annotation_result_from_query,
+    join_annotation_result_to_proteomics_dataset,
+)
+
+logger = logging.getLogger(__file__)
+
+EFS_BASE_PATH = "/seqant/user-data/"
```
akotlar, line 18: no need for this, see comment below

akotlar:
also need to fix the mypy errors :)

akotlar:
Oh, we're also assuming the proteomics upload is a feather file here. For now, I think we should not make that assumption, and instead read it as a .tsv.

This is because we don't yet process the proteomics file in any way in either the api or bystro-web, so a user couldn't use this as is

python/python/bystro/proteomics/listener_annotation_interface.py
```diff
@@ -0,0 +1,122 @@
...
+    return QueueConf(addresses=beanstalk_conf["addresses"], tubes=beanstalk_conf["tubes"])
+
+
+def handler_fn(_publisher: ProgressPublisher, job_data: ProteomicsJobData) -> str:
+    logger.info("Processing Proteomics job: %s", job_data)
+
+    efs_out_dir = Path(EFS_BASE_PATH) / job_data.out_dir
+    efs_out_dir.mkdir(parents=True, exist_ok=True)
+
+    dataset = ds.dataset([job_data.data_path], format="arrow")
```
akotlar, line 64: We did talk about assuming the proteomics file is a feather file, but until we actually process the proteomics files to be feather files, we should assume they are tsvs, and use pandas.from_csv(path, sep="\t").

akotlar (approved):
LGTM

Issue #421, opened 2024-03-08
Sprint 9 Task List


akotlar:
Webapp
- [x] Fix S3 upload  - March 13th for PR @akotlar 
- [x] Fix search cards to show all data - March 13th for PR @akotlar 
- [x] Hide settings wheel in queued file list (it is not currently enabled)  - March 11th @akotlar 
- [x] Autocomplete field names - March 15th @akotlar 
- [x] Fix search phase exception with long query: @akotlar noticed that an aggregations (Heterozygotes per variant  was cancelled after it ran for a certain amount of time). If cannot be solved, remove the offending aggregation scripts. https://github.com/bystrogenomics/bystro-web/issues/388 - @akotlar - March 28th
- [x] Finish making AMIs for bystro-web, bystro, and opensearch servers - hard deadline is March 29th.
- [x] (stretch) Add ability to rename genetic submissions in front end, before uploading them, by clicking on the name  -@akotlar 
- [x] (stretch) Add ability to stitch together datasets.
- [ ] (stretch) Add Jupyter notebook launcher
- [ ] (stretch) support (refSeq:brca1), which would expand to (refseq.\\*:brca1)
- [x] (stretch) Update numerical facets to allow selecting on ranges using slider

Annotation
- [x] Test hg19_v8 database by comparing results, row by row, on trio_trim.vcf.gz between hg19_v8 and dave's instance b10 database - March 15th @akotlar 
- [x] Create hg38 database - March 11th @akotlar 
- [x] Test hg38 database by comparing results, row by row, on trio_trim.vcf.gz between hg19_v8 and dave's instance b10 database - March 18th @akotlar 

Ancestry
- [x] Retrain ancestry model (hg38) with 76k gnomad loadings - @cristinaetrv  - March 13th
- [x] Liftover array set - @cristinaetrv - March 13th 
- [x] Liftover gnomad loadings - @cristinaetrv  - March 14th 
- [x] Add assembly to AncestryData, update bystro-web to submit assembly, and use assembly when choosing model (https://github.com/bystrogenomics/bystro/issues/419) - @akotlar - March 13th
- [x] Create ancestry docker container - @akotlar - March 14th
- [x] Add support for choosing best covariate set for ancestry - @akotlar - March 15th
- [x] Gives Thomas new ancestry code to test on healthy aging study - March 20th @cristinaetrv 
- [x] Add Ancestry table version that just has IDs and top hit superpops that can be expanded to larger table @akotlar 

[...]

PR #431, opened 2024-03-13
Issue #421: Update opensearch gnomad mappings to match updates hg19 and hg38 data

akotlar:
* Updates hg38.mapping.yamland hg19.mapping.yaml to support new our new hg19 and hg38 databases
* Updates hg38.clean.yml and hg19.clean.yml to match the bystro annotator definitions of our new databases

hg38.mapping.yml is now a separate definition, rather than a symlink to hg19.mapping.yml. It is identical to hg19.mapping.yml besides the gnomad sections, which have to be different since gnomad v4 was used in the hg38 database.

### DIFF
--- a/config/hg19.clean.yml
+++ b/config/hg19.clean.yml
@@ -1,10 +1,34 @@
 ---
 assembly: hg19
 build_author: alexkotlar
-build_date: 2023-11-08T00:03:00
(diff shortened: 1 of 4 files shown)


cristinaetrv (approved):
Lgtm!

PR #432, opened 2024-03-13
[ancestry] Add support for choosing ancestry model with lowest missingness

akotlar:
* Chooses model that results in the largest number of retained rows, which is the best single indicator of model fit

A future improvement could be checking whether the matching rows have lots of NAs, but this really won't help much, because samples with high sample missingness are liable to not be good regardless.

### DIFF
--- a/python/python/bystro/ancestry/inference.py
+++ b/python/python/bystro/ancestry/inference.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+from typing import Generator
 import psutil
 
 from msgspec import Struct
@@ -9,9 +10,9 @@ import numpy as np
 import pandas as pd
 
 import pyarrow.compute as pc  # type: ignore
-from pyarrow.dataset import Dataset  # type: ignore
+from pyarrow.dataset import Dataset, Scanner  # type: ignore
 
-from sklearn.ensemble import RandomForestClassifier
+from sklearn.ensemble import RandomForestClassifier # type: ignore
 
 from bystro.ancestry.ancestry_types import (
     AncestryResults,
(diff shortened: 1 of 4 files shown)


python/python/bystro/ancestry/listener.py
```diff
@@ -82,16 +72,43 @@ def _get_model_from_s3(assembly: str) -> AncestryModel:
...
+    )
+    array_model = _get_one_model_from_s3(
+        pca_local_key_array, rfc_local_key_array, pca_file_key_array, rfc_file_key_array
+    )
+
+    models = AncestryModels(gnomad_model, array_model)
 
     # Update the cache with the new model
-    if len(models_cache) >= 2:
+    if len(models_cache) >= 1:
```
akotlar, line 105: cache one assembly since we now have 2x the models per assembly, and in the future 3x

python/python/bystro/ancestry/inference.py
```diff
@@ -161,25 +178,60 @@ def infer_ancestry(ancestry_model: AncestryModel, genotypes: Dataset) -> Ancestr
...
+            "Found %d rows in genotypes matching array PCA loadings",
+            array_matching_row_count,
+        )
+
+        logger.info(
+            "Memory usage after dosage matrix filtered row counting: %s (MB)",
+            psutil.Process(os.getpid()).memory_info().rss / 1024**2,
+        )
+
+        if gnomad_matching_row_count >= array_matching_row_count:
```
cristinaetrv, line 213: Since the array set has about half of the gnomad set (~34k vs ~74k), the gnomad set could beat the array set even in a case where the percentage is 100% variant match to the array set with just 50% variant match to the gnomad set. Could we adjust it to take the highest percent retention instead of highest row count?

cristinaetrv (changes requested):
Recommend changing comparing counts of matching rows to percent variants retained during model selection to reduce chance of bias due to difference in variant set sizes between the models

python/python/bystro/ancestry/inference.py
akotlar: @austinTalbot7241993 what do you think? Would it be better to use absolute number of snps / informative PC's, or completeness respective to the model (or a hybrid strategy)?

akotlar: Based on Dave's comments, if you're open to it, unless Austin strongly feels otherwise, let's leave as is, and then run the test? That way we don't block deployment of something that is a strict improvement over what we had before, and can optimize in a subsequent PR?

cristinaetrv (approved):
Overall, looks good. 

After discussion with Austin and Dave, we've decided to run tests on whether percentage of variants retained or raw count of variants retained will give more accurate ancestry results when choosing between models with different SNP sets. If percentage ends up performing better after tests, adding that in will be a next step.

PR #434, opened 2024-03-14
Feature/streaming proteomics

dlin30:
Added streaming.py to fetch a file from streamFile endpoint
Added fetch command to the bystro-api cli.py


python/python/bystro/api/streaming.py
```diff
@@ -0,0 +1,59 @@
...
+        Whether to print the result of the fetch operation, by default True.
+    """
+    if not job_id:
+        raise ValueError("Please specify a job id")
+
+    state, auth_header = authenticate()
+    url = state.url + GET_STREAM_ENDPOINT.format(job_id=job_id)
+
+    payload = {
+        "output": str(output),
```
akotlar, line 32: This doesn't work. You can't cast the boolean to string because the Bystro-api endpoint you wrote expects a boolean

akotlar (changes requested):
Needs a fix

akotlar (approved):
LGTM

PR #435, opened 2024-03-14
Feature/issue 421 remove refseq clinvar

akotlar:
* We do not currently handle refSeq.clinvar in a way that makes sense compared to our current annotations; order between fields in the struct is not necessarily preserved. Once we fix it, we can add back.

Dave Cutler agreed that refSeq.clinvar is no longer useful

Stacked on https://github.com/bystrogenomics/bystro/pull/432

### DIFF
--- a/config/hg19.clean.yml
+++ b/config/hg19.clean.yml
@@ -147,18 +147,6 @@ tracks:
         - rfamAcc
         - tRnaName
         - ensemblID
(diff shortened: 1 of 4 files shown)


cristinaetrv (approved):
Lgtm!

cristinaetrv (approved):
Lgtm!

PR #438, opened 2024-03-18
Issue #421: Make discordant an actual boolean field, by outputting 'false'/'true' labels

akotlar:
* Make discordant an actual boolean field, by outputting 'false'/'true' labels

This makes it easier to search the field, as well as to import it as a boolean in Pandas/R.


dlin30 (approved):
LGTM

PR #440, opened 2024-03-19
Issue#421: Improve hg19 and hg38 opensearch index definitions

akotlar:
* We were missing id, inputRef, vcfPos, and had "alt" as a text field, rather than keyword. Text fields cannot be aggregated on, and I don't see a use case for full text matches on alt fields anyway (only use case would be insertions, but there it would be of extremely limited use, matching "+ATCGGGG" if the user entered "+ATCG"
* Fixes clinvar queries, making queries on fields like CLNSIG, such as "pathogenic likely_pathogenic Benign(2)" work without stripping special characters

Much more extensive testing is needed in the future, but this fixes a large number of queries on clinvar fields.

We do as a result lose disease synonyms on clinvar (because catenate_filter is, in Elasticsearch > 6 incompatible with synonyms filters) but that is not as much of a problem as not being able to search for terms that appear directly in results. We will resolve synonyms in the future by following the advice on this topic mentioned in https://discuss.elastic.co/t/unable-to-bypass-restriction-with-synonym-token-filter/195334/4

### DIFF
--- a/config/hg38.mapping.yml
+++ b/config/hg38.mapping.yml
@@ -35,7 +35,7 @@ index_settings:
         catenate_words: true
         catenate_numbers: true
         catenate_all: true
-        preserve_original: false
+        preserve_original: true
         generate_word_parts: false
         stem_english_possessive: true
         generate_number_parts: false
@@ -46,7 +46,7 @@ index_settings:
         catenate_words: true
         catenate_numbers: true
         catenate_all: true
-        preserve_original: false
+        preserve_original: true
         generate_word_parts: true
         stem_english_possessive: true
         generate_number_parts: false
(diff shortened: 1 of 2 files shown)


config/hg38.mapping.yml
```diff
@@ -203,6 +203,14 @@ index_settings:
           - english_stemmer
           - description_synonyms
           - disease_synonyms
+      search_english_simple:
+        type: custom
+        tokenizer: whitespace
+        filter:
+          - lowercase
+          - asciifolding
+          - catenate_filter_split
```
akotlar, line 212: using this filter, which is also used during indexing phase, appears critical.

wingolab (approved):

PR #448, opened 2024-03-22
Domain Adaptation

austinTalbot7241993:
This separates out the preprocessing into a domain adaptation model. We expand on the current method (rotation) and add methods that are more suitable for small sample sizes.


python/python/bystro/domain_adaptation/spectral_adaptation.py
```diff
@@ -0,0 +1,331 @@
...
+        The number of datasets.
+    mu_list : List[np.ndarray]
+        A list of means for each dataset in `X_list`.
+    mu_0 : np.ndarray
+        The target mean for all datasets after adaptation.
+    n_covariates : int
+        The number of features (covariates) in the datasets.
+    """
+
+    def __init__(
```
akotlar, line 65: In the future please update this to check that lambda is within 0-1 bounds.

python/python/bystro/domain_adaptation/spectral_adaptation.py
```diff
@@ -0,0 +1,331 @@
...
+
+    This adaptation method works by calculating the mean of each dataset 
+    in a list of datasets (X_list) and adjusting each dataset to have 
+    the same mean. This can be useful in scenarios where datasets from
+    different sources or conditions should be brought to a common scale 
+    or reference point before analysis.
+
+    Parameters
+    ----------
+    prior_mu : bool, default=True
```
akotlar, line 43: I would prefer that we just have lamb = 0 by default, so that we "do what I mean" (if I set lambda to 0 I want no regularization, else I do). Non-blocking comment

python/python/bystro/domain_adaptation/spectral_adaptation.py
```diff
@@ -0,0 +1,331 @@
...
+        weighted : bool, default=False
+            If True, weights the contribution of each dataset by its number 
+            of samples in calculating the overall mean.
+        """
+        self.prior_mu: bool = prior_mu
+        self.lamb: float = lamb
+        self.weighted: bool = weighted
+        self.J: int = 0
+        self.mu_list: List[np.ndarray] = []
+        self.mu_0: np.ndarray = np.array([])  # Placeholder until set
```
akotlar, line 89: we can remove the comment

akotlar (approved):
Changes are non-blocking. I will do a deeper dive next week on this module

PR #449, opened 2024-03-23
Issue #421 Bystro Python API Dockerfile & Ancestry API/CLI  & Reduce Ancestry Memory Usage

akotlar:
* Adds docker file for Bystro's python library
* Creates ancestry api and cli code for calculating ancestry scores
* Removes unneeded dependencies from Cargo.toml, to speed up builds
* Improves Makefile by introducing the ability to make production builds and install from wheel.
* Reduce ancestry memory usage by reading in sample chunks
* Cache ancestry scores to local disk to reduce S3 fetching

To test what is here:

```
docker pull akotlar/bystro-api
docker run -v /path/to/local/data:/data  akotlar/bystro-api ancestry score --in /data/trio.trim.vep.vcf.gz --assembly hg19 
```

[trio.trim.vep.vcf.gz](https://github.com/bystrogenomics/bystro/files/14730266/trio.trim.vep.vcf.gz)

The api function is a port of ancestry/listener.py handler_fn.

### DIFF
--- a/.github/workflows/conda-build.yml
+++ b/.github/workflows/conda-build.yml
@@ -18,10 +18,10 @@ jobs:
           auto-activate-base: true
           activate-environment: true
           python-version: ${{ matrix.python-version }}   
-    - name: Build Bystro using Maturin
+    - name: Install Bystro using Maturin
       run: |
         . .initialize_conda_env.sh
-        make build
+        make install
       shell: bash
     - name: Verify Bystro wheel exists
       run: |
@@ -30,3 +30,8 @@ jobs:
           exit 1
         fi
         echo "Bystro wheel found! Build succeeded."
+    - name: Verify bystro-api is installed
+      run: |
+        set -e
+        bystro-api --help
+        echo "bystro-api installed successfully"
--- a/.github/workflows/conda-build.yml
+++ b/.github/workflows/conda-build.yml
@@ -22,6 +22,7 @@ jobs:
       run: |
         . .initialize_conda_env.sh
         make install
+        echo "PATH=$CONDA/bin:$PATH" >> $GITHUB_ENV
       shell: bash
     - name: Verify Bystro wheel exists
       run: |
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,34 @@
-# Assumes you have run ". .initialize-conda-env.sh"; since each make command runs in a separate subshell we need this to happen first
+ray-start-local:
+	ray stop && ray start --head --disable-usage-stats
 
-build:
-	cd python && maturin build --release && cd ../
+ray-stop-local:
+	ray stop
 
-develop:
-	cd go/cmd/opensearch && go build && cd ../../../
-	cd python && maturin develop && cd ../
+build-python:
+	(cd python && maturin build --release)
 
-# Ray must be started with make serve-dev
-# without ray start, make serve-dev will succeed, but the handlers that rely on Ray will fail to start
-serve-dev: develop
-	ray stop && ray start --head
+build-python-dev:
+	(cd python && maturin develop)
+
+install-python: build-python
+	pip install python/target/wheels/*.whl
+
+install-go:
+	(cd ./go && go install bystro/cmd/opensearch)
+
+install: install-python install-go
+
+uninstall:
+	pip uninstall -y bystro
+	(cd ./go && go clean -i bystro/cmd/opensearch)
+
+develop: install-go build-python-dev ray-start-local
+
+run-local: install ray-start-local
+
+# Currently assumes that Perl package has been separately installed
+serve-local: ray-stop-local run-local
 	pm2 delete all 2> /dev/null || true && pm2 start startup.yml
+
+# Currently assumes that Perl package has been separately installed
+serve-dev: ray-stop-local start-local develop pm2
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,6 @@ install-python: build-python
 	pip install python/target/wheels/*.whl
 
 install-go:
-	(cd ./go && go install bystro/cmd/dosage)
 	(cd ./go && go install bystro/cmd/opensearch)
 
 install: install-python install-go
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ install-python: build-python
 	pip install python/target/wheels/*.whl
 
 install-go:
+	(cd ./go && go install bystro/cmd/dosage)
 	(cd ./go && go install bystro/cmd/opensearch)
 
 install: install-python install-go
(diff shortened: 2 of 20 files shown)


Makefile
```diff
@@ -1,14 +1,34 @@
-# Assumes you have run ". .initialize-conda-env.sh"; since each make command runs in a separate subshell we need this to happen first
```
akotlar, line 1: Will document after April 1 the new commands. The old `develop` and `serve-dev` commands we all use remain, and function as before

python/Cargo.toml
```diff
@@ -10,11 +10,4 @@ name = "bystro"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-arrow = "40.0.0"
```
akotlar, line 13: all unused

python/python/bystro/ancestry/inference.py
```diff
@@ -27,6 +28,7 @@
 from bystro.utils.timer import Timer
 
 logger = logging.getLogger(__name__)
+warnings.simplefilter(action="ignore", category=FutureWarning)
```
akotlar, line 31: removes the warning about how sklearn 1.22 is using deprecated Panda's is_sparse function. Without this every time you run ancestry on the CLI you get a lot of ugly warnings that are not useful to the user.

python/python/bystro/ancestry/listener.py
```diff
@@ -28,88 +24,6 @@
 logger = logging.getLogger()
 
 ANCESTRY_TUBE = "ancestry"
-ANCESTRY_BUCKET = "bystro-ancestry"
```
akotlar, line 31: all of the ancestry model loading was moved to a dedicated module so that it could be cleanly shared in the ancestry api module, without having to load ray and beanstalkd (this allows us to not have to ray start --head before running  ancestry cli commands)

python/python/bystro/ancestry/model.py
```diff
@@ -0,0 +1,185 @@
...
+ANCESTRY_BUCKET = "bystro-ancestry"
+GNOMAD_PCA_FILE = "gnomadset_pca.csv"
+GNOMAD_RFC_FILE = "gnomadset_rfc.skop"
+ARRAY_PCA_FILE = "arrayset_pca.csv"
+ARRAY_RFC_FILE = "arrayset_rfc.skop"
+
+models_cache: dict[str, AncestryModels] = {}
+
+
+def get_one_model_from_s3(pca_local_path, rfc_local_path, pca_file_key, rfc_file_key) -> AncestryModel:
```
akotlar, line 32: unchanged besides renamed variables and added docstring

python/python/bystro/ancestry/model.py
```diff
@@ -0,0 +1,185 @@
...
+                    f"{rfc_file_key} ancestry model not found. This assembly is not supported."
+                ) from e
+            raise
+
+    logger.debug("Downloaded PCA file and RFC file in %f seconds", timer.elapsed_time)
+
+    return get_one_model_from_file_system(pca_local_path, rfc_local_path)
+
+
+def get_models_from_s3(assembly: str) -> AncestryModels:
```
akotlar, line 73: unchanged besides renamed variables and added docstring

python/python/bystro/ancestry/model.py
```diff
@@ -0,0 +1,185 @@
...
+    if len(models_cache) >= 1:
+        # Remove the oldest loaded model to maintain cache size
+        oldest_assembly = next(iter(models_cache))
+        del models_cache[oldest_assembly]
+    models_cache[assembly] = models
+
+    return models
+
+
+def get_one_model_from_file_system(pca_path: str, rfc_path: str) -> AncestryModel:
```
akotlar, line 119: the local filesystem versions are needed because our s3 bucket is not publicly accessible, and also because in the future we will cache the models to the local disk to save us costs on s3 (frequently pulling data from s3 is not a good idea from cost standpoint)

python/python/bystro/api/auth.py
```diff
@@ -12,7 +12,6 @@
 
 if not os.path.exists(DEFAULT_DIR):
     os.makedirs(DEFAULT_DIR, exist_ok=True)
-    print(f"Created default credentials path: {CREDENTIALS_PATH}")
```
akotlar, line 15: when running from Docker, this message would pop up on every ancestry score cli run, and is completely unhelpful.

python/python/bystro/search/index/listener.py
```diff
@@ -20,7 +20,7 @@
 
 TUBE = "index"
 
-_GO_HANDLER_BINARY_PATH = None
+_GO_HANDLER_BINARY_PATH: str | None = None
```
akotlar, line 23: fix a mypy issue

python/python/bystro/utils/compress.py
```diff
@@ -31,24 +31,61 @@ def _get_gzip_program_path() -> str:
...
+        cmd = f"{cmd} | {pipe_to_cmd}"
 
+    return f"{cmd} > {out}"
 
-def get_decompress_to_pipe_cmd(input_path: str) -> str:
-    if IS_PIGZ:
-        # pigz doesn't benefit much from more than 2 threads during decompression
-        return f"{GZIP_EXECUTABLE} -d -c {input_path} -p 2"
+
+def get_decompress_to_pipe_cmd(input_path: str, pipe_to_cmd: str | None = None) -> str:
```
akotlar, line 68: updated to support bystro-vcf in a single pipe command (e.g. bgzip -d -c file.gz | bystro-vcf)

python/python/bystro/utils/compress.py
```diff
@@ -31,24 +31,61 @@ def _get_gzip_program_path() -> str:
 IS_BGZIP = GZIP_EXECUTABLE.endswith("bgzip")
 
 
-def get_compress_from_pipe_cmd(out: str) -> str:
+def get_compress_from_pipe_cmd(
```
akotlar, line 34: also updated this method to support piped commands, to be symmetrical with the decompress method

.github/workflows/conda-build.yml
```diff
@@ -18,10 +18,10 @@ jobs:
           auto-activate-base: true
           activate-environment: true
           python-version: ${{ matrix.python-version }}   
-    - name: Build Bystro using Maturin
```
akotlar, line 21: Testing whether we can install is more useful than whether we can build the wheel (both are tested here, since the wheel is made during the build-python step, which is automatically run as part of the install step)

python/python/bystro/ancestry/inference.py
```diff
@@ -229,40 +227,79 @@ def infer_ancestry(ancestry_models: AncestryModels, genotypes: Dataset) -> Ances
...
-    with Timer() as timer:
-        # Directly concatenate the batches using a generator expression
-        genotypes_df = pd.concat(_generate_batches(scanner))
+    samples = [name for name in genotypes.schema.names if name != "locus"]
 
-    logger.info(
-        "Memory usage after dosage matrix filtering: %s (MB)",
-        psutil.Process(os.getpid()).memory_info().rss / 1024**2,
-    )
+    # Take chunks of up to 500 samples
```
cristinaetrv, line 232: Lmk if I'm reading this wrong, but isn't the default chunk size 200 samples on line 34, or is it that the max 500 is a suggestion here?

cristinaetrv (approved):
Tested it, and it works as expected. Overall, looks good!

PR #450, opened 2024-03-23
[search] Switch to half_float for cadd, caddIndel, and remove phyloP, phastCons

akotlar:
* Remove phasCons, phyla which are no longer used
* Switch cadd and caddIndel.PHRED to half_float. In practice this very slightly increases index size (e.g. on 6 million sites 10.7GB vs 10.69GB), but speeds up indexing and gains us precision (vs scaled flat with factor 10, 25 minutes vs 24 minutes, 4%).

### DIFF
--- a/config/hg19.mapping.yml
+++ b/config/hg19.mapping.yml
@@ -411,8 +411,7 @@ mappings:
             dist:
               type: integer
     cadd:
-      type: scaled_float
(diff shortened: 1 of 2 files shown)


cristinaetrv (approved):

cristinaetrv:
Lgtm!

PR #451, opened 2024-03-24
[annotator] Issue#421 Stitch sample_list and genotype matrices

akotlar:
* Stitch genotype matrices
* Stitch sample lists

Both sample lists and genotype matrices are outputs of the pre-processor. When processing multiple input vcfs/snps we need to combine the sample lists and genotype matrices. This PR does that.

This is a basic, slow implementation. In a future version, I will read the genotype dosage matrices in parallel, accumulate large chunks of arrow records, and write them in batches. This will speed up processing nearly linearly in the number of threads. But for now, as a basic V1, this should be fine, and has the advantage of using very little memory, which enables us to deploy for much cheaper. 

This is live on bystro-dev


cristinaetrv (approved):
Walked through code together and tested it on bystro-dev. Made a few suggestions during discussion for UI clarity and it looks good now. Worked as expected during tests.

PR #453, opened 2024-03-28
Support empty dosage matrix outputs; this may occur if combining genotype-free files

akotlar:
This aims to prevent crashing when combining genotype-free files. I also don't prevent combining runs that have genotypes and those that don't. For now this feels safer and a tremendous edge case. We will likely disallow that in a followup PR.

### DIFF
--- a/Dockerfile.python
+++ b/Dockerfile.python
@@ -6,7 +6,7 @@ ENV GOBIN=/app/bin
 
 # Install the specific versions of the Go programs
 RUN go install github.com/akotlar/bystro-stats@1.0.0
-RUN go install github.com/bystrogenomics/bystro-vcf@2.1.1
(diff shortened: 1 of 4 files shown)


Dockerfile.python
```diff
@@ -6,7 +6,7 @@ ENV GOBIN=/app/bin
 
 # Install the specific versions of the Go programs
 RUN go install github.com/akotlar/bystro-stats@1.0.0
-RUN go install github.com/bystrogenomics/bystro-vcf@2.1.1
+RUN go install github.com/bystrogenomics/bystro-vcf@2.1.2;
```
akotlar, line 9: Needed because 2.1.2 contains a fix allowing dosageOutput when no genotypes in the VCF file (empty dosageOutput results)

dlin30 (approved):
LGTM

PR #454, opened 2024-03-28
[annotator] Improve configuration filees

akotlar:
* Rename hg38.clean.yml -> hg38.yml and hg19.clean.yml -> hg19.yml, to simplify deployment (which relies on hg38/hg19.yml)
* Remove tmp_dir by default. Doing so allows us to scale to much larger datasets by writing to EFS directly. EFS is fast enough and we only write in a sequential manner, which works well from a throughput standpoint
* Add max_slices_per_pit that is above the default (100) for the cluster, to solve saving large jobs (which otherwise crash)

### DIFF
--- a/config/hg38.yml
+++ b/config/hg38.yml
@@ -36,7 +36,7 @@ fileProcessors:
   vcf:
     args: --emptyField NA --sample %sampleList% --keepPos --keepId --dosageOutput %dosageMatrixOutPath%
     program: bystro-vcf
-files_dir: /mnt/files1/bystro_annotator/raw_files/hg38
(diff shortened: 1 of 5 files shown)


config/hg38.yml
```diff
@@ -615,7 +615,6 @@ tracks:
     - build_author: alexkotlar
       build_date: 2024-03-11T12:24:00
       build_field_transformations:
-        CLNDISDB: split [|]
```
akotlar, line 618: we don't actually have this feature

dlin30 (approved):
LGTM

PR #436, opened 2024-03-15
Add AD GWAS summary statistics for PRS C+T method.

cristinaetrv:
Add a feather file with AD GWAS summary statistics from this publication: 
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9005347/

The file has been processed to include only SNPs with p-value < 0.05 for a total of 1,197,845 sites.

Column headers are formatted to match GIANT study format:

CHR | POS | OTHER_ALLELE | EFFECT_ALLELE | P | SNPID | BETA

In future, we will add compatibility with more GWAS summary statistics or user input of summary statistics with the same headers to run PRS. We may also change where these reference files are located in future updates.

PR #457, opened 2024-04-02
Add hg19 version of AD GWAS summary statistics for PRS.

cristinaetrv:
See #436 for more general information about this dataset in hg38. 

Add a feather file with AD GWAS summary statistics in hg19 from the following publication so that PRS can be used with this dataset in hg19 or hg38:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9005347/


akotlar (approved):
LGTM! In the future we should find a remote repo to pull these from; 47MB is relatively large if we have many of these.

PR #458, opened 2024-04-03
Implementation Multitask Supervision

austinTalbot7241993:
In our paper, and more broadly general audiences, it is useful to have a (1) marginal supervision implementation and (2) to have a multitask version. In this PR we implement both of these, and improve the module in general.

### DIFF
--- a/python/python/bystro/supervised_ppca/_base.py
+++ b/python/python/bystro/supervised_ppca/_base.py
@@ -66,6 +66,41 @@ from bystro.covariance._base_covariance import (
 from bystro._template_sgd_np import BaseSGDModel
 
 
+def _get_projection_matrix(W_: Tensor, sigma_: Tensor):
+    """
+    This is currently just implemented for PPCA due to nicer formula. Will
+    modify for broader application later.
+
+    Computes the parameters for p(S|X)
+
+    Description in future to be released paper
+
+    Parameters
+    ----------
+    W_ : Tensor(n_components,p)
+        The loadings
+
+    sigma_ : Tensor
+        Isotropic noise
+
+    Returns
+    -------
+    Proj_X : Tensor(n_components,p)
+        Beta such that np.dot(Proj_X, X) = E[S|X]
+
+    Cov : Tensor(n_components,n_components)
+        Var(S|X)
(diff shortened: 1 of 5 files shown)


python/python/bystro/supervised_ppca/gf_dropout_pt.py
```diff
@@ -171,27 +164,35 @@ def fit(  # type: ignore[override]
...
             sigm = nn.Sigmoid()
-            supervision_loss: nn.BCELoss | nn.MSELoss = nn.BCELoss()
+            supervision_loss: nn.BCELoss | nn.MSELoss = nn.BCELoss(
+                reduction="mean"
+            )
+
+            mod = LogisticRegression()
+            mod.fit(X, 1.0 * y)
+            b_ = torch.tensor(mod.intercept_.astype(np.float32))
+            print(b_)
```
akotlar, line 179: do you need this print statement?

python/python/bystro/supervised_ppca/_base.py
```diff
@@ -66,6 +66,41 @@
 from bystro._template_sgd_np import BaseSGDModel
 
 
+def _get_projection_matrix(W_: Tensor, sigma_: Tensor):
+    """
+    This is currently just implemented for PPCA due to nicer formula. Will
```
akotlar, line 71: Is this comment still relevant, and if so, can we remove it and  make implementing it more broadly a Github issue?

akotlar (changes requested):
2 minor comments

python/python/bystro/supervised_ppca/_base.py
austinTalbot7241993: Someday we'll want to also include factor analysis which will have non-isotropic noise. Don't know how soon that will be necessary however. Up to you.

python/python/bystro/supervised_ppca/gf_dropout_pt.py
austinTalbot7241993: Lol no, removing now.

akotlar (approved):
LGTM

PR #459, opened 2024-04-03
[search] Fix save handler memory usage 

akotlar:
* On big_daly, previous code would OOM, because pyarrow fetches up to 17 batches of 131,000 rows (1 batch fetched + 16 pre-fetched). 
* Previously we did not report progress during dosage matrix filtering, which can take a long time for large datasets. this PRremedies that
* We just released a bystro-vcf version that stores 1 byte for each dosage, halving memory and disk requirements. Now bystro installs that version of bystro-vcf
* Previously if a dosage matrix wasn't found or was empty, such as when annotating a file that did not have a dosage matrix, the save handler would crash.

### DIFF
--- a/.github/workflows/perl-package-build.yml
+++ b/.github/workflows/perl-package-build.yml
@@ -97,7 +97,7 @@ jobs:
       - name: Install one-off Bystro dependencies
         shell: bash
         run: |
-          go install github.com/bystrogenomics/bystro-vcf@2.1.0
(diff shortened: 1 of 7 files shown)


cristinaetrv (approved):
Lgtm. We discussed memory leaks when batching caused by batch/fragment readahead defaults in advance of this PR, and this looks like it will take care of that issue.

PR #461, opened 2024-04-04
GPU Support for PPCA

austinTalbot7241993:
This modifies the PPCA module to allow for GPU support for the models.


akotlar (approved):
LGTM; prefer to have docstrings updated, but otherwise non-blocking.

Squash and merge when ready

PR #462, opened 2024-04-05
Fast dosage matrix filtering

akotlar:
* Halves the time on our 4 vCPU systems over EFS
* 2 minutes to filter big_daly genotype matrix on 32 threads + SSD
* Had to get rid of a checked in .feather file because CLRF to LF conversion forced every branch into a bad state, preventing commits. Lets discuss Monday on how to get that file back in.
* Refactors save handler to be more testable
* Fixes statistical filter pipeline not running even when defined
* Dosage matrix filtering now runs after annotation filtering and uses the loci that survive the filtering pipeline
* Improve message reporting, so that we do not set progress to 100% (or rather `progress: N variants`, when N variants are found in the query), until we actually finish dosage matrix filtering. Until then we print log messages that specify the progress of each stage (querying, annotation filtering, dosage matrix filtering) 

This requires https://github.com/bystrogenomics/bystro-vcf/pull/20 to be merged in order for tests to pass

Live on bystro-dev


Makefile
```diff
@@ -15,6 +15,7 @@ install-python: build-python
 
 install-go:
 	(cd ./go && go install bystro/cmd/dosage)
```
akotlar, line 17: Because we install here, we need to further down call the binary in the user's path, and not something in the go/cmd/** directories. This also requires us to use shell=True in subprocess.Popen; without that the environment isn't loaded and the binaries are inaccessible

python/python/bystro/search/index/listener.py
```diff
@@ -14,33 +14,12 @@
 from bystro.beanstalkd.worker import ProgressPublisher, QueueConf, listen
 from bystro.search.utils.annotation import get_config_file_path
 from bystro.search.utils.messages import IndexJobCompleteMessage, IndexJobData, IndexJobResults
-from bystro.utils.config import _get_bystro_project_root
 
 from msgspec import json
 
 TUBE = "index"
 
-_GO_HANDLER_BINARY_PATH: str | None = None
```
akotlar, line 23: Relying on a binary path to go/cmd/opensearch was incompatible with installing the tool into the user's path. This bug bit me when deploying to IBDGC.

python/python/bystro/prs/gwas_summary_stats/AD_sumstats_hg19_PMID35379992.feather
cristinaetrv, line 1: Is this just an artifact of squashing?

cristinaetrv: Ah I reread the description and imagine this is the CLRF file that needs to be converted. Do we need to delete it or can't it be converted simply?

python/python/bystro/search/save/handler.py
```diff
@@ -285,80 +327,14 @@ async def go(  # pylint:disable=invalid-name
...
-            pool = pa.default_memory_pool()
-            with Timer() as timer:
-                dataset = ds.dataset(parent_dosage_matrix_path, format="arrow")
-                mask = pc.field("locus").isin(loci)
-
-                # Use the dataset's scanner to stream batches through a filter
-                scanner = dataset.scanner(
-                    filter=mask,
-                    use_threads=True,
-                    batch_size=SAVE_FILTER_BATCH_READ_SIZE,
```
cristinaetrv, line 297: Just to confirm I'm understanding correctly, this section isn't necessary because all of this is handled in the dosage-filter go code now?

python/python/bystro/prs/gwas_summary_stats/AD_sumstats_hg19_PMID35379992.feather
cristinaetrv: [AD_sumstats_PMID35379992.feather](https://github.com/bystrogenomics/bystro/blob/master/python/python/bystro/prs/gwas_summary_stats/AD_sumstats_PMID35379992.feather) isn't producing any errors though, is that correct? They were generated with the same workflow but it's the hg38 version.

akotlar: yeah I'm not sure what's going on with these; another issue is that every change to a binary file copies the binary file in the repo, so repo size grows rapidly. I think we either want to use git-lfs for these, or, since there will be so many of these, just host them ourselves in a public bucket

python/python/bystro/search/save/handler.py
akotlar: correct!

config/hg19.yml
```diff
@@ -47,7 +47,7 @@ statistics:
   programPath: bystro-stats
   refTrackField: ref
   siteTypeField: refSeq.siteType
-temp_dir: null
+temp_dir: /mnt/annotator/tmp
```
akotlar, line 50: It turns out that for big_daly, and presumably similarly sized datasets, writing to local ssd tmp is faster, and because bystro writes data so efficiently, even for big_daly, which is 2.7TB compressed, we write only ~ 120GB compressed, so tmp storage can be fairly modest.

go/beanstalkd/beanstalkd.go
```diff
@@ -23,9 +23,16 @@ type ProgressMessage struct {
 	Event        string       `json:"event"`
 }
 
+type StringMessage struct {
```
akotlar, line 26: We report progress about filtering back to the frontend through string messages because "progress" is expected to be monotonically increasing, and not well suited for jobs that are really composed of multiple stages. So our current solution is to send log messages that are stored in the db and shown in the browser.

go/cmd/dosage-filter/main.go
```diff
@@ -0,0 +1,316 @@
...
+
+	rowsAccumulated := 0
+
+	for index := range queue {
+		record, err := fr.RecordAt(index)
+		if err != nil {
+			log.Fatalf("Failed to read record at %d: %v", index, err)
+		}
+
+		locusCol := record.Column(0).(*array.String)
```
akotlar, line 129: the locus column is always the first column, see bystro-vcf; in future iterations we could generalize this, but I don't see the need since we control the format.

go/cmd/dosage-filter/main.go
```diff
@@ -0,0 +1,316 @@
...
+			if _, exists := loci[locusValue]; exists {
+				rowsAccumulated += 1
+				rowsAccepted += 1
+				// Fill the builder with values from the original columns, filtering by selectedIndices
+				for colIdx := 0; colIdx < int(record.NumCols()); colIdx++ {
+					column := record.Column(colIdx)
+					switch column := column.(type) {
+					case *array.String:
+						builder.Field(colIdx).(*array.StringBuilder).Append(column.Value(j))
+					case *array.Uint8:
```
akotlar, line 143: the first 2 cases cover everything we expect; however we support other formats here  because there is no harm in doing so

dlin30 (approved):
LGTM

PR #464, opened 2024-04-14
Bugfix marginal likelihood supervision + SVAE

austinTalbot7241993:
This combines two tasks into a single PR. We fix one of the formulas in the marginal supervision model, which can be seen in lines 166-168 of gf_marginal, where the wrong loss function for the predictive likelihood is used. We also implemented a linear Supervised Variational Autoencoder, which offers no practical advantages, but is an important comparison method in our paper.

### DIFF
--- a/python/python/bystro/supervised_ppca/_base.py
+++ b/python/python/bystro/supervised_ppca/_base.py
@@ -100,44 +100,44 @@ def _get_projection_matrix(W_: Tensor, sigma_: Tensor, device: Any):
     Cov = torch.linalg.inv(M) * sigma_
     return Proj_X, Cov
 
+
 def kl_divergence_vae(
     mu1: torch.Tensor,
     sigma1: torch.Tensor,
 ) -> torch.Tensor:
     """
-    Computes the Kullback-Leibler divergence between two multivariate 
+    Computes the Kullback-Leibler divergence between two multivariate
     normal distributions.
-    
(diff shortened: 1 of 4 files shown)


python/python/bystro/supervised_ppca/_base.py
```diff
@@ -101,6 +101,45 @@ def _get_projection_matrix(W_: Tensor, sigma_: Tensor, device: Any):
...
+
+    Returns:
+    - torch.Tensor: A tensor containing the KL divergence for each
+      instance in the batch, shape (batch_size,).
+    """
+    d = sigma1.shape[1]  # Dimensionality of the latent space
+    term1 = -1 * torch.logdet(sigma1)  # Log determinant
+    term2 = -d  # Negative of the latent space dimensionality
+    term3 = torch.trace(sigma1)  # Trace of the covariance matrix
+    term4 = torch.sum(torch.square(mu1), dim=1)  # Sum of squares
```
akotlar, line 138: Why do we do torch.sun(torch.square(mu1), dim=1) instead of mu1.T @ mu1 or torch.dot(mu1, mu1)?

austinTalbot7241993: If mu is    Nxp then mu1.T @mu1 gives a p x p matrix. Dimensions don't work for dot. Baiscally this is a nifty way of vectorizing evaluating a term with multiple data points.

akotlar (approved):
LGTM

PR #465, opened 2024-04-15
Covariance Module Improvements

austinTalbot7241993:
We are now realizing that fitting covariance matrices is a critical task and while there is extensive work in the statistics community on improving finite-sample properties of estimation, very little has migrated to other fields. Part of this is likely due to scattered implementations across a variety of github repositories. We are implementing a large quantity of these estimators ourselves with a common format similar to sklearn. Here I implement two shrinkage estimators by Ledoit and Wolfe.

### DIFF
--- a/python/python/bystro/covariance/_covariance_np.py
+++ b/python/python/bystro/covariance/_covariance_np.py
@@ -98,7 +98,24 @@ class BayesianCovariance(BaseCovariance):
 
 
 class LinearShrinkageCovariance(BaseCovariance):
+
     def fit(self, X: NDArray[np.float_]) -> "LinearShrinkageCovariance":
+        """
+        This fits a covariance matrix using the linear shrinkage approach 
+        to combine the empirical covariance matrix with a structured 
+        estimator.
+
+        Parameters
+        ----------
+        X : np.array-like, (n_samples, n_covariates)
+            The centered data
+
+        Returns
+        -------
+        self : LinearShrinkageCovariance
+            The model, with updated covariance attribute
+        """
+
         self._test_inputs(X)
         self.N, self.p = X.shape
 
@@ -109,7 +126,7 @@ class LinearShrinkageCovariance(BaseCovariance):
         lambd = np.mean(evalues)
 
         temp = [
-            np.linalg.norm(np.outer(X[:, i], X[:, i]) - S, "fro") ** 2
+            np.linalg.norm(np.outer(X[i], X[i]) - S, "fro") ** 2
             for i in range(self.N)
         ]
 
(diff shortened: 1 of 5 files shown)


python/python/bystro/covariance/_covariance_np.py
```diff
@@ -106,3 +95,146 @@ def _fill_prior_options(self, prior_options: dict[str, Any]) -> dict[str, Any]:
...
+
+    def fit(self, X: NDArray[np.float_]) -> "LinearShrinkageCovariance":
+        """
+        This fits a covariance matrix using the linear shrinkage approach 
+        to combine the empirical covariance matrix with a structured 
+        estimator.
+
+        Parameters
+        ----------
+        X : np.array-like, (n_samples, n_covariates)
```
akotlar, line 110: Let's add a note that we do not support missing values

python/python/bystro/covariance/_covariance_np.py
```diff
@@ -106,3 +95,146 @@ def _fill_prior_options(self, prior_options: dict[str, Any]) -> dict[str, Any]:
...
+    the ratio of variables p to n is significant.
+
+    The theoretical basis of this approach is built around the optimal 
+    estimation of p parameters, which balances between overfitting p^2
+    parameters and underfitting with only 1 parameter. The nonlinear shrinkage 
+    is calculated using an oracle estimator that depends on a sample spectral density 
+    function approximated via kernel estimation using the Epanechnikov kernel. 
+    This estimation facilitates the derivation of shrinkage intensities for each 
+    eigenvalue, tailoring the adjustment to improve estimations under various 
+    data conditions. Practical implementation details are provided, adapted from a
```
akotlar, line 166: Let's remove this portion:

```
Practical implementation details are provided, adapted from a 
    Matlab script by the authors, ensuring that users can apply this method effectively 
    to empirical data sets. The docstring also underscores the utility of this estimator 
    in practical settings, providing a more parameter-efficient approach compared 
    to traditional methods, which often rely on unfeasibly large parameter sets.
    ```

File: python/python/bystro/covariance/_covariance_np.py (hunk 1 above)
Comment (austinTalbot7241993) [reply python/python/bystro/covariance/_covariance_np.py:110]: Where do you want to add it?

Comment (akotlar) [reply python/python/bystro/covariance/_covariance_np.py:109]: I was just thinking right here: "The centered data. Must not contain any missing values (np.nan)"

File: python/python/bystro/covariance/_covariance_np.py [hunk 3]
```diff
@@ -106,3 +95,146 @@ def _fill_prior_options(self, prior_options: dict[str, Any]) -> dict[str, Any]:
...
+        """
+        This fits a covariance matrix using the linear shrinkage approach 
+        to combine the empirical covariance matrix with a structured 
+        estimator.
+
+        Parameters
+        ----------
+        X : np.array-like, (n_samples, n_covariates)
+            The centered data
+
```
akotlar, line 111: Let's also add a Raises section:

```py
Raises
------
ValueError:
     A value error will be raised if missing data is found in X ( np.isnan(X) evaluates to true ), or if X is not an NDArray
```

akotlar: This follows numpy conventions: https://numpydoc.readthedocs.io/en/latest/format.html#raises

akotlar (approved):
LGTM

PR #467, opened 2024-04-16
Soft impute CV

austinTalbot7241993:
This implements an object to automatically cross-validate over the regularization strength in the soft impute algorithm and use the best value to impute the missing data.

### DIFF
--- a/python/python/bystro/imputation/soft_impute_cv.py
+++ b/python/python/bystro/imputation/soft_impute_cv.py
@@ -31,6 +31,7 @@ def nan_with_probability(X, p, rng):
         a boolean array where True indicates a value was set to
         NaN by this function.
     """
+    X = X.copy()
     # Create a mask of the same shape as X, initially all False
     mask = np.zeros(X.shape, dtype=bool)
     # Find indices of non-NaN values in X
--- a/python/python/bystro/imputation/soft_impute_cv.py
+++ b/python/python/bystro/imputation/soft_impute_cv.py
@@ -1,0 +1,144 @@
+from typing import Any, Dict
+import numpy as np
+
+from bystro.imputation.fancyimpute.soft_impute import SoftImpute
+
+
+def nan_with_probability(X, p, rng):
+    """
+    Takes an input array X and a float p.
+    Of the values in X that are not NaN, sets the elements to NaN with
+    probability p. Returns the modified array X and a mask indicating
+    which values were set to NaN.
+
+    Parameters
+    ----------
+    X: np.array,
+        the input array possibly containing NaNs.
+
+    p: float,
+        probability with which a non-NaN element of X is set to NaN.
+
+    rng : numpy random generator
+        The random number generator
+
+    Returns
+    -------
+    X_modified: np.array,
+        the modified array with additional NaNs.
+
+    mask: np.array,
+        a boolean array where True indicates a value was set to
+        NaN by this function.
+    """
+    # Create a mask of the same shape as X, initially all False
+    mask = np.zeros(X.shape, dtype=bool)
+    # Find indices of non-NaN values in X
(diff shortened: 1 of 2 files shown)


python/python/bystro/imputation/soft_impute_cv.py
```diff
@@ -0,0 +1,144 @@
...
+    # Determine which values to set to NaN based on probability p
+    to_nan = random_numbers < p
+    # Apply the decision to X and mask
+    X[non_nan_indices[0][to_nan], non_nan_indices[1][to_nan]] = np.nan
+    mask[non_nan_indices[0][to_nan], non_nan_indices[1][to_nan]] = True
+    return X, mask
+
+
+class SoftImputeCV:
+    def __init__(
```
akotlar, line 49: Could we document these init arguments?

python/python/bystro/imputation/soft_impute_cv.py
```diff
@@ -0,0 +1,144 @@
+from typing import Any, Dict
+import numpy as np
+
+from bystro.imputation.fancyimpute.soft_impute import SoftImpute
+
+
+def nan_with_probability(X, p, rng):
```
akotlar, line 7: Do we want to copy X within this method to avoid mutation at the function level rather than requiring the user to X.copy()?

Else, let's document that this mutates the input X.

akotlar (changes requested):
Document X mutation in nan_with_probaiblity

python/python/bystro/imputation/soft_impute_cv.py
austinTalbot7241993: Oh we definitely want to change that, didn't realize it was mutated.

python/python/bystro/imputation/soft_impute_cv.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def fit_transform(self, X):
+        rng = np.random.default_rng(self.seed)
+        regs = np.logspace(-2, 2, self.Cs)  # Regularization strengths
+
+        imputation_var = np.zeros((self.Cs, self.k_fold))
+
+        for i in range(self.Cs):
+            for j in range(self.k_fold):
+                X_modified, mask = nan_with_probability(
```
akotlar, line 73: We don't need to run nan_with_probability in this loop, unless I'm missing something:


```python
def fit_transform(self, X):
        rng = np.random.default_rng(self.seed)
        regs = np.logspace(-2, 2, self.Cs)  # Regularization strengths

        imputation_var = np.zeros((self.Cs, self.k_fold))

        X_modified, mask = nan_with_probability(X.copy(), self.prob_holdout, rng)
        vals_original = X[mask]
        

        for i in range(self.Cs):
            for j in range(self.k_fold):

                 model_si = SoftImpute(
                    shrinkage_value=regs[i],
                    training_options=self.training_options,
                )
                X_complete_si = model_si.fit_transform(X_modified)

                vals_imputed_si = X_complete_si[mask]

                random_guess_mse = np.mean(
                    (vals_original - np.mean(vals_original)) ** 2
                )
                si_mse = np.mean((vals_original - vals_imputed_si) ** 2)
                imputation_var[i, j] = 1 - si_mse / random_guess_mse

        avg_var = np.mean(imputation_var, axis=1)
        opt_val = regs[avg_var == np.amax(avg_var)]
        model_si = SoftImpute(
            shrinkage_value=opt_val, training_options=self.training_options
        )
        X_complete = model_si.fit_transform(X)

        return X_complete
```

austinTalbot7241993: That's an error on my part. nan_with_probability should be within the loop. Will fix.

akotlar:
LGTM

akotlar (approved):

PR #470, opened 2024-04-17
Improved initialization scheme

austinTalbot7241993:
With regression case, we get dramatic performance improvements by initializing the predictive coefficients to be reasonable, based on the initialization of the loadings W. This incorporates that change.


akotlar (approved):
LGTM

akotlar (approved):
LGTM

PR #475, opened 2024-04-25
Add back support for nulls in fast dosage matrix calculation

akotlar:
* Add back support for nulls
* Change final message about annotation filtering to reflect the fact that we don't actually know the total number of rows, since we only read the file once.


akotlar:
The commit history looks crazy because I merged upstream back into this branch, but the diff is correct.

austinTalbot7241993 (approved):
Looks good to me.

PR #477, opened 2024-04-26
[search] Within query slice, paginate 

akotlar:
* Paginate within each query slice, to ensure we fetch all records
* Make startup.yml depend on opensearch.yml, not elastic-config2.yml, which is not checked into the repo
* Move opensearch.yml to opensearch.clean.yml, to make it clearer that it is not a full configuration and must be edited
* Remove dependence of proteomics tests on opensearch.yml
* Rely on public bystro-ancestry-public bucket for ancestry models by default
* Increases logging of queries (how many records fetched per slice)


python/python/bystro/search/save/handler.py
```diff
@@ -32,7 +32,9 @@
 logger = logging.getLogger(__name__)
 
 MAX_QUERY_SIZE = 10_000
-MAX_SLICES = 1e6
+# TODO 2024-04-26 @akotlar Investigate the impact of large numbers of slices
+# and whether 20_000 slices is a reasonable limit
+MAX_SLICES = 20_000
```
akotlar, line 37: 20000 slices gives us up to 170 million variant queries before a single slice must scroll. I am not confident there aren't drawbacks to enormous numbers of slices, and we don't actually need 1 slice per 10,000 variants anymore. As the TODO says, we will need to test this.

dlin30 (approved):
LGTM

PR #478, opened 2024-04-27
[annotation] Add isCanonical transcript annotation

akotlar:
* Creates isCanonical transcript feature, which will be outputted as 'TRUE' or 'FALSE' depending on whether there is a matching UCSC knownCanonical feature

We output 'TRUE', 'FALSE' here instead of 1/0, because pandas, polars (the most popular data science libraries in python), R standard dataframe library, and opensearch will interpret this as the boolean True, or False. We also use TRUE, FALSE for discordant for similar reasons. In the case of refSeq this isn't as useful because we often have multiple transcripts, so I'm open to going back to using 0/1.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -95,47 +95,41 @@ Once Bystro is installed, it needs to be configured. The easiest step is choosin
 
 1. Download the Bystro database for your species/assembly
 
-- **Example:** hg38 (human reference GRCh38): `wget https://s3.amazonaws.com/bystro-db/hg38.tar.gz`</strong>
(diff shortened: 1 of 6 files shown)


wingolab (approved):
might want to keep the specific paths out of the repo but otherwise looks good to me.

akotlar:
> might want to keep the specific paths out of the repo but otherwise looks good to me.

Good point. I’ve started including them because it makes setup easier than having a “clean” database file, but this isn’t the right solution. I will write a cli tool to download a Bystro database and configure the YAML config with the appropriate paths.

PR #479, opened 2024-04-28
Incorporated basic shrinkage to adversarial model

austinTalbot7241993:
This adds regularization into the basic adversarial model. This allows for it to effectively learn latent representations with smaller sample sizes, important in many of our applications.

### DIFF
--- a/python/python/bystro/supervised_ppca/ppca_adversarial.py
+++ b/python/python/bystro/supervised_ppca/ppca_adversarial.py
@@ -11,12 +11,17 @@ latent variables. Two key parameters, 'mu' and 'eps', control the strength
 of the adversarial component and the conditioning of inverses, 
 respectively, enhancing the robustness and stability of the model.
 """
-
+from typing import Union
 import numpy as np
 import numpy.linalg as la
 
 from bystro.supervised_ppca._base import BaseGaussianFactorModel
 from numpy.typing import NDArray
+
+from bystro.covariance._covariance_np import (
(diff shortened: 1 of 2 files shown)


python/python/bystro/supervised_ppca/ppca_adversarial.py
```diff
@@ -89,17 +106,62 @@ def fit(
...
+        model_cov: Union[
+            LinearShrinkageCovariance, NonLinearShrinkageCovariance
+        ]
+        if regularization_options["method"] == "Empirical":
+            B_11 = X_dm.T @ X_dm
+            B_12 = X_dm.T @ Y_dm
+            diag_reg = self.eps * np.eye(X_dm.shape[1])
+            XtX = X_dm.T @ X_dm + diag_reg
+            B_22 = B_12.T @ la.solve(XtX, B_12)
+
```
akotlar, line 118: unnecessary space, I'll fix

python/python/bystro/supervised_ppca/ppca_adversarial.py
```diff
@@ -11,13 +11,18 @@
 of the adversarial component and the conditioning of inverses, 
 respectively, enhancing the robustness and stability of the model.
 """
-
+from typing import Union
```
akotlar, line 14: we don't really want/need to import union; the modern way to do this is to use a pipe character `int | None`

akotlar (approved):
LGTM

PR #482, opened 2024-05-03
[search] Add back autocomplete to CLNSIG

akotlar:
* CLNSIG currently does not use an autocomplete filter, which means "patho" will not match any terms in CLNSIG. With this change "patho" and "pathogenic" will both match `pathogenic` and `likely_pathogenic`. While this slightly decreases specificity (because it now matches "likely_pathogenic", that can always be added back using quotes: "pathogenic"

### DIFF
--- a/config/hg19.mapping.yml
+++ b/config/hg19.mapping.yml
@@ -129,14 +129,6 @@ index_settings:
           - catenate_filter_split
           - english_stemmer
           - autocomplete_filter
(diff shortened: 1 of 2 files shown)


akotlar:
This actually has some drawbacks currently, closing for now.

Namely, this will match conflicting_interpretations_of_pathogenicity, which we don't really want we because that isn't really a pathogenic label at all (it can correspond to benign sites)

PR #485, opened 2024-05-04
Supervision Augmented PPCA

austinTalbot7241993:
This implements the remaining model of the augmented PCA package, which allows for supervision of the eigenvector decomposition for finding predictive networks. It also allows for extra covariance matrix estimation methods. Finally, it implements a novel method for removing some information and encouraging other information in the latent representation.


akotlar (approved):
Tests pass, code changes seem sensible and come from existing package with license. LGTM

PR #486, opened 2024-05-05
[search] Improve clnsig analysis

akotlar:
* improve clinvarVcf.CLNSIG analysis, so that we can autocomplete pathogen, without matching conflicting_classifications_of_pathogenicity. This is accomplished by blocking terms that start with “conflicting” and applying edge ngram transform in all other values. 

This gives a good balance between accuracy and sensitivity.

With this, `clinvarVcf.CLNSIG:pathogenic` matches `likely_pathogenic` and `pathogenic`, even if the terms are found comma separated. However "conflicting_classifications_of_pathogenicity" will not match, if that is the only term ("conflicting_classifications_of_pathogenicity,pathogenic" would match however).

Live on bystro-dev

Test queries:
1. `other` matches anything with `other` as a complete term (meaning conflicting_classification_of_pathogenicity,other matches)
2. `pathogenic` https://bystro-dev.emory.edu/results?_id=6637a7caa0e17a1660ba743b&search&q=clinvarvcf.clnsig:pathogenic&size=10&from=0
3. `likely_pathogenic` https://bystro-dev.emory.edu/results?_id=6637a7caa0e17a1660ba743b&search&q=clinvarvcf.clnsig:likely_pathogenic&size=10&from=0
4. `(likely pathogenic)` https://bystro-dev.emory.edu/results?_id=6637a7caa0e17a1660ba743b&search&q=clinvarvcf.clnsig:(likely%20pathogenic)&size=10&from=0
5. `pa`, `pat`, `path`, `pathogen`, etc, all match the same set, because only "pathogenic" starts with these characters: https://bystro-dev.emory.edu/results?_id=6637a7caa0e17a1660ba743b&search&q=clinvarvcf.clnsig:pathog&size=10&from=0
6. `like`, `likel`, `likely` etc., all match terms starting with `likely`, such as `likely_pathogenic` and `likely_benign`: https://bystro-dev.emory.edu/results?_id=6637a7caa0e17a1660ba743b&search&q=clinvarvcf.clnsig:likely&size=10&from=0

### DIFF
--- a/config/hg19.mapping.yml
+++ b/config/hg19.mapping.yml
@@ -28,9 +28,24 @@ index_settings:
           - uppercase
           - asciifolding
     filter:
-      exclude_pathogenic:
+      clnsig_conditional_underscore_split:
+        type: condition
+        filter: ["clnsig_underscore_split_filter"]
+        script:
+          source: "token.getTerm().toString().startsWith('conflicting') == false"
+      clnsig_conditional_filter:
+        type: condition
+        filter: ["clnsig_edge_ngram_filter"]
+        script:
+          source: "token.getTerm().toString().startsWith('conflicting') == false"
+      clnsig_edge_ngram_filter:
+        type: edge_ngram
+        min_gram: 1
+        max_gram: 20
+      clnsig_underscore_split_filter:
         type: pattern_capture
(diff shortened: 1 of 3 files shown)


dlin30 (approved):
LGTM

PR #487, opened 2024-05-06
Add compressed versions of AD summary stats in hg19 and hg38

cristinaetrv:
Previously, AD summary stats were in uncompressed feather files and were giving CRLF to LF warnings. These versions are compressed feather files and .gitattributes has also been updated to mark feather files as binary so that github is aware of the filetype and treats them accordingly.


akotlar (approved):
LGTM

PR #489, opened 2024-05-07
Add LD maps for hg38 and organize LD maps by build for PRS.

cristinaetrv:
Add hg38 CEU LD map for testing initial version of PRS C+T. Currently these correspond to bins of size 1cM for CEU LD, but clumping will be flexible in future versions and LD maps may be relocated to somewhere outside this repo.


akotlar (approved):
lgtm

PR #490, opened 2024-05-07
[annotation] Remove duplicate entries due to isCanonical

akotlar:
* Our isCanonical field addition to refSeq track was done using a LEFT JOIN, which resulted in occasional duplicate refSeq transcript entires. This PR solves that by removing the reliance on LEFT JOIN, and instead take the maximum value ("true" > "false) from a list of matching isCanonical possibilities.

#### Justification:
We annotate against refGene transcripts, which are identified by refGene.name. isCanonical indicates whether a knownGene transcript is canonical or not. For each of our refSeq track (refGene UCSC mysql table) transcripts, we list every knownGene transcript that matches (kgXref.mrna = refGene.name). 

When there are multiple knownGene transcripts for a single refGene transcript, isCanonical either has to be an array (one per knownGene entry), which will be hard to make use of, or a single entry that is "true" if any members of the array is "true". I've chosen the latter, because I think it will be more useful


wingolab (approved):
The main change is the coalesce and max approach to parsing of the knownCanonical table - i.e., `COALESCE((SELECT MAX(CASE WHEN k.transcript IS NOT NULL THEN 'true' ELSE 'false' END) FROM knownCanonical k WHERE k.transcript IN (SELECT kgID FROM kgXref x WHERE x.refseq=r.name)), 'false') AS isCanonical`. It seems reasonable to me.

PR #493, opened 2024-05-08
Optimize PRS C+T preprocess for performance.

cristinaetrv:
Optimizes the C+T preprocess by performing clumping and thresholding on the gwas summary stat scores using only the overlap of loci between the locus column from the dosage matrix rather than including the entire datasets in preprocessing steps.

### DIFF
--- a/python/python/bystro/prs/preprocess_for_prs.py
+++ b/python/python/bystro/prs/preprocess_for_prs.py
@@ -15,9 +15,11 @@ logger = logging.getLogger(__name__)
 
 AD_SCORE_FILEPATH = "gwas_summary_stats/AD_sumstats_hg38_PMID35379992.feather"
 
+
 class StrEnum(str, Enum):
     def __str__(self):
         return self.value
+
 
 class GwasStatsLocusKind(StrEnum):
     MATCH = "Direct Match"
@@ -83,17 +85,17 @@ def _extract_nomiss_dosage_loci(dosage_matrix_path: str) -> pd.DataFrame:
     pd.DataFrame: A DataFrame with a single 'locus' column containing loci with no missing
     dosage values across all other columns.
     """
-    dataset = ds.dataset(dosage_matrix_path, format='feather')
+    dataset = ds.dataset(dosage_matrix_path, format="feather")
     conditions = []
     for column in dataset.schema.names:
-        if column != 'locus':
-            condition = (ds.field(column) != -1)
+        if column != "locus":
+            condition = ds.field(column) != -1
             conditions.append(condition)
     combined_condition = conditions[0]
     for condition in conditions[1:]:
         combined_condition = combined_condition & condition
     filtered_dataset = dataset.filter(combined_condition)
-    filtered_loci = filtered_dataset.to_table(columns=['locus']).to_pandas()
+    filtered_loci = filtered_dataset.to_table(columns=["locus"]).to_pandas()
     return filtered_loci
 
 
@@ -163,18 +165,17 @@ def _preprocess_genetic_maps(map_directory_path: str) -> dict[int, list[int]]:
 
 
 def read_feather_in_chunks(file_path, columns=None, chunk_size=1000):
-    # Read only the necessary columns to filter by P-value
+    """Read a Feather file in chunks as pandas Dataframes."""
     table = feather.read_table(file_path, columns=columns)
     for i in range(0, table.num_rows, chunk_size):
         chunk = table.slice(i, chunk_size).to_pandas()
-        # Ensure the chunk maintains its original index
         chunk.index = range(i, i + len(chunk))
         yield chunk
 
 
 def get_p_value_thresholded_indices(df, p_value_threshold):
     """Return indices of rows with P-values less than the specified threshold."""
-    return df.index[df['P'] < p_value_threshold].tolist()
+    return df.index[df["P"] < p_value_threshold].tolist()
 
 
 def generate_thresholded_overlap_scores_dosage(
@@ -183,7 +184,7 @@ def generate_thresholded_overlap_scores_dosage(
     """Compare and restrict to overlapping loci between dosage matrix and thresholded scores."""
     scores = _preprocess_scores(gwas_scores_path)
     dosage_loci_nomiss = _extract_nomiss_dosage_loci(dosage_matrix_path)
-    
+
     # TODO: Add customizable p value threshold and option for multiple thresholds
     p_value_threshold = 0.05
     thresholded_indices_set = set()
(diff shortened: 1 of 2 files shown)


python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -73,6 +71,34 @@ def _load_genetic_maps_from_feather(map_directory_path: str) -> dict[str, pd.Dat
...
+    pd.DataFrame: A DataFrame with a single 'locus' column containing loci with no missing
+    dosage values across all other columns.
+    """
+    dataset = ds.dataset(dosage_matrix_path, format="feather")
+    conditions = []
+    for column in dataset.schema.names:
+        if column != "locus":
+            condition = ds.field(column) != -1
+            conditions.append(condition)
+    combined_condition = conditions[0]
```
akotlar, line 94: This make sense as a first step; could @austinTalbot7241993 and you work together on an imputation strategy (maybe SoftImputeCV?) for future iterations?

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -125,38 +165,51 @@ def _preprocess_genetic_maps(map_directory_path: str) -> dict[int, list[int]]:
     return bin_mappings
 
 
-def filter_scores_by_p_value(scores: pd.DataFrame, p_value_threshold: float) -> pd.DataFrame:
-    """Filter to keep rows with P-values less than the specified threshold for C+T method."""
-    return scores[scores["P"] < p_value_threshold]
+def read_feather_in_chunks(file_path, columns=None, chunk_size=1000):
+    """Read a Feather file in chunks as pandas Dataframes."""
+    table = feather.read_table(file_path, columns=columns)
```
akotlar, line 170: Neat! This is way better than what I was doing before (which was lower level using dataset api)

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -125,38 +165,51 @@ def _preprocess_genetic_maps(map_directory_path: str) -> dict[int, list[int]]:
...
-    """Compare and restrict to overlapping loci between dosage matrix and scores."""
+) -> pd.DataFrame:
+    """Compare and restrict to overlapping loci between dosage matrix and thresholded scores."""
     scores = _preprocess_scores(gwas_scores_path)
-    dosage_feather = _load_preprocessed_dosage_matrix(dosage_matrix_path)
+    dosage_loci_nomiss = _extract_nomiss_dosage_loci(dosage_matrix_path)
+
     # TODO: Add customizable p value threshold and option for multiple thresholds
-    thresholded_scores = filter_scores_by_p_value(scores, 0.05)
+    p_value_threshold = 0.05
```
akotlar, line 190: Reading in row and column chunks, very nice

python/python/bystro/prs/preprocess_for_prs.py
cristinaetrv: Maybe it makes sense to do imputation on the dosage matrix upstream of analyses overall rather than per analysis since we're imputing for ancestry as well? Would it make sense to have a version of the dosage matrix with imputation even as the dosage matrix is generated?

python/python/bystro/prs/tests/test_preprocess_for_prs.py
```diff
@@ -171,38 +187,6 @@ def test_filter_scores_by_p_value(mock_processed_scores_df: pd.DataFrame):
     ), "Filtered scores should contain expected SNP(s)."
```
cristinaetrv, line 189: Planning on adding back a new test for this function in future once the dust has settled on optimizing this code, for now all the changes made it really difficult to get the mock files to work for this test so I opted to take it out while changes are still being made. Tested that all this code works on a small dataset though so the only problems are related to mocking it correctly.

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -125,38 +164,60 @@ def _preprocess_genetic_maps(map_directory_path: str) -> dict[int, list[int]]:
...
+    thresholded_scores = thresholded_scores.set_index("SNPID")
+
+    if len(thresholded_scores) == 0:
+        raise ValueError("No thresholded scores available; cannot proceed with PRS calculation.")
+
     set_A = set(thresholded_scores.index)
-    set_B = set(dosage_feather["locus"])
+    set_B = set(dosage_loci_nomiss["locus"])
+
     overlap_snps = set_A.intersection(set_B)
```
akotlar, line 204: Instead of doing this, add an optional argument to `_extract_nomiss_dosage_loci`: `desired_loci`. Pass in the set of loci, which are `set(threshold_scores.index)`, and make one of the filter conditions the selection of just those loci. You will further reduce memory usage.

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -172,11 +233,10 @@ def compare_alleles(row: pd.Series, col1: str, col2: str) -> GwasStatsLocusKind:
...
 
 
-def adjust_dosages(row: pd.Series) -> pd.Series:
-    """Adjust the dosage for if the effect allele is not the alternate allele."""
-    if row["allele_comparison"] == "effect allele is ref":
+def adjust_dosages(row):
+    if row["allele_comparison"] == "reversed match":
         for col in row.index:
-            if col != "allele_comparison":
+            if col != "allele_comparison" and row[col] != -1:
```
akotlar, line 239: I would leave this a bit more general to work with null missing values as well: https://github.com/bystrogenomics/bystro/blob/master/python/python/bystro/ancestry/inference.py#L276

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -209,64 +267,78 @@ def calculate_abs_effect_weights(merged_scores_genos: pd.DataFrame) -> pd.DataFr
...
+        compare_alleles, col1="SNPID", col2="ID_effect_as_alt", axis=1
+    )
+    scores_overlap.insert(0, "allele_comparison", allele_comparison_results)
+    scores_overlap_w_bins = assign_bins(scores_overlap, bin_mappings)
+    scores_overlap_abs_val = calculate_abs_effect_weights(scores_overlap_w_bins)
+    max_effect_per_bin = select_max_effect_per_bin(scores_overlap_abs_val)
+    return clean_scores_for_analysis(max_effect_per_bin, "ID_effect_as_ref")
+
+
+def extract_clumped_thresholded_genos(
```
akotlar, line 305: this will not work as written for large dataframes because even for the subset of loci, we will be unable to always load this into memory; we'll want to perform PRS on groups of samples, and save out just the PRS scores per sample

akotlar (changes requested):
See comments; primarily we need to adjust  `_extract_nomiss_dosage_loci`

python/python/bystro/prs/preprocess_for_prs.py
akotlar: This would be good, but not important for now.

akotlar (approved):
Approved; see comment on extract_clumped_thresholded_genos

PR #494, opened 2024-05-08
fix get_user cli and api commands

akotlar:
* fix get_user cli and api commands
* fix query api
* add basic api documentation

### DIFF
--- a/python/python/bystro/api/auth.py
+++ b/python/python/bystro/api/auth.py
@@ -87,7 +87,7 @@ class UserProfile(Struct, rename="camel"):
     email: str
     accounts: list[str]
     role: str
-    last_login: datetime.datetime
+    last_login: datetime.datetime | None = None
 
 
 def _fq_host(host: str, port: int | None = None) -> str:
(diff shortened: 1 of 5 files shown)


python/python/bystro/api/auth.py
```diff
@@ -87,7 +87,7 @@ class UserProfile(Struct, rename="camel"):
     email: str
     accounts: list[str]
     role: str
-    last_login: datetime.datetime
+    last_login: datetime.datetime | None = None
```
akotlar, line 90: When the user first signs up, there is no last_login property

python/python/bystro/cli/cli.py
```diff
@@ -109,7 +109,7 @@ def create_jobs_cli(args: argparse.Namespace) -> list[dict]:
     )
 
 
-def get_user_cli() -> UserProfile:
+def get_user_cli(_args: argparse.Namespace) -> UserProfile:
```
akotlar, line 112: even though we don't use the argument, it is needed, since argparse passes it

cristinaetrv (approved):
Lgtm!

PR #495, opened 2024-05-09
Adding Documentation to deploy annotator

dlin30:
- Added an amazon 2023 specific installation guide for bystro's annotator with FAQ with problems i ran into
- Updated scripts with updated versions and other dependencies to prevent parts of the script to fail


akotlar:
Will close this for now, as the work is now outdated, but I will use some of the README.md changes. Thanks Dennis!

PR #496, opened 2024-05-11
[infrastructure] Improve Python build process

akotlar:
* Rely on pyproject.toml instead of requirements.txt
* Update dependencies to ensure that we are compatible with M1+ Macs
* Create CI workflow to build and publish wheels to PyPI on Mac, Linux, and Windows (Windows support is limited and experimental, as noted in INSTALL.md). Mac supports aarch64 and x86_64. Linux is currently x86_64, as is Windows.


python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -219,6 +219,11 @@ def select_max_effect_per_bin(merged_scores_genos: pd.DataFrame) -> pd.DataFrame
 def clean_dosage_for_analysis(merged_scores_genos: pd.DataFrame, column_to_drop: str) -> pd.DataFrame:
     """Drop missing values and adjust scores."""
     column_to_drop_index = merged_scores_genos.columns.get_loc(column_to_drop)
+
+    # If there is more than 1 column found, our "columns_to_keep" function will not work
+    if not isinstance(column_to_drop_index, int):
```
akotlar, line 224: This was needed because get_loc doesn't actually return an int unless the column value being searched is unique. When I updated the pandas stubs, mypy began to complain about this, which is pretty neat, statically preventing errors!

python/pyproject.toml
```diff
@@ -1,20 +1,46 @@
 [build-system]
-requires = ["maturin>=1.3.0,<1.4.0", "setuptools", "wheel", "Cython"]
+requires = ["maturin>=1.5.1", "setuptools", "wheel", "Cython"]
 build-backend = "maturin"
 
 [project]
 name = "bystro"
-requires-python = ">=3.10"
+requires-python = ">=3.11.0"
```
akotlar, line 7: We could lower this back to 3.10, but I would argue we shouldn't unless there is a user need: python 3.11 is notably faster than python 3.10, introduces some useful syntax, and is now a year old. Additionally all of our CI is against 3.11, so we haven't actually been testing 3.10.

README.md
```diff
@@ -20,9 +20,31 @@ The web app gives full access to all of Bystro's capabilities, provides a conven
 
 ## Installing Bystro
 
-The easiest way is to run from Docker: `docker pull akotlar/bystro:latest && docker run bystro:latest bystro-annotate.pl`
+Bystro consists of 2 main components: the Bystro Python package, which consists of the Bystro ML library, CLI tool, and a collection of easy to use biology methods like globaly ancestry, and the Bystro annotator (Perl).
```
cristinaetrv, line 23: Found a typo and a couple other suggestions:
```suggestion
Bystro consists of 2 main components: the Bystro Python package, which consists of the Bystro ML library, CLI tool, and a collection of easy to use biology tools including global ancestry and the Bystro annotator (Perl).
```

python/pyproject.toml
```diff
@@ -1,20 +1,46 @@
...
 classifiers = [
     "Programming Language :: Rust",
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
-version = "1.0.0"
+version = "1.0.3"
+dependencies = [
```
cristinaetrv, line 14: Does moving the dependencies here fix the error where requirements.txt wasn't being automatically installed?

python/python/bystro/prs/preprocess_for_prs.py
cristinaetrv: I don't believe this exists in the prs code anymore, but that is pretty neat

cristinaetrv (approved):
Lgtm! The instructions seem more clear and the process more straightforward. I just had one small suggestion for the README but otherwise looks good to go.

python/python/bystro/prs/preprocess_for_prs.py
akotlar: it's still there, you renamed it though `format_col_index`

PR #498, opened 2024-05-14
Add Euclidean and Riemann median calculation functions

IlhaH:
Adds Euclidean median function with tests and initial attempt at Riemann median function with tests


austinTalbot7241993 (approved):
LGTM

PR #499, opened 2024-05-14
[infrastructure] make canopy direct dependency until SomaLogic makes pypi package

akotlar:
* Fixes release creation, which was failing because we rely on Canopy, which does not have a pypi package. To solve this we update our instructions to inform users that they for now need to install canopy
* fixes perltidy error, which for some reason was only see on release
* fixes how liftover is imported, so we do not query the UCSC server unless necessary
* updates liftover dependency, so that it can use alternate UCSC mirrors
* adds missing ancestry api test

This solution is temporary; once SomaLogic makes a pypi package, we will depend on that instead: https://github.com/SomaLogic/Canopy/issues/6. If they choose not to, we will publish our own.


python/pyproject.toml
```diff
@@ -10,11 +10,12 @@ classifiers = [
...
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
-version = "1.0.3"
+version = "2.0.0-beta11"
 dependencies = [
+    "openpyxl==3.1.2",
     "attrs==23.1.0",
     "boto3==1.28.9",
-    "liftover==1.1.17",
+    "liftover==1.2.2",
```
akotlar, line 18: fixes issue with the wrong UCSC mirror being used

IlhaH (approved):

PR #503, opened 2024-05-16
fix install to reflect the fact that we're still using beta tags

akotlar:
* We need to use --pre to install pip packages with beta tags. While we will release a production Bystro 2.0 this summer, we are not there yet, so we need to use --pre.


austinTalbot7241993 (approved):
LGTM

PR #504, opened 2024-05-17
Feature/issue 421 better proteomic filtering

akotlar:
* Add ability to get annotation dataframe based on any field


python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -48,93 +42,182 @@ def _flatten(xs: Any) -> list[Any]:  # noqa: ANN401 (`Any` is really correct her
...
+            value = value[key]
+        except (KeyError, TypeError):
+            return None  # Returns None if the field path is not found
+    return value
 
 
-def _get_samples_genes_dosages_from_hit(hit: dict[str, Any]) -> pd.DataFrame:
-    """Given a document hit, return a dataframe of samples, genes and dosages."""
+def _get_samples_genes_dosages_from_hit(
+    hit: dict[str, Any], additional_fields: list[str] | None = None
```
cristinaetrv, line 58: With additional fields, can you pull any field you would like or are there any limitations?

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -48,93 +42,182 @@ def _flatten(xs: Any) -> list[Any]:  # noqa: ANN401 (`Any` is really correct her
...
+                    "is_canonical": is_canonical[gene_idx],
+                    "dosage": dosage_label,
+                    "heterozygosity": heterozygosity,
+                    "homozygosity": homozygosity,
+                    "missingness": missingness,
+                    "gnomad_genomes_af": gnomad_genomes_af,
+                    "gnomad_exomes_af": gnomad_exomes_af,
                 }
-            )
+                # Add additional fields
```
cristinaetrv, line 158: More of a stylistic question, but is there a reason why you wouldn't want to handle all fields using the code below instead of specifying each one of the defaults? Is it so that you can change the name of the fields to something friendlier to read/interpret or a different reason?

cristinaetrv (approved):
Lgtm!

python/python/bystro/proteomics/annotation_interface.py
akotlar: no limitations; just if you provide fields that are already in the defaults, the defaults would be removed from the additional fields

python/python/bystro/proteomics/annotation_interface.py
akotlar: We could specify a list of default "fields to add"; we just wouldn't be able to rename them as we do above. I'm not committed to the renaming; do you think it is useful?

PR #505, opened 2024-05-17
Add POIROT methods, testing file, and pickle file with test data

IlhaH:
Implemented POIROT methods for detecting parent-of-origin effects. Created a testing script to validate the POIROT methods, asserting that the output is the same as the original R code from POIROT-POE


austinTalbot7241993 (approved):
LGTM

PR #506, opened 2024-05-17
Implement geodesic functions and add unit tests with expected results

IlhaH:
Implement geodesic functions from pyRiemann.
Add test cases to compare custom implementations with pyriemann's geodesic functions, test different alpha values and ensure results match expected outputs.


austinTalbot7241993 (approved):
LGTM

PR #507, opened 2024-05-20
Add adjustable p value thresholds for PRS C+T.

cristinaetrv:
Previously, the p-value threshold was hardcoded for early testing. Now it can take in p-value thresholds as floats between 0 and 1.

### DIFF
--- a/python/python/bystro/prs/tests/test_preprocess_for_prs.py
+++ b/python/python/bystro/prs/tests/test_preprocess_for_prs.py
@@ -28,7 +28,7 @@ def mock_scores_df() -> pd.DataFrame:
             "POS": [566875, 728951],
             "OTHER_ALLELE": ["C", "G"],
             "EFFECT_ALLELE": ["T", "A"],
-            "P": [0.699009, 0.030673],
+            "P": [0.699009, 0.0030673],
             "BETA": [0.007630, -0.020671],
         }
     )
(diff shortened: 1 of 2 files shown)


python/python/bystro/prs/tests/test_preprocess_for_prs.py
```diff
@@ -28,7 +28,7 @@ def mock_scores_df() -> pd.DataFrame:
             "POS": [566875, 728951],
             "OTHER_ALLELE": ["C", "G"],
             "EFFECT_ALLELE": ["T", "A"],
-            "P": [0.699009, 0.030673],
+            "P": [0.699009, 0.0030673],
```
akotlar, line 31: These no longer add up to 1; is that intended?

akotlar: And as a follow-on, if this is due to p-value thresholding; should we renormalize?

cristinaetrv: They're p-values for the two separate loci, so they don't need to add up to anything specific but I wanted the test to have a separate p-value threshold from the default so adjusted the mock data to reflect the new threshold I was testing p < 0.005.

akotlar (approved):
LGTM

PR #508, opened 2024-05-21
[api, proteomics] Proxy OpenSearch Requests

akotlar:
* Proxy OpenSearch requests, allowing users to directly query the OpenSearch cluster, even outside the cluster, by way of the Bystro API server

This will enable useful telemetry, while still ensuring adequate security. Not all requests are handled, but all the requests we need for fetching annotation data in bulk, and joining on proteomic data, without needing to go through the queue.

This makes our proteomic filtering functionality directly useful, and opens the door to enable users to programmatically do other things that require the search engine, like filtering annotations, without requiring batch job submissions through the queue. This will probably be popular for cases where the data is small enough, or the user wants to manually distribute execution (say they want to use some other batch scheduler)

In a future PR we will move the `get_annotation_result_from_query` and `get_annotation_result_from_query_async` functions out of proteomics.py, and into api/search.py

Example Notebooks:
[ProteomicsProxiedQuery(1).zip](https://github.com/bystrogenomics/bystro/files/15384205/ProteomicsProxiedQuery.1.zip)
[BasicProxyQuery.zip](https://github.com/bystrogenomics/bystro/files/15384206/BasicProxyQuery.zip)

Corresponds to: https://github.com/bystrogenomics/bystro-web/pull/448

### DIFF
--- a/python/python/bystro/api/search.py
+++ b/python/python/bystro/api/search.py
@@ -212,7 +212,6 @@ def get_async_proxied_opensearch_client(auth: CachedAuth, job_id: str, client_ar
 
     # Initialize OpenSearch client with custom connection class
 
-    print(client_args)
     client = AsyncOpenSearch(
         hosts=[{"host": host, "port": int(port)}],
         use_ssl=bool(protocol == "https" or int(port) == 443),
--- a/python/python/bystro/api/search.py
+++ b/python/python/bystro/api/search.py
@@ -5,7 +5,7 @@ from opensearchpy import OpenSearch, RequestsHttpConnection, AsyncOpenSearch, As
 
 from bystro.api.auth import CachedAuth
 
-QUERY_ENDPONIT = "/api/jobs/{job_id}/opensearch"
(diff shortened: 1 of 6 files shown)


python/python/bystro/api/search.py
```diff
@@ -0,0 +1,264 @@
+from typing import Optional, Mapping, Any, Union, Collection
+from requests.auth import AuthBase
+
+from opensearchpy import OpenSearch, RequestsHttpConnection, AsyncOpenSearch, AsyncHttpConnection
+
+from bystro.api.auth import CachedAuth
+
+QUERY_ENDPONIT = "/api/jobs/{job_id}/opensearch"
```
cristinaetrv, line 8: Is this a typo? It matches line 219 and 257 but it seems like it should be QUERY_ENDPOINT?

cristinaetrv (approved):
Lgtm!

python/python/bystro/api/search.py
akotlar: hah yeah

PR #510, opened 2024-05-22
Optimize PRS C+T for memory.

cristinaetrv:
Optimizes the C+T preprocess and PRS calculation by replacing pandas operations with pyarrow operations and doing these operations in chunks.

Also addresses remaining suggestions from #493.

### DIFF
--- a/python/python/bystro/prs/preprocess_for_prs.py
+++ b/python/python/bystro/prs/preprocess_for_prs.py
@@ -10,10 +10,6 @@ from enum import Enum
 import pandas as pd
 from pyarrow import feather  # type: ignore
 import pyarrow.dataset as ds  # type: ignore
-import pyarrow.compute as pc  # type: ignore
-import tempfile
-import os
-import psutil
 
 logger = logging.getLogger(__name__)
 
@@ -93,15 +89,15 @@ def _extract_nomiss_dosage_loci(dosage_matrix_path: str, score_loci: set, chunk_
     dataset = ds.dataset(dosage_matrix_path, format="feather")
     filtered_loci = set()
     for batch in dataset.to_batches(batch_size=chunk_size):
-        df = batch.to_pandas()
-        df = df[df['locus'].isin(score_loci)]
-        for column in df.columns:
-            #Checking for missingness will be removed after imputation to the dosage matrix
-            #TODO: Add imputation preprocess step before PRS, expected 8/2024
-            if column != 'locus':
-                df = df[df[column] >= 0]
-                df = df.dropna(subset=[column])
-        filtered_loci.update(df['locus'].tolist())
+        chunk = batch.to_pandas()
+        chunk = chunk[chunk["locus"].isin(score_loci)]
+        for column in chunk.columns:
+            # Checking for missingness will be removed after imputation to the dosage matrix
+            # TODO: Add imputation preprocess step before PRS, expected 8/2024
+            if column != "locus":
+                chunk = chunk[chunk[column] >= 0]
+                chunk = chunk.dropna(subset=[column])
+        filtered_loci.update(chunk["locus"].tolist())
     return filtered_loci
 
 
@@ -175,10 +171,10 @@ def read_feather_in_chunks(file_path, columns=None, chunk_size=1000):
     if columns:
         columns = list(columns)
     for batch in dataset.to_batches(batch_size=chunk_size):
-        df = batch.to_pandas()
+        chunk = batch.to_pandas()
         if columns:
-            df = df[columns]
-        yield df
+            chunk = chunk[columns]
+        yield chunk
 
 
 def get_p_value_thresholded_indices(df, p_value_threshold: float) -> set:
@@ -188,11 +184,9 @@ def get_p_value_thresholded_indices(df, p_value_threshold: float) -> set:
     return set(df.index[df["P"] < p_value_threshold])
 
 
-def generate_overlap_scores_dosage(
-    thresholded_score_loci: set, filtered_dosage_loci: set, p_value_threshold: float
-) -> set:
+def generate_overlap_scores_dosage(thresholded_score_loci: set, filtered_dosage_loci: set) -> set:
     """Compare and restrict to overlapping loci between dosage matrix and thresholded scores."""
-    
+
     overlap_loci = thresholded_score_loci.intersection(filtered_dosage_loci)
     if len(overlap_loci) == 0:
         raise ValueError(
@@ -288,23 +282,21 @@ def ld_clump(scores_overlap: pd.DataFrame, map_directory_path: str) -> tuple[pd.
 
 def finalize_scores_after_c_t(
     gwas_scores_path: str, dosage_matrix_path: str, map_directory_path: str, p_value_threshold: float
-) -> tuple[str, pd.DataFrame]:
+) -> tuple[pd.DataFrame, pd.DataFrame]:
     """Finalize scores for PRS calculation."""
     scores = _load_association_scores(gwas_scores_path)
     preprocessed_scores = _preprocess_scores(scores)
     thresholded_score_loci = get_p_value_thresholded_indices(preprocessed_scores, p_value_threshold)
     dosage_loci_nomiss = _extract_nomiss_dosage_loci(dosage_matrix_path, thresholded_score_loci)
-    overlap_loci = generate_overlap_scores_dosage(
-        thresholded_score_loci, dosage_loci_nomiss, p_value_threshold
-    )
-    scores_overlap = preprocessed_scores[preprocessed_scores.index.isin(overlap_loci)] 
+    overlap_loci = generate_overlap_scores_dosage(thresholded_score_loci, dosage_loci_nomiss)
+    scores_overlap = preprocessed_scores[preprocessed_scores.index.isin(overlap_loci)]
     scores_after_c_t, loci_and_allele_comparison = ld_clump(scores_overlap, map_directory_path)
     scores_after_c_t = scores_after_c_t.sort_index()
     return scores_after_c_t, loci_and_allele_comparison
 
 
 def finalize_dosage_after_c_t(
-    dosage_overlap: pd.DataFrame, scores_after_c_t: pd.DataFrame, loci_and_allele_comparison: pd.DataFrame
+    dosage_overlap: pd.DataFrame, loci_and_allele_comparison: pd.DataFrame
 ) -> pd.DataFrame:
     """Finalize dosage matrix for PRS calculation."""
     merged_allele_comparison_genos = loci_and_allele_comparison.join(dosage_overlap, how="inner")
@@ -322,23 +314,23 @@ def generate_c_and_t_prs_scores(
     p_value_threshold: float = 0.05,
 ) -> dict[str, float]:
     """Calculate PRS."""
-    #This part goes through dosage matrix the first time to get overlapping loci
+    # This part goes through dosage matrix the first time to get overlapping loci
     scores_after_c_t, loci_and_allele_comparison = finalize_scores_after_c_t(
         gwas_scores_path, dosage_matrix_path, map_directory_path, p_value_threshold
     )
 
-    #This part goes through dosage matrix the second time, adjusts dosages,
-    #transposes genotypes, and calculates PRS
-    prs_scores = pd.Series(dtype=float)
+    # This part goes through dosage matrix the second time, adjusts dosages,
+    # transposes genotypes, and calculates PRS
+    prs_scores: pd.Series = pd.Series(dtype=float)
     beta_values = scores_after_c_t["BETA"]
     loci = scores_after_c_t.index
-    for chunk_idx, chunk in enumerate(read_feather_in_chunks(dosage_matrix_path, columns=None, chunk_size=1000)):
-        chunk = chunk[chunk['locus'].isin(loci)]        
+    for chunk_idx, chunk in enumerate(
+        read_feather_in_chunks(dosage_matrix_path, columns=None, chunk_size=1000)
+    ):
+        chunk = chunk[chunk["locus"].isin(loci)]
         if not chunk.empty:
             chunk = chunk.set_index("locus")
-            genos_transpose = finalize_dosage_after_c_t(
-                chunk, scores_after_c_t, loci_and_allele_comparison
-            )
+            genos_transpose = finalize_dosage_after_c_t(chunk, loci_and_allele_comparison)
             prs_scores_chunk = genos_transpose @ beta_values.loc[genos_transpose.columns]
             prs_scores = prs_scores.add(prs_scores_chunk, fill_value=0)
     return prs_scores.to_dict()
--- a/python/python/bystro/prs/preprocess_for_prs.py
+++ b/python/python/bystro/prs/preprocess_for_prs.py
@@ -316,12 +316,12 @@ def generate_c_and_t_prs_scores(
     prs_scores: pd.Series = pd.Series(dtype=float)
     beta_values = scores_after_c_t["BETA"]
     finalized_loci = scores_after_c_t.index
-    dataset = ds.dataset(dosage_matrix_path, format="feather")
-    for batch in dataset.to_batches(batch_size=1000, columns=None, batch_readahead=1):
+    score_loci_filter = pc.field("locus").isin(pa.array(list(finalized_loci)))
+    dosage_ds = ds.dataset(dosage_matrix_path, format="feather").filter(score_loci_filter)
+    for batch in dosage_ds.to_batches(batch_size=1000, columns=None, batch_readahead=1):
         chunk = batch.to_pandas()
         if chunk.empty:
             continue
-        chunk = chunk[chunk["locus"].isin(finalized_loci)]
         chunk = chunk.set_index("locus")
         genos_transpose = finalize_dosage_after_c_t(chunk, loci_and_allele_comparison)
         prs_scores_chunk = genos_transpose @ beta_values.loc[genos_transpose.columns]
--- a/python/python/bystro/prs/preprocess_for_prs.py
+++ b/python/python/bystro/prs/preprocess_for_prs.py
@@ -75,7 +75,7 @@ def _load_genetic_maps_from_feather(map_directory_path: str) -> dict[str, pd.Dat
     return genetic_maps
 
 
-def _extract_nomiss_dosage_loci(dosage_matrix_path: str, score_loci: set) -> set:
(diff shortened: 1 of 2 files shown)


python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -184,39 +195,41 @@ def generate_thresholded_overlap_scores_dosage(
...
         thresholded_indices = get_p_value_thresholded_indices(chunk, p_value_threshold)
-        thresholded_indices_set.update(thresholded_indices)
-    scores_filtered_indices = sorted(list(thresholded_indices_set))
-    thresholded_scores = scores.iloc[scores_filtered_indices]
-    thresholded_scores = thresholded_scores.set_index("SNPID")
+        if thresholded_indices:
+            thresholded_scores_chunks.append(chunk.loc[thresholded_indices])
 
-    if len(thresholded_scores) == 0:
+    if not thresholded_scores_chunks:
```
akotlar, line 205: ValueErrors are reserved for cases where the input arguments are wrong. Lets use RuntimeError here

python/python/bystro/prs/tests/test_preprocess_for_prs.py
```diff
@@ -91,28 +91,38 @@ def mock_bin_mappings():
...
-            "Sample1": (0 * 0.1) + (1 * -0.2),
-            "Sample2": (2 * 0.1) + (0 * -0.2),
-        }
+        [0.7, 0.30000000000000004, 0.5], index=["sample1", "sample2", "sample3"]
     )
+
     return genos_transpose_mock, scores_overlap_adjusted_mock, expected_results
 
 
+def mock_read_feather_in_chunks(**kwargs):
```
akotlar, line 113:
```suggestion
def mock_read_feather_in_chunks(*args, **kwargs):
```

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -184,39 +195,41 @@ def generate_thresholded_overlap_scores_dosage(
...
         raise ValueError("No thresholded scores available; cannot proceed with PRS calculation.")
```
akotlar, line 206:
```suggestion
        raise RuntimeError("No thresholded scores available; cannot proceed with PRS calculation.")
```

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -184,39 +195,41 @@ def generate_thresholded_overlap_scores_dosage(
     gwas_scores_path: str, dosage_matrix_path: str, p_value_threshold: float
```
akotlar, line 195: generate_thresholded_overlap_scores_dosage seems like an important function, could we add a test for it?

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -184,39 +195,41 @@ def generate_thresholded_overlap_scores_dosage(
...
+    preprocessed_thresholded_scores = preprocessed_thresholded_scores.set_index("SNPID")
+    thresholded_score_loci = set(preprocessed_thresholded_scores.index)
+    dosage_loci_nomiss = _extract_nomiss_dosage_loci(dosage_matrix_path, thresholded_score_loci)
+    filtered_dosage_loci = set(dosage_loci_nomiss["locus"])
 
-    overlap_snps = set_A.intersection(set_B)
-    remaining_snps_scores = len(thresholded_scores) - len(overlap_snps)
-    if remaining_snps_scores == 0:
+    overlap_loci = thresholded_score_loci.intersection(filtered_dosage_loci)
+    remaining_loci = len(preprocessed_thresholded_scores) - len(overlap_loci)
```
akotlar, line 216: Is the purpose of this check to make sure we have some overlap? If so, it doesn't work when there is perfect overlap. Do we want to check `len(overlap_loci) > 0` instead?

akotlar, line 216: ###

akotlar, line 216:
```suggestion
```

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -184,39 +195,41 @@ def generate_thresholded_overlap_scores_dosage(
...
+    if remaining_loci == 0:
         raise ValueError(
-            "No SNPs match between base and target dataset; cannot proceed with PRS calculation."
+            "No loci match between base and target dataset; cannot proceed with PRS calculation."
         )
-    remaining_snps_frac = (len(thresholded_scores) - len(overlap_snps)) / len(thresholded_scores)
+    remaining_snps_frac = (len(thresholded_scores) - len(overlap_loci)) / len(thresholded_scores)
```
akotlar, line 221: Following my last comment, I think we want len(preprocessed_thresholded_scores) / len(thresholded_score_loci) here?

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -184,39 +195,41 @@ def generate_thresholded_overlap_scores_dosage(
...
         (
-            "Only %s of SNPs out of %s SNPs remain which may limit scoring accuracy",
-            remaining_snps_scores,
+            "Only %s of variants out of %s variants remain for PRS, which may limit scoring accuracy",
+            remaining_loci,
             len(thresholded_scores),
         )
-    scores_overlap = thresholded_scores[thresholded_scores.index.isin(overlap_snps)]
+    scores_overlap = preprocessed_thresholded_scores[
+        preprocessed_thresholded_scores.index.isin(overlap_loci)
```
akotlar, line 231:
```suggestion
        preprocessed_thresholded_scores.index.isin(filtered_dosage_loci)
```

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -332,7 +347,10 @@ def finalize_dosage_scores_after_c_t(
     genos_adjusted = merged_allele_comparison_genos.apply(adjust_dosages, axis=1)
     genotypes_adjusted_only = genos_adjusted.iloc[:, 1:]
     genos_transpose = genotypes_adjusted_only.T
-    return genos_transpose, scores_after_c_t
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".feather") as tmpfile:
```
akotlar, line 350: Can these matrix chunks be large? If so, could we either use the folder that the dosage matrix came from (can make a tmp dir in that folder), or thread through a user-provided temp_dir (which in the case of the cluster version, will use say the hg38.yml `temp_dir`, and default to the dosage matrix's folder if no temp_dir specified).

The reason is that the dosage matrix can be large, and currently we don't require that the worker nodes have large drives.

akotlar, line 350: I'm a little concerned that we are reading through the dosage matrix 3 times by this point. Also, what is the memory saving if `dosage_overlap` is already in memory? Can't we just return genos_transpose directly, instead of writing to temp storage and reading back again in the caller `generate_c_and_t_prs_scores`

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -343,10 +361,25 @@ def generate_c_and_t_prs_scores(
 ) -> pd.Series:
     """Calculate PRS."""
     # TODO: Add covariates to model
-    genos_transpose, scores_after_c_t = finalize_dosage_scores_after_c_t(
+    adjusted_dosage_path, scores_after_c_t = finalize_dosage_scores_after_c_t(
```
akotlar, line 364:
```suggestion
    genos_transpose, scores_after_c_t = finalize_dosage_scores_after_c_t(
```

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -343,10 +361,25 @@ def generate_c_and_t_prs_scores(
...
         gwas_scores_path, dosage_matrix_path, map_directory_path, p_value_threshold
     )
-    return genos_transpose @ scores_after_c_t["BETA"]
+    try:
+        prs_scores = pd.Series(dtype=float)
+        beta_values = scores_after_c_t["BETA"]
+        loci = scores_after_c_t.index
+        for chunk in read_feather_in_chunks(adjusted_dosage_path, columns=loci, chunk_size=1000):
```
akotlar, line 371:
```suggestion
        prs_scores = genos_transpose @ beta_values
```

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -343,10 +361,25 @@ def generate_c_and_t_prs_scores(
...
+            prs_scores_chunk = chunk @ beta_values.loc[chunk.columns]
+            prs_scores = prs_scores.add(prs_scores_chunk, fill_value=0)
+    finally:
+        cleanup_temp_file(adjusted_dosage_path)
+
+    return prs_scores
+
+
+def cleanup_temp_file(file_path):
```
akotlar, line 380: you may not need this anymore

akotlar (changes requested):
Please see comments

python/python/bystro/prs/preprocess_for_prs.py
akotlar: ValueErrors are reserved for cases where the input arguments are wrong. Lets use RuntimeError here

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -71,39 +71,40 @@ def _load_genetic_maps_from_feather(map_directory_path: str) -> dict[str, pd.Dat
     return genetic_maps
 
 
-def _extract_nomiss_dosage_loci(dosage_matrix_path: str) -> pd.DataFrame:
+def _extract_nomiss_dosage_loci(dosage_matrix_path: str, score_loci: set, chunk_size=1000) -> set:
```
akotlar, line 74: I think this is not quite the right way to go about this.

What you want is to use the filter operation, because 1) there are opportunities to skip reading data (called predicate pushdown; won't work for us because of how we've structured the locus column, but in the future we will change this), 2) we keep the filtering in arrow, which will be faster and more memory efficient than converting to pandas first

akotlar:
```suggestion
def _extract_nomiss_dosage_loci(dosage_matrix_path: str, score_loci: set, chunk_size=1000) -> set:
    """
    Reads a dataset from the provided dosage matrix Feather file path and filters out
    rows where any column other than 'locus' contains a missing dosage value (represented by -1)
    and loci not in the base GWAS summary statistics dataset.

    Args:
    ----
    dosage_matrix_path (str): The path to the Feather file containing the target dataset.
    score_loci (set): A set of loci from a base GWAS summary statistics dataset.

    Returns:
    -------
    set: A set containing loci with no missing dosage values across all other columns.
    """
    dataset = ds.dataset(dosage_matrix_path, format="feather")
    filtered_loci = set()

    # Creating a filter for score_loci
    score_loci_filter = pc.field("locus").isin(pa.array(list(score_loci)))

    samples = [name for name in dataset.schema.names if name != "locus"]
    filters = [pc.field(sample) >= 0 for sample in samples]

    # Combine the filters using boolean and logic
    combined_filter = filters[0]
    for f in filters[1:]:
        combined_filter = pc.and_(combined_filter, f)

    for batch in dataset.filter(score_loci_filter).to_batches(batch_size=chunk_size):
        chunk = batch.filter(combined_filter).to_table()

        filtered_loci.update(chunk["locus"].to_pylist())

    return filtered_loci```

Comment (akotlar) [reply python/python/bystro/prs/preprocess_for_prs.py:74]: The suggestion needs to be tested, but is more idiomatic, and efficient

File: python/python/bystro/prs/tests/test_preprocess_for_prs.py [hunk 13]
```diff
@@ -88,48 +94,52 @@ def mock_bin_mappings():
...
+def mock_finalize_dosage_after_c_t():
+    def _mock_finalize_dosage_after_c_t(chunk, loci_and_allele_comparison):  # noqa: ARG001
+        return pd.DataFrame(
+            {"ID00096": [1, 1], "ID00097": [0, 1]}, index=["chr1:566875:C:T", "chr1:728951:A:G"]
+        ).transpose()
+
+    return _mock_finalize_dosage_after_c_t
+
+
+def test_extract_nomiss_dosage_loci(mock_dosage_df, mock_scores_loci):
```
akotlar, line 135:
```suggestion
def test_extract_nomiss_dosage_loci(tmp_path):
    # Create a temporary Feather file for testing
    test_file = tmp_path / "test_dosage_matrix.feather"
    data = {
        'locus': ['locus1', 'locus2', 'locus3', 'locus4'],
        'sample1': [0.1, -1, 0.3, 0.4],
        'sample2': [0.2, 0.3, -1, 0.5],
        'sample3': [0.3, 0.4, 0.5, 0.6]
    }
    table = pa.table(data)
    pa.feather.write_feather(table, test_file)
    
    score_loci = {'locus1', 'locus3', 'locus4'}
    
    result_loci = _extract_nomiss_dosage_loci(test_file, score_loci)
    # Only locus4 has no missing values
    assert result_loci == {'locus1', 'locus4'}
```

I prefer testing end-end where possible; in the suggested test we read from disk, as we would in the actual use

python/python/bystro/prs/preprocess_for_prs.py
akotlar: My version also happens to be much faster; leverages multithreading

akotlar: [test_memory_usage_extract.py.zip](https://github.com/user-attachments/files/15537585/test_memory_usage_extract.py.zip)

```
pip install memory-profiler
time  mprof run test_memory_usage_extract.py &> test_memory_usage_extract_ak_suggestion.log
```

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -165,59 +165,34 @@ def _preprocess_genetic_maps(map_directory_path: str) -> dict[int, list[int]]:
 
 
 def read_feather_in_chunks(file_path, columns=None, chunk_size=1000):
```
akotlar, line 167:
```suggestion
def read_feather_in_chunks(file_path, columns=None, chunk_size=1000):
    """Read a Feather file in chunks as pandas DataFrames."""
    dataset = ds.dataset(file_path, format="feather")

    if columns:
        columns = list(columns)

    for batch in dataset.to_batches(batch_size=chunk_size, columns=columns):
        yield batch.to_pandas()
```

This may be more performant, because it will skip materializing the entire dataset in pandas before dropping columns.

akotlar: I think we'll want to set batch_readahead as well, 1 for now. IO will improve with larger batch_readahead, but at likely too much cost to memory based on our previous experiences with ancestry and saving

python/python/bystro/prs/tests/test_preprocess_for_prs.py
cristinaetrv: Ok, should I use a temp feather file everywhere that it goes into the actual functions? I was trying to avoid I/O for tests in general so that they run quickly, which is why I was trying to avoid using temp feather files, though if you think that's better I can start including I/O in tests.

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -305,48 +281,60 @@ def ld_clump(scores_overlap: pd.DataFrame, map_directory_path: str) -> tuple[pd.
...
     merged_allele_comparison_genos = merged_allele_comparison_genos.sort_index()
-    scores_after_c_t = scores_after_c_t.sort_index()
     genos_adjusted = merged_allele_comparison_genos.apply(adjust_dosages, axis=1)
     genotypes_adjusted_only = genos_adjusted.iloc[:, 1:]
     genos_transpose = genotypes_adjusted_only.T
-    return genos_transpose, scores_after_c_t
+    return genos_transpose
 
 
 def generate_c_and_t_prs_scores(
```
akotlar, line 311: Let's simplify this a bit; we don't need the read_feather_in_chunks anymore, and can directly filter in pyarrow which will be faster. Please try this and update your test:

```suggestion
def generate_c_and_t_prs_scores(
    gwas_scores_path: str,
    dosage_matrix_path: str,
    map_directory_path: str,
    p_value_threshold: float = 0.05,
) -> dict[str, float]:
    """Calculate PRS."""
    # This part goes through dosage matrix the first time to get overlapping loci
    scores_after_c_t, loci_and_allele_comparison = finalize_scores_after_c_t(
        gwas_scores_path, dosage_matrix_path, map_directory_path, p_value_threshold
    )

    # This part goes through dosage matrix the second time, adjusts dosages,
    # transposes genotypes, and calculates PRS
    prs_scores: pd.Series = pd.Series(dtype=float)
    beta_values = scores_after_c_t["BETA"]
    loci = scores_after_c_t.index

    score_loci_filter = pc.field("locus").isin(pa.array(list(loci)))
    dosage_ds = ds.dataset(dosage_matrix_path, format="feather").filter(score_loci_filter)

    for batch in dosage_ds.to_batches(batch_size=1000, batch_readahead=1):
        chunk = batch.to_pandas()
        if chunk.empty:
            continue

        chunk = chunk.set_index("locus")
        genos_transpose = finalize_dosage_after_c_t(batch, loci_and_allele_comparison)
        prs_scores_chunk = genos_transpose @ beta_values.loc[genos_transpose.columns]
        prs_scores = prs_scores.add(prs_scores_chunk, fill_value=0)

    return prs_scores.to_dict()
```

python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -324,16 +315,17 @@ def generate_c_and_t_prs_scores(
...
-            genos_transpose = finalize_dosage_after_c_t(chunk, loci_and_allele_comparison)
-            prs_scores_chunk = genos_transpose @ beta_values.loc[genos_transpose.columns]
-            prs_scores = prs_scores.add(prs_scores_chunk, fill_value=0)
+    finalized_loci = scores_after_c_t.index
+    dataset = ds.dataset(dosage_matrix_path, format="feather")
+    for batch in dataset.to_batches(batch_size=1000, columns=None, batch_readahead=1):
+        chunk = batch.to_pandas()
+        if chunk.empty:
+            continue
+        chunk = chunk[chunk["locus"].isin(finalized_loci)]
```
akotlar, line 324: this isn't quite right; please look at my suggestion; there we do the filtering on the dataset

akotlar (approved):

akotlar:
LGTM

akotlar:
The failure is unrelated; a mysql connection issue to UCSC

PR #512, opened 2024-05-23
[annotation, proteomics, search] Simple interface for querying Bystro Search Engine

akotlar:
* Intuitive interface for querying Bystro search engine and outputting the results to a Pandas dataframe. 
* Flattens representations where possible, outputting array values only when necessary; for consistency some tracks always have their values outputted as array values (configurable, by default the gene track and nearest gene tracks we define in our default yaml)
* Add ability to melt the resulting dataframe by heterozygotes, homozygotes, and missingGenos columns, resulting in a "sample" and "dosage" column, where "sample" records the sample name, and "dosage" is 1 if the sample came from the heterozygotes column, 2 if from homozygotes, or -1 if from missingGenos.
* Change the proteomics joining function to `join_annotation_result_to_fragpipe_dataset` to make it clear that this is specific to fragpipe. 
* Joined proteomic dataset no longer has duplicate columns sample_id and gene_name (which have corresponding `sample` and `refSeq.name2` columns in the annotation dataframe), and we have standardized the protein abundance column, previously called `value` to `protein_abundance`.

Example output attached

WIP label will be removed when tests added

Example outputs:
[example_query_to_df_outputs.zip](https://github.com/bystrogenomics/bystro/files/15412755/example_query_to_df_outputs.zip)

Example notebook:
[ExampleQueryToDataFrame.ipynb.zip](https://github.com/bystrogenomics/bystro/files/15412764/ExampleQueryToDataFrame.ipynb.zip)

TL;DR:

```python
from bystro.api import auth
from bystro.proteomics.annotation_interface import get_annotation_result_from_query

user = auth.login('email', 'password', 'https://bystro-dev.emory.edu')
index_name = '663ac4bfa0e17a1660ba8130_657a50d4b2d0278938ba791d'

query_result_df = get_annotation_result_from_query(
    query_string="*",
    index_name= index_name,
    bystro_api_auth=user,
    melt_by_samples=False
)
query_result_df.head(n=10)
```

Fields of interest can be provided (though the locus fields chrom, pos, inputRef, alt, type, and a generated "locus" field will always be provided):

```python
query_result_df = get_annotation_result_from_query(
    query_string="*",
    index_name= index_name,
    bystro_api_auth=user,
    fields=['gnomad.exomes.AF', 'gnomad.genomes.AF'] # will only output the locus fields + these 2 will be fetched and outputted,
    melt_by_samples=False
)
```


python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -21,15 +22,563 @@
...
+DEFAULT_MULTI_VALUED_TRACKS = [
+    "refSeq",
+    "nearest.refSeq",
+    "nearestTss.refSeq",
+]
+
+# The refSeq.clinvar is a legacy track, which has been removed in all recent database versions
+# During join, the refSeq.clinvar track excludes missing values
+# resulting in a reduction in dimension relative to the primary key
+# This libarary assumes that the length of the primary key array
```
cristinaetrv, line 114:
```suggestion
# This library assumes that the length of the primary key array
```
Found typo, and I'm not sure if this statement is missing something, that the length of the primary key array is ____?

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -21,15 +22,563 @@
...
+# {  # noqa
+#   "foo": ["a", "b", "c"],  # noqa
+#   "bar": [["x1", "x2"], ["y1", "y2"], ["z1", "z2"]]  # noqa
+# }  # noqa
+# but not
+# {  # noqa
+#  "foo": ["a", "b", "c"],  # noqa
+#  "bar": ["x", "y"]  # noqa
+# }  # noqa
+# because we no longer know which value of bar corresponds to which value of foo
```
cristinaetrv, line 130: I like this explanation with examples

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -21,15 +22,563 @@
...
+                num = obj
+
+                if convert_key not in not_numeric_fields and _looks_like_float(obj):
+                    num = float(obj)
+                if not isinstance(num, (int, float)):
+                    return obj
+                if num == 0:
+                    return 0
+                abs_num = abs(num)
+                if abs_num < 0.0001 or abs_num > 1000000:
```
cristinaetrv, line 423: What is this range of numbers meant to represent?

python/python/bystro/examples/ExampleQueryToDataFrame.ipynb
```diff
@@ -0,0 +1,1306 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# An example of a large query\n",
+    "import time\n",
+    "\n",
+    "start = time.time()\n",
+    "query_result_df_big_daly = get_annotation_result_from_query(\n",
```
cristinaetrv, line 1220: I know this is the largest dataset we can use right now, but I think would change the name or use a public dataset for example datasets, especially if people want to test the same dataset and get the same result, or just put down the stats for how long it takes for a dataset of size ___ .

python/python/bystro/proteomics/tests/example_abundance_gene_MD.tsv
```diff
@@ -0,0 +1,1000 @@
+Index	NumberPSM	ProteinID	MaxPepProb	ReferenceIntensity	CPT0088900003	CPT0079270003	CPT0088920001	CPT0079300001	CPT0088550004	QC6	CPT0014450004	CPT0088570001	CPT0014470001	CPT0006730001	CPT0069190001	CPT0092730003	CPT0092740003	QC7	CPT0006630003	CPT0025920001	CPT0025880003	CPT0069160003	CPT0007870001	CPT0001270001	CPT0077110003	QC8	CPT0077140001	CPT0001260009	CPT0076350001	CPT0076330003	CPT0007860003	CPT0079380003	CPT0015810003	CPT0086030003	CPT0085670003	CPT0025230003	CPT0065750003	CPT0015730003	CPT0078800003	CPT0079230003
```
cristinaetrv, line 1: Are these mock examples or actual datasets?

python/python/bystro/proteomics/tests/example_experiment_annotation_file.tsv
```diff
@@ -0,0 +1,41 @@
+plex	channel	sample	sample_name	condition	replicate
```
cristinaetrv, line 1: Same question as for example_annotation_df

python/python/bystro/proteomics/annotation_interface.py
akotlar: Good catch; this was a carryover from the javascript code that formatted large numbers in scientific notation, to make displaying those values in cards easier/more readable

python/python/bystro/proteomics/tests/example_abundance_gene_MD.tsv
akotlar: For posterity: Cristina and I discussed this, and these are example fragpipe outputs, shared by the lab that maintains that pipeline.

cristinaetrv (approved):
Lgtm!

PR #513, opened 2024-05-24
WIP Integrate covariance methods into poe analysis methods

IlhaH:
Adds basic framework of allowing covariance regularization and any kind of optimal_shrinkage loss to be used when calculated POE effects. Assumes that an instance of the covariance regularization class is passed in, and a string of the singular value shrinkage loss is passed in.


austinTalbot7241993 (approved):
LGTM

PR #514, opened 2024-05-27
[WIP][annotation, proteomics] Refine annotation api interface,  proteomic api join interface, add ability to join somascan data

akotlar:
* Refines annotation dataset: melt_by_field => explode_field, melt_by_samples => melt_samples, and improve docstring explanations for both. These names more accurately describe what these arguments do
* Refine proteomic join interface, and TandemMassTagDataset: TandemMassTagDataset class is now a msgspec.Struct, and we do not force drop fields. We also now do not force rename fields, besides Index => gene_name; this may be restored in the future as we support additional variations of Frappe datasets
* Enable ability to join on either gene_name or ProteinID columns.
* Improve the example proteomic notebook to better demonstrate melting_samples and explode_field

WIP because we still need to add the function to join on somascan data, and expand tests.


akotlar:
closing in favor of https://github.com/bystrogenomics/bystro/pull/515

With Github's implementation of squash merge, rebasing is too complex

PR #515, opened 2024-05-28
[annotation, proteomics] Refine annotation api interface, proteomic api join interface

akotlar:
* Refines annotation dataset: melt_by_field => explode_field, melt_by_samples => melt_samples, and improve docstring explanations for both. These names more accurately describe what these arguments do
* Refine proteomic join interface, and TandemMassTagDataset: TandemMassTagDataset class is now a msgspec.Struct, and we do not force drop fields. We also now do not force rename fields, besides Index => gene_name; this may be restored in the future as we support additional variations of Frappe datasets
* Enable ability to join on either gene_name or ProteinID columns.
* Improve the example proteomic notebook to better demonstrate melting_samples and explode_field

### DIFF
--- a/python/python/bystro/examples/ExampleQueryToDataFrame.ipynb
+++ b/python/python/bystro/examples/ExampleQueryToDataFrame.ipynb
@@ -250,7 +250,7 @@
     }
    ],
    "source": [
-    "# Default behavior is to show 1 sample per row, e.g. `melt_by_samples=True`\n",
+    "# Default behavior is to show 1 sample per row, e.g. `melt_samples=True`\n",
     "query_result_df_array_of_structs_select_fields = get_annotation_result_from_query(\n",
     "    query_string=\"*\",\n",
     "    index_name=index,\n",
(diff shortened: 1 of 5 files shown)


python/python/bystro/examples/ExampleQueryToDataFrame.ipynb
```diff
@@ -598,14 +598,14 @@
     }
    ],
    "source": [
-    "# You can disabled sample melting by setting `melt_by_samples=False`\n",
+    "# You can disabled sample melting by setting `melt_samples=False`\n",
```
cristinaetrv, line 601:
```suggestion
    "# You can disable sample melting by setting `melt_samples=False`\n",
```

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -614,29 +616,48 @@ async def execute_query(
...
+        fields (list[str] | None):
+            A list of fields to include in the DataFrame.
+        structs_of_arrays (bool)
+             Whether to return structs of arrays, defaults to True.
+        melt_samples (bool):
+            Whether to unpivot `heterozygotes`, `homozygotes`, and `missingGenos` fields.
+            When `True` the resulting DataFrame will have 2 new columns: `samples` and `dosage`,
+            and `heterozygotes`, `homozygotes`, and `missingGenos` will be removed.
+
+            The `dosage` column will have values of 1, 2, or -1, corresponding to whether the sample was
```
cristinaetrv, line 628: What about the case of homozygote ref (0)?

python/python/bystro/proteomics/annotation_interface.py
```diff
@@ -614,29 +616,48 @@ async def execute_query(
...
+        if fields is not None and explode_field not in fields:
             raise ValueError(
-                f"melt_by_field={melt_by_field} is not in fields={fields}, but must be present."
+                f"explode_field={explode_field} is not in fields={fields}, but must be present."
             )
 
         if structs_of_arrays is False:
             raise ValueError(
-                "Cannot yet, melt by field when structs_of_arrays is False, "
+                "Cannot yet, explode field when structs_of_arrays is False, "
```
cristinaetrv, line 660: I'm a little confused what this error message means

cristinaetrv (approved):
Lgtm!

python/python/bystro/proteomics/annotation_interface.py
akotlar: Won’t happen. By definition we don’t store samples that are homozygous reference, because they don’t have the variant in the row. Thats a good point though, we should think about how to handle this case

python/python/bystro/proteomics/annotation_interface.py
akotlar: I’ll improve it. It means that you are trying to explode list values in say refseq.name2, but there is no refseq.name2 column, because you’ve chosen the array of structs output, where there is a refSeq column that is a dict of {name2: gene, name: transcript, etc}. 

Say one of those fields is array values. We would need to generate multiple dicts with one value each for that key in each of them. We don’t do that yet.

PR #519, opened 2024-06-05
[api, cli] Allow login/signup with --host that has trailing slash and stream files to disk or stdout

akotlar:
* Fix logging in with host that has trailing slash; when fetching files
* Stream outputs to stdout or to disk


cristinaetrv (approved):
Lgtm!

PR #520, opened 2024-06-06
[annotation, api] Improve streaming interface

akotlar:
* Adds ability to yield chunks rather than write them to stdout or out_dir
* Creates stream_and_decompress_file which will also decompress the chunks

TL;DR:
```python
line = ''
lines = []
for chunk in streaming.stream_and_decompress_file(job_id, output=True, key_path='annotation'):
    line_chunk = chunk.decode('utf-8')
    
    if "\n" in line_chunk:
      parts = line_chunk.split("\n")
      lines.append(line + parts[0])
      line = parts[1]
```


cristinaetrv (approved):
Lgtm!

Issue #501, opened 2024-05-15
Sprint 12 Task List


cristinaetrv:
Due date: June 6 2024

Documentation

- [ ] Separate out annotator description/perl side including performance figures, describe every piece that repo has including Machine Learning subsection, Bioinformatics tools subsection (installation first) - 6/6/24
- [ ] GIF of how you would use general purpose ML library - 6/6/24

Proteomics

- [x] Review and complete Dennis's PRs (somascan upload, API endpoint to submit filtering prot jobs) @akotlar - 5/20/24
- [x] Run pQTL analysis and compare domain adaptation to TAMPOR results (compare plots) @akotlar - 5/22/24 - done; we didn't run pQTL analysis because the n=300 dataset doesn't have genetic data we have been able to access, yet. However, Cutler signed off on the results looking good and wants to publish a paper on the work.
- [x] Demo notebook similar to Dec demo but updated to show updates to filtering by annotation @akotlar - 5/22/24  - https://github.com/bystrogenomics/bystro/blob/069b8d8bf7e47a071f71a11f76af97e2e2af0f58/python/python/bystro/examples/ProteomicsProxiedQuery.ipynb
- [x] Add ability to query OpenSearch indices from outside the cluster - @akotlar 5/22/24 - https://github.com/bystrogenomics/bystro/blob/069b8d8bf7e47a071f71a11f76af97e2e2af0f58/python/python/bystro/examples/ExampleQueryToDataFrame.ipynb
- [x] Add ability to filter proteomic data from outside the cluster - @akotlar 5/22/24
- [x] Hook up the filtering by annotation to queue @akotlar - 5/24/24 - update 5/23/24; we don't need this now that we have filtering outside the cluster, until we have a concrete use case in the web app.
- [ ] Filtering needs to be generalized to SomaScan @akotlar -5/24/24
- [x] Generate network analysis results using SPPCA on ~300 sample dataset @akotlar -5/29/24

PRS

- [ ] Automatically launch PRS after ancestry from API server - @akotlar @cristinaetrv -5/31/24
- [x] Adjustable p-value thresholds - @cristinaetrv  - 5/24/24 - https://github.com/bystrogenomics/bystro/pull/507
- [ ] Display basic PRS results in webapp (table with individuals and their score) - @akotlar - 6/6/24 - https://github.com/bystrogenomics/bystro/pull/509
[...]

PR #523, opened 2024-06-08
Issue #501: Add simple parsing function to explode on transcript

akotlar:
* Adds function to take a Bystro annotation tsv and explode it on refSeq.name, segregating the refSeq annotations as necessary to make sure each field has only the values for that transcript

Before:
[trio_trim_vep_vcf-20240524162149542.annotation.tsv.gz](https://github.com/user-attachments/files/15746318/trio_trim_vep_vcf-20240524162149542.annotation.tsv.gz)

After:
[test_output.tsv.gz](https://github.com/user-attachments/files/15746319/test_output.tsv.gz)

Notice at loci like `chrom:chr2	pos:207647981	type:SNP	inputRef:C	alt:T` we no longer have `FASTKD2;FASTKD2;FASTKD2;MIR3130-1;MIR3130-2` in the refSeq.name2 field (the gene name field) and `NM_001136194;NM_014929;NM_001136193;NR_036077;NR_036078` in the refSeq.name field (the transcript name field). Instead we have 5 rows, with 1 transcript/gene per row (the same is true for all of the refSeq.* fields; they are exploded relative to the refSeq.name field).

### DIFF
--- a/python/python/bystro/utils/parse_annotation.py
+++ b/python/python/bystro/utils/parse_annotation.py
@@ -42,7 +42,7 @@ def parse_tsv_streaming(input_file_path, output_file_path):
 
     for row in reader:
         # For refSeq fields, we annotate one entry per disrupted base in indels
-        # We deduplicate the outpu, so we just need to figure ou the maximum length
+        # We deduplicate the output, so we just need to figure out the maximum length
         exploded_rows = []
         refseq_values = {field: row[field].split("|") for field in refseq_fields}
         max_length = max(len(values) for values in refseq_values.values())
--- a/python/python/bystro/utils/parse_annotation.py
+++ b/python/python/bystro/utils/parse_annotation.py
@@ -1,0 +1,112 @@
+import csv
+import gzip
+import sys
+
+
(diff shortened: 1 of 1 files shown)


python/python/bystro/utils/parse_annotation.py
```diff
@@ -0,0 +1,112 @@
...
+    assert primary_key in refseq_fields
+
+    non_refseq_fields = [field for field in fieldnames if field not in refseq_fields]
+
+    writer = csv.DictWriter(outfile, fieldnames=fieldnames, delimiter="\t")
+    writer.writeheader()
+
+    for row in reader:
+        # For refSeq fields, we annotate one entry per disrupted base in indels
+        # We deduplicate the outpu, so we just need to figure ou the maximum length
```
cristinaetrv, line 45:
```suggestion
        # We deduplicate the output, so we just need to figure out the maximum length
```

cristinaetrv (changes requested):
We discussed where this code should live and decided it makes more sense for this to be a cli tool rather than being in utils.

akotlar:
we will revisit in the future

PR #524, opened 2024-06-08
[chore] Issue #501: Update dependencies and remove attrs

akotlar:
* Updates version of pandas, numpy, ray, torch, tqdm
* tqdm update resolve dependabot low severity issue
* Finish removing `attrs` from codebase


python/python/bystro/proteomics/tests/test_proteomics_types.py
```diff
@@ -11,15 +10,6 @@
 def test_ProteomicsSubmission():
     ProteomicsSubmission("foo.tsv")
 
-
-def test_ProteomicsSubmission_bad_input():
```
akotlar, line 15: This test isn't needed because mypy type rules correctly disallow ProteomicsSubmission(3) (why Pat had to provide type: ignore on that line). 

These types are entirely used for internal response processing, so not user facing. Also, in general we do not want to rely on runtime error checking in Python code.

python/python/bystro/proteomics/proteomics_types.py
```diff
@@ -1,56 +1,44 @@
...
 
 
-def _tsv_validator(_self: object, _attribute: attrs.Attribute, tsv_filename: str) -> None:
-    """Check that tsv_filename ends with ".tsv."""
-    if not isinstance(tsv_filename, str):
-        err_msg = (
-            f"tsv_filename must be of type str, got: {tsv_filename} of type {type(tsv_filename)} instead"
-        )
-        raise TypeError(err_msg)
-    if not tsv_filename.endswith(".tsv"):
```
akotlar, line 39: We don't care about whether it ends with TSV. We should duck type where possible (try to decode). We already have strict checking for FragPipe data in the fragpipe module (fragpipe data is what the "tsv" here refers to; we can rename in the future)

austinTalbot7241993 (approved):
LGTM

PR #530, opened 2024-06-18
[annotator] Ensure annotation server re-connects before sending messages

akotlar:
* GIN's AWS network topology appears to force a TTL on all connections. This presents a problem for our workers, which rely on an open connection to beanstalkd to communicate. This PR attempts to solve the problem for the bystro annotation server. In future PR's we will explore measures in other places (annotator worker process, other servers, workers)
* Also adds more thorough error logging


cristinaetrv (approved):
Lgtm!

PR #529, opened 2024-06-17
[PRS] Ancestry-aware PRS with queue support

akotlar:
* Supports calculating PRS scores from queue request, and writes the results as a tsv
* Enables Pandas 3.0 string inference, using `pd.options.future.infer_string = True`. Going forward we should add this to the tops of all files, to prepare us for Pandas 3.0 (any module in which we import pandas)
* Performs LD clumping using the ancestry top_hit population
* Weighs Betas by allele frequency using the superpopulation probabilities, through the formula:
```python
# for a single person at a single locus
Beta * [ dosage_person1 - 2*( gnomad_af_sas* ancestry_prob_sas_person1 + gnomad_af_afr * ancestry_prob_afr_person1 + gnomad_af_eas * ancestry_prob_eas_person1 + gnomad_af_nfe * ancestry_prob_eur_person1 + gnomad_af_amr * ancestry_prob_amr_person1) ]
```
  - Notice that we consider NFE = EUR super population
  - For hg19, we currently use these allele frequencies:
  ```python
  HG19_GNOMAD_AF_SUPERPOPS = [
      "gnomad.genomes.AF_afr",
      "gnomad.genomes.AF_amr",
      "gnomad.genomes.AF_eas",
      "gnomad.genomes.AF_nfe"
  ]
  ```

  - Trivially AF_sas is always 0, because gnomad.gnomes v2.1.1 doesn't have an SAS population. We renormalize the other superpopulations to sum to 1.

- For hg38 we use:
 ```python
 HG38_GNOMAD_AF_SUPERPOPS = [
    "gnomad.genomes.AF_joint_afr",
    "gnomad.genomes.AF_joint_amr",
    "gnomad.genomes.AF_joint_eas",
    "gnomad.genomes.AF_joint_nfe",
    "gnomad.genomes.AF_joint_sas",
]
```

   - We also renormalize these superpopulation frequencies to sum to 1.

Corresponds to this PR for API server and web app client support: https://github.com/bystrogenomics/bystro-web/pull/463

PR #533, opened 2024-06-18
[WIP] Add ability to extract gnomad AFs for PRS.

cristinaetrv:
Add extraction of gnomad AFs for the 5 superpopulations using query of target dataset annotation in order to enable weighing PRS by gnomad AFs.

### DIFF
--- a/python/python/bystro/prs/preprocess_for_prs.py
+++ b/python/python/bystro/prs/preprocess_for_prs.py
@@ -10,6 +10,7 @@ from pyarrow import feather  # type: ignore
 import pyarrow.dataset as ds  # type: ignore
 import pyarrow.compute as pc  # type: ignore
 import pyarrow as pa  # type: ignore
+from bystro.proteomics.annotation_interface import get_annotation_result_from_query, execute_query, async_get_annotation_result_from_query
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


python/python/bystro/prs/preprocess_for_prs.py
```diff
@@ -61,6 +62,47 @@ def _load_genetic_maps_from_feather(map_path: str) -> dict[str, pd.DataFrame]:
...
+        "(gnomad.genomes.AF_amr:>0)",
+        "(gnomad.genomes.AF_eas:>0)",
+        "(gnomad.genomes.AF_nfe:>0)",
+        "(gnomad.genomes.AF_sas:>0)"
+    ]
+    full_query = allele_frequency_filters + finalized_loci
+    return f'"{" || ".join(full_query)}"'
+
+    
+def _extract_af_and_loci_overlap(score_loci: set, target_dataset_index: str, user: str) -> pd.DataFrame:
```
akotlar, line 86: user is CachedAuth

akotlar: and then we want to take an optional opensearch_cluster_config, which is a dict

akotlar: and both user and opensearch_cluster_config should be optional; we just need to check that one is not None (and has values)

akotlar (changes requested):
A few small changes then LGTM

cristinaetrv:
Included in #529

PR #534, opened 2024-06-20
Bootstrap and permutation test

austinTalbot7241993:
In this PR we add bootstrapped confidence intervals and permutation-test p values to the parent of origin caller.


akotlar (approved):
LGTM; chatted with Austin about the math offline

PR #537, opened 2024-06-27
[scheduler] Periodically touch beanstalkd python jobs

akotlar:
* Periodically touches beanstalkd jobs, to ensure 1) liveness of the benastalkd connection when faced with AWS load balancer TLL, 2) that very long lived jobs remain alive, even if exceeding the TTR
* Don't re-use client connections, to ensure that load balancer TTL doesn't prevent job reservation
* Fix cases where PRS summary stat loci are not present in list of gnomad loci

Because of https://github.com/ray-project/ray/issues/46366, we use Python `threading` threads rather than ray workers to spawn the ray handler function. We will further investigate Ray memory usage characteristics with Arrow Datasets and make a longer term decision on whether to drop Ray entirely, or find a workaround for memory usage.


python/python/bystro/ancestry/listener.py
```diff
@@ -25,33 +25,6 @@
 
 ANCESTRY_TUBE = "ancestry"
 
-class AncestryJobData(BaseMessage, frozen=True, rename="camel"):
```
akotlar, line 28: I had to move this, because listener.py can be a __main__ module. See https://github.com/jcrist/msgspec/issues/394

cristinaetrv (approved):
Lgtm!

PR #539, opened 2024-06-27
Update LD clumping to take loci with best p-value per bin.

cristinaetrv:
Previously we were taking the max absolute value of beta, now we're taking the best (lowest) p-value per bin.


akotlar:
LGTM

akotlar (approved):

PR #543, opened 2024-07-01
Parent of Origin- Multi-snp and Reorg

austinTalbot7241993:
This separates out POEs from the rare variant directory and implements a multivariate version of POE for testing a bunch of SNPs.


akotlar (approved):
LGTM

PR #544, opened 2024-07-06
[scheduler, annotation, saveFromQuery, beanstalkd] Auto-reconnecting perl event client, report total sites processed in completion messages

akotlar:
* Add logic in annotator's (perl) progress publisher to 1) not stall forever if put is attempted when the Perl beanstalkd client connected successfuly to the beanstalkd server, but by the time the put operation is attempted the beanstalkd server is inaccessible, 2) reconnect to the beanstalkd server every 30 seconds to reduce the likelihood that stalling happens
* Ensure that annotator's perl server does not delete a job unless the completion or failure message has successfully been transmitted to the beanstalkd server
* Report totalAnnotated and totalSkipped in annotation and saveFromQuery completion messages, so that if progress messages are not delivered, the API server can still faithfully report the total progress
* Report totalIndexed in index completion messages, for the same reason.


perl/bin/bystro-server.pl
```diff
@@ -71,7 +71,13 @@
 
 say "Running Annotation queue server";
 
-my $BEANSTALKD_TIMEOUT = 60;
+my $BEANSTALKD_RESERVE_TIMEOUT = 10;
+# Needs to be larger than RESERVE_TIMEOUT
```
cristinaetrv, line 75: How large is RESERVE_TIMEOUT? Is that client side?

akotlar: The reserve timeout is client side yeah. Basically what will happen is that we try to put the job, and if we don't get back a response in 10 seconds, we stop trying.

For context, a typically put request takes 1-2 milliseconds, so most cases that takes > 10 seconds probably mean the connection died or the beanstalkd server is dead

akotlar: Before we attempt to reserve however, we attempt to connect to the server. If the connection fails, the put command will also fail, with a "timeout" error.

If connection succeeds, but the server dies during reservation, I think we can still have an issue. That chance is small though, since it would have to happen in the millisecond before the reserve command was received. It may also be possible for the issue to happen during reservation, but 10 seconds is short enough that that is unlikely (and AWS load balancer TTL default is 60 seconds)

akotlar: I could however wrap the reserve command in a way that makes it time out client side to prevent stalling

python/python/bystro/beanstalkd/messages.py
```diff
@@ -13,7 +13,8 @@
 
 # seconds; default AWS load balancer TTL is 60 seconds
 # set the interval 2 seconds beforehand to make room for timing / network delays
-QUEUE_HEARTBEAT_INTERVAL = max(1, float(os.getenv("BYSTRO_BEANSTALKD_HEARTBEAT_INTERVAL", "20")) - 2)
+QUEUE_HEARTBEAT_INTERVAL = max(1, float(os.getenv("BYSTRO_BEANSTALKD_HEARTBEAT_INTERVAL", "30")))
+QUEUE_PRODUCER_TIMEOUT = max(1, float(os.getenv("BYSTRO_BEANSTALKD_PRODUCER_TIMEOUT", "10")))
 
 
 class Event(str, Enum):
```
akotlar, line 20: One note on the python client. In testing, the python beanstalkd client does client-side timeouts if the operation doesn't succeed within the connection (socket) timeout time (here `QUEUE_PRODUCER_TIMEOUT`). So unlike python and javascript, we don't need to enforce our own client-side timeout.

cristinaetrv (approved):
Lgtm!

PR #545, opened 2024-07-08
[scheduler] Even more resilient perl beanstalkd annotation worker

akotlar:
* This change makes it possible to have multiple annotation workers competing for jobs, even in the presence of network instability, and removes race conditions that could result in double submission and parallel workers writing to the same directory in the case of such instability. This change also makes Python beanstalkd workers more robust to network communication issues, such as application load balancer enforced connection interruptions on idle TCP connections.

* Enables annotation jobs to run indefinitely, but while also making those jobs' TTR (time to run, see https://github.com/beanstalkd/beanstalkd/blob/master/doc/protocol.txt) potentially as short as a few seconds. We now use a keep-alive mechanism on job submissions. This makes it possible to retry jobs faster if the worker that picked up the job times out.

* bystro-annotate.pl and Seq/ packages now consistently output debug messages to STDERR, and you can now pipe the output from bystro-annotate.pl for downstream processing; the output is a JSON message containing the result paths and result summary:

* Semaphores to make double writes impossible: Seq.pm takes an exclusive lock on`bystro_annotation.lock` file in the target output directory. If it cannot, annotation fails. This file lock is kept until annotation is completed. 
  * Additionally, a `bystron_annotation.completed` file is written before the exclusive lock is released. Presence of this file is checked before annotation begins (after exclusive lock is acquired), to ensure that we do not accidentally re-attempt processing of an annotation after completion. The reason this is important is that in the case of submissions through the Bystro API server / UI, after the job is completed, the output files are used for automatic ancestry, search, and (potentially) PRS calculations. If an annotation worker re-attempts to write the annotation while indexing/ancestry/prs (and in the future) other workers are processing data, we have introduced a race condition that potentially leads to corruption (the annotation files could be in an incomplete state just before the indexer attempts to read them, or during read they could be modified, for instance).

* Remove Perl Bystro Annotation server (bystro-server.pl) in favor of calling the bystro-annotate.pl command line program from the Python beanstalkd worker. This is done to consolidate beanstalkd communication code, to improve ease of debugging.

* python/beanstalkd/worker.py: now runs `handler_fn` using ProcessPoolExecutor (instead of ThreadPoolExecutor), so as to not conflict with Ray's signal handler, and to make it possible to kill all handler_fn child processes.

* python/beanstalkd/worker.py: kills the child processes associated with `handler_fn` if the job touch() fails with NOT_FOUND while the `handler_fn` child processes are still running. `NOT_FOUND` indicates that the job is no longer available to the worker for processing: at this point the worker should attempt to stop job processing (whatever is running in handler_fn), and must ensure that any other workers of the same kind that pick up the job would supersede it.

* python/beanstalkd/worker.py: finer-grained error handling, and will not die on client timeout errors. Instead we sleep for 1s and re-attempt connection, indefinitely (or until the process is killed).

Example of new perl/bin/bystro-annotate.pl output:
```sh
(bystro) (base) ubuntu@ip-10-98-135-15:~/bystro/perl$ perl bin/bystro-annotate.pl --in ~/bystro/trio.trim.vep.vcf.gz --out test_tri/test --config ~/bystro/config/hg19.yml | jq
{
  "totalSkipped": 0,
  "totalProgress": 13341,
  "error": null,
  "results": {
    "annotation": "test.annotation.tsv",
    "config": "hg19.yml",
    "log": "test.annotation.log.txt",
    "sampleList": "test.sample_list",
    "dosageMatrixOutPath": "test.dosage.feather",
    "header": "test.annotation.header.json",
    "statistics": {
      "json": "test.statistics.json",
      "qc": "test.statistics.qc.tsv",
      "tab": "test.statistics.tsv"
    }
  }
}
```

Additional Background and Motivation:
1. Previously, there was a small chance that operations on jobs like delete, release could stall the worker forever, if the beanstalkd server became unavailable in the microseconds between the job connection establishment and delete/release (we always connect before attempting those, so failure would have to occur between those two operations). Additionally, if the beanstalkd server became unavailable in the millisecond or so while those operations were running, the worker could stall.
  * The Python beanstalkd worker solves this because all operations on the socket respect the socket_timeout, whereas in Perl they do not and require us to implement client-side mechanisms

2. Perl MCE (Many Cores Engine), at least in our hands, could not launch a 2nd parallel process for touching jobs (would conflict with launching the process pool for annotation), making it necessary to fork the annotation process anyway.

3. By having the beanstalkd worker run bystro-annotate.pl, rather than calling Seq.pm directly, we ensure our command line interface gets use, and that we discover usability improvements, as we have done in this PR.

4. Annotation jobs required long TTR (time to run) leases, because the annotation workers would not periodically touch the job to refresh the lease. If the client became unresponsive, say due to network outage, such that the job would not be completed or failed (or communication from the worker to beanstalkd server during delete/release operations failed), the job would only be retried after the TTR lease expired. Currently that lease is 48 hours. With this change jobs can run as long as needed, even with short TTRs, so that retrying the job after unresponsive client happened much faster (we could set the TTR to say 30 minutes).


perl/lib/Seq/Role/Message.pm
```diff
@@ -24,8 +24,8 @@ use Carp        qw/croak/;
 use Time::HiRes qw(time);
 use Try::Tiny;
 
-my $PUBLISHER_ACTION_TIMEOUT  = 30;
-my $PUBLISHER_CONNECT_TIMEOUT = 10;
+my $PUBLISHER_ACTION_TIMEOUT  = 20;
```
akotlar, line 27: I increased these, because in the case of network issues, it is probably better to waste a few minutes than stop sending messages altogether; this also now matches the Python version.

Also, I set the ACTION_TIMEOUT (which controls the heartbeat) to be smaller than the CONNECT_TIMEOUT which controls the socket timeout, because in some socket implementations (e.g. Python) operations on the socket are subject to the socket timeout, and so it would be impossible for the heartbeat to ever run. That isn't the case for Perl, but that is an implementation detail, and we sacrifice nothing by making the connection timeout longer than the heartbeat interval.

perl/lib/Seq/Role/Message.pm
```diff
@@ -208,7 +208,7 @@ sub publishMessage {
   my $timeSinceLastInteraction = time() - $lastPublisherInteractionTime;
   if ( $timeSinceLastInteraction >= $PUBLISHER_ACTION_TIMEOUT ) {
     say
-      "Attempting to reconnect progress publisher because time since last interaction is $timeSinceLastInteraction seconds.";
+      "Attempting to reconnect to publisher in publishMessage because time since last interaction is $timeSinceLastInteraction seconds.";
```
akotlar, line 211: We specify where we're calling the publisher to be more specific in logs; otherwise we see sometimes 2 "Attempting to reconnect" back to back.

python/python/bystro/beanstalkd/messages.py
```diff
@@ -13,8 +13,8 @@
 
 # seconds; default AWS load balancer TTL is 60 seconds
 # set the interval 2 seconds beforehand to make room for timing / network delays
-QUEUE_HEARTBEAT_INTERVAL = max(1, float(os.getenv("BYSTRO_BEANSTALKD_HEARTBEAT_INTERVAL", "30")))
-QUEUE_PRODUCER_TIMEOUT = max(1, float(os.getenv("BYSTRO_BEANSTALKD_PRODUCER_TIMEOUT", "10")))
+QUEUE_HEARTBEAT_INTERVAL = max(1, float(os.getenv("BYSTRO_BEANSTALKD_HEARTBEAT_INTERVAL", "20")))
+QUEUE_PRODUCER_TIMEOUT = max(1, float(os.getenv("BYSTRO_BEANSTALKD_PRODUCER_TIMEOUT", "30")))
```
akotlar, line 17: In Python, the socket timeout (which QUEUE_PRODUCER_TIMEOUT controls) must be longer than the timeout on operations.

startup.yml
```diff
@@ -2,7 +2,7 @@ apps:
   - name: BystroAnnotationServer
     script: perl/bin/bystro-server.pl
     interpreter: "perl"
-    args: -q config/beanstalk.yml --debug
+    args: -q config/beanstalk.yml
```
akotlar, line 5: we now log all debug messages by default (and have removed the debug parameter). better to have detailed logging for diagnosing network issues.

perl/lib/Interface.pm
```diff
@@ -51,7 +52,7 @@ has config => (
   is            => 'ro',
   isa           => 'Str',
   coerce        => 1,
-  required      => 1,
+  required      => 0,
```
akotlar, line 55: This is no longer required, so that we can defer the check of whether we have `config` until we've seen whether we have `json_config`, which may define `config`.

python/python/bystro/beanstalkd/worker.py
```diff
@@ -41,15 +42,46 @@
 logger = logging.getLogger(__name__)
 
 
+class FunctionWrapper:
```
akotlar, line 45: Python ProcessPoolExecutor cannot serialize (to send to other processes) functions that are not defined as top-level functions (e.g. defined in this module, or imported from another module); the functions we're executing are instead passed to the listen() function.

To get around that, we use a custom serializer: cloud pickle, which is the serializer used by Ray, for the same purpose.

python/python/bystro/beanstalkd/worker.py
```diff
@@ -41,15 +42,46 @@
...
+    def __call__(self, *args, **kwargs):
+        worker_pid = os.getpid()
+        self.queue.put(worker_pid)
+
+        fn = cloudpickle.loads(self.fn_ser)
+        return fn(*args, **kwargs)
+
+
 # Signal handler function
+def kill_child_processes(parent_pid=os.getpid(), kill_parent=False):
```
akotlar, line 59: We need this to ensure that the process that`handler_fn` is running in, as well as any child processes handler_fn spawns are cleaned up either on program exit, or if touching the job fails with NOT_FOUND (or another exception), because if that happens the present worker no longer owns the job, and therefore the job is free to be picked up by another worker (applies when multiple workers are found; but even with a single worker we can get behavior that appears like multiple workers, because event message lifetimes are decoupled from submission message lifetimes).

akotlar:
Since the main comment is busy:

1. I've tested this extensively on bystro-dev, under a variety of workloads, and doing my best to break it, by using iptables to drop packets, restarting beanstalkd during submissions, using `conntrack -D -p tcp --dport 11300 --timeout N` to drop connections older than N seconds, etc. All works as expected. Workers, if they ever find themselves concurrently processing an annotation, which is only possible under issues of severe network fuckery, will result in the losing worker failing the job. It should also be possible under odd circumstances for the job to copmlete, but in either case, we will never get corruption (unless FLOCK doesn't work on the filesystem). There is room for improvement here, but for now this is good enough for what should be an extreme corner case, and practically impossible with sufficiently long TTR.

2. This is live on bystro-dev. Please try to submit jobs and break things.

perl/lib/Seq.pm
```diff
@@ -586,6 +612,26 @@ sub annotateFile {
...
+  }
+
+  $self->log( 'info', "Created completion file" );
+
+  try {
+    close($lockFh);
+    unlink($lockPath);
+  }
+  catch {
+    $self->log( 'warn', "Failed to close and delete lock file: $_" );
```
cristinaetrv, line 632: Change to STDERR

cristinaetrv (approved):
Walked through the overall process changes in miro (https://miro.com/app/board/uXjVKzUcjLM=/?share_link_id=838501974882), recommended to add workflow to documentation in the future.
Overall, lgtm!

PR #552, opened 2024-07-29
Added BCA confidence intervals

austinTalbot7241993:
This uses the BCA algorithm from the bootstrap to compute confidence intervals on the POE estimates.

### DIFF
--- a/python/python/bystro/parent_of_origin/parent_of_origin.py
+++ b/python/python/bystro/parent_of_origin/parent_of_origin.py
@@ -42,6 +42,7 @@ from tqdm import trange
 from numba import jit  # type: ignore
 
 from sklearn.utils import resample  # type: ignore
+from scipy.stats import norm  # type: ignore
 
 from bystro.covariance.optimal_shrinkage import optimal_shrinkage
 from bystro.covariance._covariance_np import (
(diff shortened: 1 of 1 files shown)


python/python/bystro/parent_of_origin/parent_of_origin.py
```diff
@@ -362,16 +363,59 @@ def fit(
...
+                j_values_vector_component = np.array(
+                    [
+                        np.delete(bootstrap_vector_component, i).mean()
+                        for i in range(n_p_b)
+                    ]
+                )
+                jackknife_mean_vector_component = (
+                    j_values_vector_component.mean()
+                )
+                a_vector_component = np.sum(
```
akotlar, line 384: Could you explain the constants here (** 3, 6.0 * np.sum, **1.5, etc) or reference the equation?

akotlar (approved):
LGTM

python/python/bystro/parent_of_origin/parent_of_origin.py
austinTalbot7241993: Referenced the equation, it requires a whole chapter to explain basically.

PR #577, opened 2024-09-23
Extra checks, allow for alternative allle to be reference

austinTalbot7241993:
This allows for cases where the alternative allele is more common than the reference allele. Can be disabled.

## Summary by CodeRabbit

- **New Features**
	- Introduced additional validation for input values, ensuring only 0, 1, or 2 are accepted.
	- Added a new configuration option to allow setting an alternative allele as the reference during processing.

- **Bug Fixes**
	- Improved logic for determining homozygote indices based on input counts, ensuring accurate computations.

### DIFF
--- a/python/python/bystro/parent_of_origin/parent_of_origin.py
+++ b/python/python/bystro/parent_of_origin/parent_of_origin.py
@@ -84,8 +84,6 @@ class BasePOE:
             raise ValueError("Too small of homozygous sample size, (>30)")
         if np.sum(y == 1) < 30:
             raise ValueError("Too small of heterozygous sample size, (>30)")
-        if not np.all(np.isin(y, [0, 1, 2])):
-            raise ValueError("The array contains values other than 0, 1, or 2")
 
     def transform(
         self, X: np.ndarray, return_inner: bool = False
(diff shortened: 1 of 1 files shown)


python/python/bystro/parent_of_origin/parent_of_origin.py
```diff
@@ -254,7 +261,11 @@ def fit(
         self._test_inputs(X, y)
         self.n_phenotypes = X.shape[1]
 
-        X_homozygotes = X[y == 0]
+        n_0 = np.sum(y == 0)
+        n_2 = np.sum(y == 2)
+        hetero_idx = 2 if self.allow_alternative_reference & (n_2 > n_0) else 0
+
+        X_homozygotes = X[y == hetero_idx]
```
akotlar, line 268: homo_idx?

python/python/bystro/parent_of_origin/parent_of_origin.py
```diff
@@ -84,6 +84,8 @@ def _test_inputs(self, X: np.ndarray, y: np.ndarray) -> None:
             raise ValueError("Too small of homozygous sample size, (>30)")
         if np.sum(y == 1) < 30:
             raise ValueError("Too small of heterozygous sample size, (>30)")
+        if not np.all(np.isin(y, [0, 1, 2])):
+            raise ValueError("The array contains values other than 0, 1, or 2")
```
akotlar, line 88: We should explicitly exclude values less than 0 or which are nan. Bystro genotype dosage matrices encode missing genotypes as -1

akotlar (changes requested):

python/python/bystro/parent_of_origin/parent_of_origin.py
austinTalbot7241993: You mean eliminate them rather than having the method fail?

akotlar (approved):

PR #585, opened 2024-10-02
Add ancestry train updates

cristinaetrv:
The ancestry code has been updated offline to train models in 1.5.1. This PR updates the changes to the main train.py code and adds gnomad model-specific training code. A future PR will include preprocessing steps for the current models that have not been recorded yet.


akotlar (approved):
Tests pass, the failing check is due to a golang issue in the test runner that is not reproducible in other test runners or locally. Will investigate further in a future PR, as this is unrelated.

PR #587, opened 2024-10-03
[wip] begin separating platform from open source features

akotlar:
* We are separating out components that are entirely open source, from those that depend on the closed platform and in practice require that for use. By mixing these in 1 repo we dilute the usefulness of the library for those that are not using the Bystro platform.

Going forward the **perl** folder will contain the open-source genomic annotator, and the **python** folder will contain Bystro ML module, geared at high dimensional statistics along with some example open-source applications in genetics/proteomics.


akotlar:
This needs a bit more thought, will close for now

